PR #471, opened 2019-02-07
drgporep: add parents to `decode_domain_block`

schomatis:
```
Add the `parents` of node `v` as an argument to `decode_domain_block` to avoid
requesting (and maybe recalculating) them twice for the same node.
`decode_domain_block` is only called in `prove` where the `parents` of node `v`
(`challenge`, in the caller) are already available.
```

### DIFF
--- a/storage-proofs/src/vde.rs
+++ b/storage-proofs/src/vde.rs
@@ -92,13 +92,12 @@ pub fn decode_domain_block<'a, H, G>(
     replica_id: &'a H::Domain,
     data: &'a [H::Domain],
     v: usize,
+    parents: Vec<usize>,
 ) -> Result<H::Domain>
 where
     H: Hasher,
     G: Graph<H>,
 {
-    let parents = graph.parents(v);
-
     let byte_data = data
         .iter()
         .flat_map(H::Domain::into_bytes)
--- a/storage-proofs/src/vde.rs
+++ b/storage-proofs/src/vde.rs
@@ -86,8 +86,8 @@ where
     Ok(H::sloth_decode(&key, &node_data, sloth_iter))
 }
 
-pub fn decode_domain_block<'a, H, G>(
-    graph: &'a G,
+pub fn decode_domain_block<'a, H>(
+    degree: usize,
     sloth_iter: usize,
     replica_id: &'a H::Domain,
     data: &'a [H::Domain],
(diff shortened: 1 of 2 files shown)


storage-proofs/src/vde.rs
```diff
@@ -86,25 +86,23 @@ where
     Ok(H::sloth_decode(&key, &node_data, sloth_iter))
 }
 
-pub fn decode_domain_block<'a, H, G>(
-    graph: &'a G,
+pub fn decode_domain_block<'a, H>(
+    degree: usize,
```
porcuquine: To be honest, we probably don't need the degree either. It only eventually gets used to assert that we have the right number of parents when performing the kdf. But that actually must be true anyway, if `parents` is calculated correctly. For that reason, I'm tentatively in favor of removing `degree` here — since it's only being passed around to perform a run-time verification which was probably useful when written but doesn't justify contorting the plumbing throughout.

schomatis: > But that actually must be true anyway, if `parents` is calculated correctly.

Why?

schomatis:
@porcuquine I'd rather evaluate removing `degree` in a separate issue.

Even if it does make sense it seems out of the original scope of this PR, if it was only removing an extra line I'd just push that here but removing `degree` seems to entail refactoring the `kdf` function in *many* places, I think that warrants its own PR. (If I'm off here please point me to where a reduced set of changes could be made in this thread, if not feel free to create a separate issue and assign it to me.)

storage-proofs/src/vde.rs
porcuquine: Because the definitions of the `parents` methods for both `Graph`  and `ZigZagGraph` ensure that the length of the results they return will be equal to the graph degree. If we want an explicit guarantee of this, we could put the assertion in the implemention of `parents`.

porcuquine:
@schomatis, That's fine. I'm just following the conversation to its logical conclusion as we ponder a refactor.

schomatis:
Rebased, is this good to go then?

schomatis:
@porcuquine Is there a way to change the GitHub blocking "outdated, need to merge to master" into a rebase (or provide it as an extra option)? While in the process of review I generated 3 `Merge branch 'master' into fix/prove/request-parents-once` commits that end up in `master` which makes the history harder to read. (I'll try to rebase them myself locally before pushing, but just wanted to check if this can be made into an option in the GUI.)

porcuquine:
I don't think that option exists, unfortunately. If you make your final merge using the 'squash' option, then you can avoid those merge commits — but then you don't get a merge commit for your work. Squashing and rebasing is probably the best option, but it does then require seeking reapproval (with our current settings). Let's discuss how we want to approach this tomorrow.

PR #478, opened 2019-02-14
memory cache for groth params and verifying key

### LINKED ISSUE #468: distribute verifying key independent of groth parameters
We currently derive the verifying key from the groth parameters. This means we have to both load and distribute the parameters themselves in order to perform verification. When we support multiple sector sizes, miners should only fetch the parameters for the proofs they need to generate. It should be cheap to download and cache the verifying keys for all proofs, though.

- Make `PreparedVerifyingKey` serializable.
- When generating parameters, also generate and write the verifying key to the disk cache.
- When verifying, load only the verifying key (and probably cache in memory).
- Ensure these keys can be distributed using the parameter delivery mechanism (may require no change).

### LINKED ISSUE #477: Cache groth params in memory
### Description

When groth params are needed, they are either loaded from disk or generated then written out to disk.

Loading is faster, but it's still slow. There exists a mechanism for caching the 'official' parameters in memory. See `ZIGZAG_PARAMS` in `internal.rs`.

We should extend this mechanism to handle at least all parameters that are needed by `go-filecoin`. This can be considered the same set which is generated by `paramcache`. It's not clear whether *all* generated groth parameters should be cached, since this could potentially lead to a lot of memory usage during tests.

### Acceptance criteria

- Examination of logs in `go-filecoin` shows that groth parameters are not being repeatedly slowly loaded. (will require first integrating wiht `go-filecoin` to verify) — so this should not block merging of a tentative solution. However, if the underlying problem is not addressed, this issue should be reopened.

### Risks + pitfalls

- The set of parameters we need to hold in memory may change over time, and it may also become the case that there are too many potential sets to always hold all of them in memory. So we should prefer a solution which will automatically adapt to a changing set, and ideally which is flexible enough to allow further configuration over time.

### Where to begin

- The most natural solution may be to add an in-memory cache to `parameter_cache.rs` and prefer this to the disk cache when configured to do so. Make this the default for all parameter files initially.

porcuquine:
Closes #477.
Closes #468.

---
**UPDATE**

I added support for verifying key as well. This parallels groth parameter generation throughout.

This means that clients who want only to verify (but not generate) proofs will be able to fetch only the verifying keys they require. This is not necessarily an issue now, but may become one in the future as number and size of circuits grows. Creation of verifying keys still necessarily requires groth params (whether generated or fetched).

**NOTE**

This means that there are now new cache files — just like the groth param files, but with a `.vk` suffix added to the corresponding parameter filename.

I published the initial verifying keys corresponding to the already-published groth params, and updated `parameters.json` here.

---

Unlike the approach described in #477, I ended up putting this in the API internal implementation.

It's not clear that we always want this behavior (especially during tests — where many parameters might be generated).

As importantly, when I tried to implement this in `parameter_cache.rs`, I ran up against a problem:
 - The implementation has to be generic over `JubjubEngine` — but the static `HashMap` has to have a concrete type. Since callers of `get_groth_params()` are also generic, code implementing the cache doesn't know for what type to specialize. That problem is avoided by making the cache a more local construct isolated to a specific usage in which the concrete type (`Bls12`) is known.

Because we don't have the full `parameter_set_identifier()` mechanism, parameters are only distinguished by size (bytes) and type ("ZIGZAG" or "POST"). This is fine as long as we don't have multiple sectors of the same size with different security parameters — which we probably never will.

The following log output shows this is working as designed.
```
➜  rust-proofs git:(feat/in-memory-param-cache-alt) time ./target/release/examples/ffi
Feb 13 18:32:15.070 INFO encoding, layer {}: 0, place: storage-proofs/src/layered_drgporep.rs:398 storage_proofs::layered_drgporep, root: storage-proofs
Feb 13 18:32:15.071 INFO encoding, layer {}: 1, place: storage-proofs/src/layered_drgporep.rs:398 storage_proofs::layered_drgporep, root: storage-proofs
Feb 13 18:32:15.071 INFO encoding, layer {}: 2, place: storage-proofs/src/layered_drgporep.rs:398 storage_proofs::layered_drgporep, root: storage-proofs
Feb 13 18:32:15.071 INFO encoding, layer {}: 3, place: storage-proofs/src/layered_drgporep.rs:398 storage_proofs::layered_drgporep, root: storage-proofs
Feb 13 18:32:15.072 INFO returning tree, layer: 0, place: storage-proofs/src/layered_drgporep.rs:391 storage_proofs::layered_drgporep, root: storage-proofs
Feb 13 18:32:15.072 INFO returning tree, layer: 1, place: storage-proofs/src/layered_drgporep.rs:391 storage_proofs::layered_drgporep, root: storage-proofs
Feb 13 18:32:15.072 INFO returning tree, layer: 2, place: storage-proofs/src/layered_drgporep.rs:391 storage_proofs::layered_drgporep, root: storage-proofs
Feb 13 18:32:15.072 INFO returning tree, layer: 3, place: storage-proofs/src/layered_drgporep.rs:391 storage_proofs::layered_drgporep, root: storage-proofs
Feb 13 18:32:15.073 INFO returning tree, layer: 4, place: storage-proofs/src/layered_drgporep.rs:391 storage_proofs::layered_drgporep, root: storage-proofs
Feb 13 18:32:15.073 INFO setting tau/aux, layer: 0, place: storage-proofs/src/layered_drgporep.rs:437 storage_proofs::layered_drgporep, root: storage-proofs
Feb 13 18:32:15.073 INFO setting tau/aux, layer: 1, place: storage-proofs/src/layered_drgporep.rs:437 storage_proofs::layered_drgporep, root: storage-proofs
Feb 13 18:32:15.073 INFO setting tau/aux, layer: 2, place: storage-proofs/src/layered_drgporep.rs:437 storage_proofs::layered_drgporep, root: storage-proofs
Feb 13 18:32:15.073 INFO setting tau/aux, layer: 3, place: storage-proofs/src/layered_drgporep.rs:437 storage_proofs::layered_drgporep, root: storage-proofs
Feb 13 18:32:15.073 INFO trying memory cache for: ZIGZAG[1024], target: params, place: filecoin-proofs/src/api/internal.rs:95 filecoin_proofs::api::internalFeb 13 18:32:15.074 INFO parameter set identifier for cache: layered_drgporep::PublicParams{ drg_porep_identifier: drgporep::PublicParams{graph: zigzag_graph::ZigZagGraph{expansion_degree: 8 base_graph: drgraph::BucketGraph{size: 32; degree: 5} }; sloth_iter: 0}, challenges: Tapered { layers: 4, count: 2, taper: 0.3333333333333333, taper_layers: 2 } }, target: params, , place: storage-proofs/src/parameter_cache.rs:54 storage_proofs::parameter_cache, root: root: filecoin-proofs
storage-proofs
Feb 13 18:32:15.074 INFO checking cache_path: "/tmp/filecoin-proof-parameters/v9-zigzag-proof-of-replication-f8b6b5b4f1015da3984944b4aef229b63ce950f65c7f41055a995718a452204d", target: params, place: storage-proofs/src/parameter_cache.rs:83 storage_proofs::parameter_cache, root: storage-proofs
Feb 13 18:32:15.074 INFO reading groth params from cache: "/tmp/filecoin-proof-parameters/v9-zigzag-proof-of-replication-f8b6b5b4f1015da3984944b4aef229b63ce950f65c7f41055a995718a452204d", target: params, place: storage-proofs/src/parameter_cache.rs:126 storage_proofs::parameter_cache, root: storage-proofs
Feb 13 18:32:24.978 Feb 13 18:32:24.978 INFOINFO got groth params (1024) while sealing groth_parameter_bytes: 770902584, target: stats, target, place: : storage-proofs/src/parameter_cache.rs:131 storage_proofs::parameter_cache, root: storage-proofs
params, place: filecoin-proofs/src/api/internal.rs:467 filecoin_proofs::api::internal, root: filecoin-proofs
Feb 13 18:32:45.374 INFO trying memory cache for: ZIGZAG[1024], target: params, place: filecoin-proofs/src/api/internal.rs:95 filecoin_proofs::api::internal, root: filecoin-proofs
Feb 13 18:32:45.375 INFO got groth params (1024) while verifying seal, target: params, place: filecoin-proofs/src/api/internal.rs:592 filecoin_proofs::api::internal, root: filecoin-proofs
./target/release/examples/ffi  279.49s user 7.31s system 920% cpu 31.153 total
```
Note in the final lines that parameters were acquired in 1ms with no reading.

Compare with the first report of 'trying memory cache' — which takes 9 seconds and reads from disk.

Unfortunately, the log output is slightly mangled because of interleaved entries.

### DIFF
--- a/filecoin-proofs/src/api/internal.rs
+++ b/filecoin-proofs/src/api/internal.rs
@@ -1,13 +1,15 @@
+use std::collections::hash_map::Entry;
+use std::collections::HashMap;
 use std::fs::File;
 use std::io::{BufWriter, Read, Write};
 use std::path::PathBuf;
+use std::sync::Mutex;
 
 use bellman::groth16;
 use pairing::bls12_381::{Bls12, Fr};
 use pairing::{Engine, PrimeField};
 use sapling_crypto::jubjub::JubjubBls12;
 
-use sector_base::api::disk_backed_storage::LIVE_SECTOR_SIZE;
 use sector_base::api::sector_store::SectorConfig;
 use sector_base::io::fr32::write_unpadded;
 use std::path::Path;
@@ -22,7 +24,6 @@ use storage_proofs::hasher::pedersen::{PedersenDomain, PedersenHasher};
 use storage_proofs::hasher::{Domain, Hasher};
 use storage_proofs::layered_drgporep::{self, LayerChallenges};
 use storage_proofs::merkle::MerkleTree;
-use storage_proofs::parameter_cache::{parameter_cache_dir, read_cached_params};
 use storage_proofs::porep::{replica_id, PoRep, Tau};
 use storage_proofs::proof::ProofScheme;
 use storage_proofs::vdf_post::{self, VDFPoSt};
@@ -31,6 +32,7 @@ use storage_proofs::zigzag_drgporep::ZigZagDrgPoRep;
 use storage_proofs::zigzag_graph::ZigZagBucketGraph;
 
 use crate::error;
+use crate::FCP_LOG;
 
 type Commitment = Fr32Ary;
 type ChallengeSeed = Fr32Ary;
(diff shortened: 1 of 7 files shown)


porcuquine:
@phritz @mishmosh @laser Is `go-filecoin` ready for this? Should we prepare a `rust-fil-proofs` release  including this change?

porcuquine:
@laser @sidke See update, FYI. Please let me know if this introduces complexities I'm overlooking.

@dignifiedquire Would appreciate your review.

@whyrusleeping This now sings *and* dances.

whyrusleeping:
@porcuquine youre a saint

filecoin-proofs/src/api/internal.rs
```diff
@@ -70,6 +74,57 @@ lazy_static! {
...
+}
+
+fn lookup_groth_params<F: FnOnce() -> error::Result<Bls12GrothParams>>(
+    identifier: String,
+    generator: F,
+) -> error::Result<Bls12GrothParams> {
+    let cache = &mut (*GROTH_PARAM_MEMORY_CACHE).lock().unwrap();
+    info!(FCP_LOG, "trying groth parameters memory cache for: {}", &identifier; "target" => "params");
+    let params = match cache.entry(identifier) {
+        Entry::Vacant(entry) => entry.insert(generator()?).clone(),
```
dignifiedquire: It seems pretty inefficient to `clone` these values if we fetch them from the cache, we should be able to return a reference to them. 

Doesn't have to be in this PR, I looked into the changes involved and it is a bit involved pulling `Box`s and `&`s through, but should be worth it to reduce memory usage

filecoin-proofs/src/api/internal.rs
```diff
@@ -70,6 +74,57 @@ lazy_static! {
...
+fn lookup_verifying_key<F: FnOnce() -> error::Result<Bls12VerifyingKey>>(
+    identifier: String,
+    generator: F,
+) -> error::Result<Bls12VerifyingKey> {
+    let cache = &mut (*VERIFYING_KEY_MEMORY_CACHE).lock().unwrap();
+    let vk_identifier = format!("{}-verifying-key", &identifier);
+
+    info!(FCP_LOG, "trying verifying key memory cache for: {}", &vk_identifier; "target" => "verifying_key");
+    let verifying_key = match cache.entry(vk_identifier) {
+        Entry::Vacant(entry) => entry.insert(generator()?).clone(),
```
dignifiedquire: same as above

storage-proofs/src/parameter_cache.rs
```diff
@@ -106,6 +106,46 @@ where
...
     }
+
+    fn get_verifying_key(circuit: C, pub_params: &PP) -> Result<groth16::VerifyingKey<E>> {
+        let generate = || -> Result<groth16::VerifyingKey<E>> {
+            let groth_params = Self::get_groth_params(circuit, pub_params)?;
+            info!(SP_LOG, "Getting verifying key."; "target" => "verifying_key");
+            Ok(groth_params.vk)
+        };
+
+        match Self::cache_identifier(pub_params) {
```
dignifiedquire: as far as I can tell `cache_identifier` always returns `Some`, can we just stop returning an `Option` from it?

filecoin-proofs/src/api/internal.rs
porcuquine: If you know how to make this work (completely), I would love to see. I tried a number of things and couldn't avoid the `clone`.

I think the root issue is that `create_proof` takes ownership of the groth params it is passed — so it seems we have to clone them prior to use.

I could easily be missing a sophistication, though. Please feel free to attack this in a follow-up.

storage-proofs/src/parameter_cache.rs
porcuquine: It was there so required implementations could opt out of caching if needed, but in practice it hasn't been needed, so … goodbye.

porcuquine:
@laser This is  approved, rebased and just needs to be checked for `go-filecoin` integration. Please signal gfc integration is verified by approving this PR (if you don't have other blocking requests).

filecoin-proofs/src/api/internal.rs
dignifiedquire: yeah, I have fought this problem recently somewhere else and I am pretty sure I know how to make it work: will work in on it soonish: https://github.com/filecoin-project/rust-fil-proofs/issues/499

porcuquine: Looking more closely at the source, I see that we *should* be able to pass a reference to `create_proof` — so… should work out in the end. Thanks for flagging.

laser:
I will re-integrate this branch with go-filecoin once the code which removes `params.out` from existence lands.

porcuquine:
@laser What is this `params.out` of which you speak? Annihilation complete.

laser:
I've just integrated this with go-filecoin `467462f4218c475f38c49e3c22a8566f2f4cfd13`. Tests pass:

https://circleci.com/gh/filecoin-project/go-filecoin/11469

PR #479, opened 2019-02-14
Various spellery: the, mac OS, license, etc.

comradekingu:
A potential detail is how Rust calls it "Stdout", which in turn is "stdout" https://doc.rust-lang.org/1.29.0/std/io/struct.Stdout.html

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 # Filecoin Proving Subsystem (FPS)
 
-The **Filecoin Proving Subsystem** provides the storage proofs required by the Filecoin protocol.  It is implemented entirely in Rust, as a series of partially inter-dependent crates – some of which export C bindings to the supported API. This decomposition into distinct crates/modules is relatively recent, and in some cases current code has not been fully refactored to reflect the intended eventual organization.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -23,17 +23,17 @@ There are currently four different crates:
...
 
-If at any point it were to become clear that the FFI approach is irredeemably problematic, the option of moving to a standalone **FPS** remains. However, we have now already solved the majority of technical problems associated with calling from Go into Rust even while allowing for a high degree of runtime configurability. Therefore, it seems most likely that we continue down the path in which we have already invested and begin to reap reward.
+If at any point it were to become clear that the FFI approach is irredeemably problematic, the option of moving to a standalone **FPS** remains. However, the majority of technical problems associated with calling from Go into Rust are now solved, even while allowing for a high degree of runtime configurability. Therefore, continuing down the same path we have already invested in, and begin to reap rewards from, seems likely.
```
porcuquine: If we rewrite the last sentence for stylistic reasons, let's also change 'begin' to 'beginning' so we don't introduce a grammatical error.

Personally, I'm not sold on the premise that moving to the passive voice throughout is an improvement. If you drop the changes to Design Notes, this PR seems fine.

comradekingu: @porcuquine I fixed it in English now, some part of my consideration is for how easy it will be to translate into other languages.

porcuquine: Understood — thank you for your efforts.

PR #483, opened 2019-02-15
verify parameter integrity

sidke:
resolves https://github.com/filecoin-project/go-filecoin/issues/1925

introduces two things:
1. less important, but useful is to pass a parameter pointing to the json file to use
2. more important, generate a digest when publishing parameters, and checking fetched parameters against that digest

currently, i opted to just say "hey the data looks like it's corrupt please delete it and try fetching again" because i didn't really want to touch/delete any of the parameter files with the tool

this approach can be backed out of if we want to use ipfs instead to verify cids and, ideally, also drop curling in favor of the ipfs cli


laser:
I recommend tapping @schomatis <https://github.com/schomatis> for review since porcu and I are OOO.

filecoin-proofs/src/bin/paramfetch.rs
```diff
@@ -31,32 +41,46 @@ Set $FILECOIN_PARAMETER_CACHE to specify parameter directory. Defaults to '{}'
...
+                match fetch_parameter_file(&parameter_map, parameter.to_string()) {
+                    Ok(_) => {
+                        println!("checking digest...");
+                        let digest =
+                            get_parameter_digest(parameter.to_string()).expect(ERROR_DIGEST);
 
-                match fetch_parameter_file(p.to_string()) {
-                    Ok(_) => println!("ok"),
+                        if digest != parameter_map.get(parameter).unwrap().digest {
+                            println!("failed integrity check. you should remove this parameter and try again");
```
schomatis: I would consider an intermediate solution here of at least renaming the already downloaded (invalid) parameters file (appending something like `-invalid-digest`) to make it explicit that they don't match the digest (since a `println` can be easily overlooked). That way any posterior script that was waiting for a specific file would fail and would force the user to manually rename the invalid one.

filecoin-proofs/src/param.rs
```diff
@@ -75,18 +81,42 @@ pub fn load_parameter_map() -> Result<ParameterMap> {
...
+    } else {
+        Err(err_msg(ERROR_PARAMETER_FILE))
+    }
+}
+
+pub fn get_parameter_digest(parameter: String) -> Result<String> {
+    let path = parameter_file_path(parameter)?;
+    let mut file = File::open(path)?;
+    let mut buffer = Vec::new();
+    let _ = file.read_to_end(&mut buffer);
```
dignifiedquire: this reads the whole file into memory, which is not needed, you should be able to do sth like this

```rust
let mut file = File::open(parameter)?;
let mut hasher = Blake2b::new();
std::io::copy(&mut file, &mut hasher)?;
let hash = hasher.result();
```

filecoin-proofs/src/bin/parampublish.rs
```diff
@@ -47,31 +47,30 @@ Set $FILECOIN_PARAMETER_CACHE to specify parameter directory. Defaults to '{}'
...
         }
         .expect(ERROR_PARAMETERS_LOCAL);
 
-        if parameters.len() > 0 {
+        if parameter_ids.len() > 0 {
             println!("publishing parameters");
 
-            for parameter in parameters.iter() {
-                println!("publishing '{}'...", parameter);
+            for parameter_id in parameter_ids.iter() {
```
dignifiedquire: if ypu use into_iter you get an owned string so you can use that to insert into the hashmap below

Issue #492, opened 2019-02-20
How do we choose VDF PoSt public parameters "sectors count" when generating Groth parameters for distribution?


laser:
### Description

A rust-fil-proofs developer generates Groth parameters and publishes them to IPFS. go-filecoin developers can fetch these Groth parameters from IPFS so that they don't have to generate them locally. 

When generating Groth parameters, a developer runs the `VDFPostCompound::get_groth_params` function. The `VDFPostCompound::get_groth_params` function requires VDF PoSt public parameters as input.

VDF PoSt public parameters [has a "sectors count"](https://github.com/filecoin-project/specs/blob/8ee10388bedfd9570508a7bf9307e638159a0939/drafts/proof-of-spacetime.md#vdf-post-proof-of-spacetime-based-on-vdfs), which is the "number of sectors over which the proof is performed."

My question is: How do we select a "sectors count" when generating and publishing Groth parameters?

laser:
cc @dignifiedquire

laser:
After a long conversation with @porcuquine today I have learned that my premise was flawed. We don't yet support (and may never support) dynamic sectors count.

VDF PoSt sectors count is currently fixed to 2. This means that VDF PoSt can be generated for 2 sectors at a time.

Example:

Suppose that a miner wants to prove 11 sectors during a proving period. It will need to generate 6 PoSts.

PR #494, opened 2019-02-21
bugs/493 dynamic public parameters sectors count

### LINKED ISSUE #493: Cannot generate a proof-of-spacetime for more or less than 2 sectors
### Description

VDF PoSt setup parameters are [created with a `sectors_count` of 2](https://github.com/filecoin-project/rust-fil-proofs/blob/master/filecoin-proofs/src/api/internal.rs#L149). 

VDF PoSt [prove](https://github.com/filecoin-project/rust-fil-proofs/blob/master/storage-proofs/src/vdf_post.rs#L169) asserts that the number sectors being proven match the public parameters `sectors_count`. 

Since `sectors_count` is always 2, we cannot generate a proof-of-spacetime for a number of sectors which is not equal to 2.

laser:
Fixes #493 and unblocks go-filecoin issue [`#1996`](https://github.com/filecoin-project/go-filecoin/issues/1996) (_PoSt Randomness outside the State Machine_).

## What's in this PR?

- BUGFIX: make VDF PoSt public parameter-generation accept a value to use as "sectors count" instead of hard-coding to 2
- Modify FFI-exposed `generate_post` to call `internal::generate_post`
- Add PoSt generate and verify to FFI-level tests

## Controversial Stuff: newtypes

I've introduced two [newtypes](https://doc.rust-lang.org/1.0.0/style/features/types/newtype.html), `PaddedBytesAmount` and `UnpaddedBytesAmount` and modified the `SectorManager` and `SectorConfig` traits to return/accept values of these types instead of `u64`. This will improve safety (from a programmer bug-producing standpoint) by preventing a programmer from passing a padded bytes amount to a place where an unpadded bytes amount is expected (and vice-versa).

## Controversial Stuff: Groth parameter-generation in `paramcache`

The problem described in #492 means that I had to hard-code a "sectors count" value in the `paramcache` program.

### DIFF
--- a/filecoin-proofs/examples/ffi/main.rs
+++ b/filecoin-proofs/examples/ffi/main.rs
@@ -19,6 +19,7 @@ use rand::{thread_rng, Rng};
 use std::env;
 use std::error::Error;
 use std::ptr;
+use std::slice::from_raw_parts;
 use std::sync::atomic::AtomicPtr;
 use std::sync::mpsc;
 use std::thread;
@@ -310,6 +311,50 @@ unsafe fn sector_builder_lifecycle(use_live_store: bool) -> Result<(), Box<Error
         }
 
         assert_eq!(1, (*resp).sectors_len);
+    }
+
+    // generate and then verify a proof-of-spacetime for the sealed sectors
+    {
+        let resp = get_sealed_sectors(sector_builder_b);
+        defer!(destroy_get_sealed_sectors_response(resp));
+
+        if (*resp).status_code != 0 {
+            panic!("{}", c_str_to_rust_str((*resp).error_msg))
+        }
+
+        let sealed_sector_metadata: FFISealedSectorMetadata =
+            from_raw_parts((*resp).sectors_ptr, (*resp).sectors_len)[0];
+        let sealed_sector_replica_commitment: [u8; 32] = sealed_sector_metadata.comm_r;
+        let challenge_seed: [u8; 32] = [0; 32];
+
+        let resp = generate_post(
+            sector_builder_b,
+            &sealed_sector_replica_commitment[0],
+            32,
+            &challenge_seed,
+        );
+        defer!(destroy_generate_post_response(resp));
+
+        if (*resp).status_code != 0 {
+            panic!("{}", c_str_to_rust_str((*resp).error_msg))
+        }
+
+        let resp = verify_post(
+            &sizes.store,
+            &sealed_sector_replica_commitment[0],
+            32,
+            &challenge_seed,
+            &((*resp).proof),
+            (*resp).faults_ptr,
+            (*resp).faults_len,
+        );
+        defer!(destroy_verify_post_response(resp));
+
+        if (*resp).status_code != 0 {
+            panic!("{}", c_str_to_rust_str((*resp).error_msg))
+        }
+
+        assert!((*resp).is_valid)
     }
 
     // after sealing, read the bytes (causes unseal) and compare with what we
--- a/filecoin-proofs/src/api/internal.rs
+++ b/filecoin-proofs/src/api/internal.rs
@@ -8,7 +8,7 @@ use pairing::{Engine, PrimeField};
 use sapling_crypto::jubjub::JubjubBls12;
 
 use sector_base::api::disk_backed_storage::LIVE_SECTOR_SIZE;
-use sector_base::api::sector_store::SectorConfig;
+use sector_base::api::sector_store::{SectorConfig, UnpaddedBytesAmount};
 use sector_base::io::fr32::write_unpadded;
 use std::path::Path;
 use storage_proofs::circuit::multi_proof::MultiProof;
@@ -31,6 +31,7 @@ use storage_proofs::zigzag_drgporep::ZigZagDrgPoRep;
 use storage_proofs::zigzag_graph::ZigZagBucketGraph;
 
 use crate::error;
+use sector_base::api::sector_store::PaddedBytesAmount;
 
 type Commitment = Fr32Ary;
 type ChallengeSeed = Fr32Ary;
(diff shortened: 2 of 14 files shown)


schomatis:
I'm not familiar with some of the components being worked on here but I can give this PR a first pass of reviews before someone more experienced takes over.

laser:
Excellent. Thanks, Lucas. I’m curious to learn how you feel about the two newtypes which I added.

filecoin-proofs/examples/ffi/main.rs
```diff
@@ -312,6 +313,50 @@ unsafe fn sector_builder_lifecycle(use_live_store: bool) -> Result<(), Box<Error
...
+            (*resp).faults_len,
+        );
+        defer!(destroy_verify_post_response(resp));
+
+        if (*resp).status_code != 0 {
+            panic!("{}", c_str_to_rust_str((*resp).error_msg))
+        }
+
+        assert!((*resp).is_valid)
+    }
```
schomatis: From a non-Rustacean perspective these big scopes within this long functions feel like they should actually be functions of their own. (Would that help simplify all this `defer!` logic?)

filecoin-proofs/examples/ffi/main.rs
```diff
@@ -312,6 +313,50 @@ unsafe fn sector_builder_lifecycle(use_live_store: bool) -> Result<(), Box<Error
...
+        defer!(destroy_get_sealed_sectors_response(resp));
+
+        if (*resp).status_code != 0 {
+            panic!("{}", c_str_to_rust_str((*resp).error_msg))
+        }
+
+        let sealed_sector_metadata: FFISealedSectorMetadata =
+            from_raw_parts((*resp).sectors_ptr, (*resp).sectors_len)[0];
+        let sealed_sector_replica_commitment: [u8; 32] = sealed_sector_metadata.comm_r;
+        let challenge_seed: [u8; 32] = [0; 32];
```
schomatis: Where is this seed set? Or are we purposefully setting it to zero?

filecoin-proofs/examples/ffi/main.rs
```diff
@@ -312,6 +313,50 @@ unsafe fn sector_builder_lifecycle(use_live_store: bool) -> Result<(), Box<Error
...
+
+        let resp = generate_post(
+            sector_builder_b,
+            &sealed_sector_replica_commitment[0],
+            32,
+            &challenge_seed,
+        );
+        defer!(destroy_generate_post_response(resp));
```
schomatis: nit (feel free to ignore): up to now I was find reusing the `resp` name but looking at the "destructor" I now realize that those are different types of responses, maybe that could be reflected on the names so we're sure we're calling the correct destructor.

filecoin-proofs/src/api/internal.rs
```diff
@@ -200,24 +212,11 @@ pub struct PoStInput {
...
-    })
-}
-
-pub fn generate_post(sector_bytes: u64, input: PoStInput) -> error::Result<PoStOutput> {
+pub fn generate_post(sector_config: &SectorConfig, input: PoStInput) -> error::Result<PoStOutput> {
     let faults: Vec<u64> = Vec::new();
 
     let setup_params = compound_proof::SetupParams {
-        vanilla_params: &post_setup_params(sector_bytes as usize),
+        vanilla_params: &post_setup_params(input.input_parts.len(), sector_config.sector_bytes()),
```
schomatis: If I'm following this correctly, the number of sectors is the number of `PoStInputPart`s, right?

filecoin-proofs/src/api/internal.rs
```diff
@@ -262,7 +265,7 @@ pub fn generate_post(sector_bytes: u64, input: PoStInput) -> error::Result<PoStO
 
     let priv_inputs = vdf_post::PrivateInputs::<PedersenHasher>::new(&borrowed_trees[..]);
 
-    let groth_params = get_post_params(sector_bytes as usize)?;
+    let groth_params = get_post_params(borrowed_trees.len(), sector_config.sector_bytes())?;
```
schomatis: And here also, the tree is derived from `input_parts`.

filecoin-proofs/src/api/internal.rs
```diff
@@ -295,7 +298,7 @@ pub fn verify_post(
     };
 
     let compound_setup_params = compound_proof::SetupParams {
-        vanilla_params: &post_setup_params(sector_bytes as usize),
+        vanilla_params: &post_setup_params(comm_rs.len(), sector_config.sector_bytes()),
```
schomatis: Also, number of sectors, but derived from the commitments, this is a bit tricky to follow at times but I guess those are the only oracles of how many sectors we are working with at the moment the functions are called.

filecoin-proofs/src/api/mod.rs
```diff
@@ -136,68 +136,69 @@ pub unsafe extern "C" fn generate_post(
 ///
 #[no_mangle]
 pub unsafe extern "C" fn verify_post(
-    _flattened_comm_rs_ptr: *const u8,
```
schomatis: I think this is inadvertently reverting the work from https://github.com/filecoin-project/rust-fil-proofs/pull/490.

laser: > I think this is inadvertently reverting the work from #490.

In fact I did want to replace that portion of #490. I want `generate_post` `verify_post` in `mod.rs` to call in to their equivalents in `internal.rs` so that I could test those functions in the `ffi/main.rs` test.

laser:
Talked with @porcuquine at length today. "sectors count" is in fact part of public and setup parameters, but is fixed. If we want to prove a number of sectors which is not equal to "sectors count," which is currently 2, we need to generate multiple PoSts.

That said, this PR is trash. I'll close it and put the newtype stuff in a new PR.

Issue #481, opened 2019-02-15
disk-backed merkle trees


porcuquine:
# Original write-up

### Description

We devote a lot of memory to holding onto merkle trees during replication. We should eliminate this requirement, which limits the sector size we can replicate. The trees should live on disk; they don't need to be in memory between when they are generated and when they are used to create the proof. The *root* of each tree will be needed to calculate `comm_r_star`, so it should probably be extracted and held before flushing to disk, though.

Whether the best solution is to construct the tree in mmapped data, or to just dump and load it from a file is an expedient question for now.

### Acceptance criteria

Ideally, we would see memory profiling evidence before and after.

Alternately, empirical evidence of replicating a sector size which fails for lack of memory before then succeeds after would do.

### Risks + pitfalls

As we consider generalizing our approach to (partial and complete) merkle tree caching, and to sharing the mechanism betwee PoRep and PoSt, the design space becomes somewhat dense. Don't get caught up in trying to solve future problems — but do bear in mind the design trajectory, to the extent you're familiar with it.

### Where to begin

Merkle tree generation: https://github.com/filecoin-project/rust-proofs/blob/master/storage-proofs/src/drgraph.rs#L26-L28

We have all the trees together here: https://github.com/filecoin-project/rust-proofs/blob/master/storage-proofs/src/layered_drgporep.rs#L427-L444

We use all the trees when proving here: https://github.com/filecoin-project/rust-proofs/blob/master/storage-proofs/src/layered_drgporep.rs#L237-L240

What we hold onto and pass around in `aux` should not require a a full merkle-tree's worth of memory, except when the relevant layer's tree is being used to generate the proof: https://github.com/filecoin-project/rust-proofs/blob/master/storage-proofs/src/layered_drgporep.rs#L259-L261

-----------------------------------------------------------------

# Work Log

Most of the current work is devoted to setting up the tools to automate the memory profiling of the `zigzag` example (the application of the `mmap` itself doesn't seem to present much of a problem).

## Memory profiling

[...]

PR #495, opened 2019-02-21
encode: preallocate the vector used in `create_key`

schomatis:
While profiling for https://github.com/filecoin-project/rust-fil-proofs/issues/481 I've found a repetitive allocation during replication, basically for each node in each layer we were allocating a vector with a size `degree()` times bigger.

Before:
```
replication_time: 127.529965113s, target: stats, place: filecoin-proofs/examples/zigzag.rs:164 zigzag, root: filecoin-proofs
replication_time/byte: 3.799µs, target: stats, place: filecoin-proofs/examples/zigzag.rs:165 zigzag, root: filecoin-proofs
replication_time/GiB: 4080.958883615s, target: stats, place: filecoin-proofs/examples/zigzag.rs:170 zigzag, root: filecoin-proofs

```

After:
```
replication_time: 139.750112713s, target: stats, place: filecoin-proofs/examples/zigzag.rs:164 zigzag, root: filecoin-proofs
replication_time/byte: 4.164µs, target: stats, place: filecoin-proofs/examples/zigzag.rs:165 zigzag, root: filecoin-proofs
replication_time/GiB: 4472.003606815s, target: stats, place: filecoin-proofs/examples/zigzag.rs:170 zigzag, root: filecoin-proofs
```

The results are slightly worse but I don't think it's related to the patch.

### DIFF
--- a/storage-proofs/src/vde.rs
+++ b/storage-proofs/src/vde.rs
@@ -24,6 +24,10 @@ where
     // The only subtlety is that a ZigZag graph may be reversed, so the direction
     // of the traversal must also be.
 
+    let mut ciphertexts = vec![0u8; 32 * (graph.degree() + 1)];
+    // Preallocate the KDF vector where the parents of the current
+    // node `n` will be loaded to be hashed.
+
     for n in 0..graph.size() {
         let node = if graph.forward() {
             n
@@ -35,7 +39,14 @@ where
         let parents = graph.parents(node);
         assert_eq!(parents.len(), graph.degree(), "wrong number of parents");
 
-        let key = create_key::<H>(replica_id, node, &parents, data, degree)?;
+        let key = create_key::<H>(
+            replica_id,
+            node,
+            &parents,
+            data,
+            degree,
+            Some(&mut ciphertexts),
+        )?;
         let start = data_at_node_offset(node);
         let end = start + 32;
 
@@ -79,7 +90,7 @@ where
     G: Graph<H>,
 {
     let parents = graph.parents(v);
-    let key = create_key::<H>(replica_id, v, &parents, &data, graph.degree())?;
+    let key = create_key::<H>(replica_id, v, &parents, &data, graph.degree(), None)?;
     let node_data = H::Domain::try_from_bytes(&data_at_node(data, v)?)?;
 
     // TODO: round constant
@@ -104,7 +115,7 @@ where
         .flat_map(H::Domain::into_bytes)
         .collect::<Vec<u8>>();
 
-    let key = create_key::<H>(replica_id, v, &parents, &byte_data, graph.degree())?;
+    let key = create_key::<H>(replica_id, v, &parents, &byte_data, graph.degree(), None)?;
     let node_data = data[v];
 
     // TODO: round constant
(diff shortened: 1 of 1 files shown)


storage-proofs/src/vde.rs
```diff
@@ -117,11 +128,20 @@ fn create_key<H: Hasher>(
...
     // ciphertexts will become a buffer of the layout
     // id | encodedParentNode1 | encodedParentNode1 | ...
 
-    let mut ciphertexts = vec![0u8; 32 * (parents.len() + 1)];
+    let mut owned_vector: Vec<u8>;
+    let ciphertexts: &mut Vec<u8> = match ciphertexts {
+        Some(preallocated) => preallocated,
+        None => {
+            owned_vector = vec![0u8; 32 * (parents.len() + 1)];
+            owned_vector.as_mut()
```
schomatis: Please tell me there's a cleaner way to do this.

dignifiedquire: this is already pretty good, though I would use `&mut [u8]` instead of `Vec`. One nicer alternative would be to reuse this scratch memory across calls using `thread_local`. You can see an example of this here: 

- https://github.com/dignifiedquire/num-bigint/blob/div/src/algorithms/div.rs#L147-L149
- https://github.com/dignifiedquire/num-bigint/blob/div/src/algorithms/div.rs#L217

porcuquine:
First comment based only on the benchmark results. The after times are much worse than the before, so we definitely wouldn't want to introduce changes including that evidence. Can you re-run the before/after benchmarks in a way which addresses whatever you believe the actual root cause of the difference is? I'm not sure what your hypothesis is, but we should see verify it and see evidence of improvement (or at least a lack of evidence of regression) before accepting these changes.

To be clear: I haven't looked at the changes yet — so my comments above just relate to the benchmark results — which I don't think can be discounted without further investigation.

schomatis:
Ok, will apply @dignifiedquire suggestions first and do a more in-depth profile. I'll try to use the `rust-proofs-infra` repo in the future to avoid this error-prone benchmarking method.

dignifiedquire:
@schomatis what's the state of this?

schomatis:
@dignifiedquire Blocked until I refine the benchmarking process enough to not be a time sink, progress in  https://github.com/filecoin-project/rust-proofs-infra/issues/17. @porcuquine is right, I should be able to explain this 10% increase, but first I'd like to confirm it's actually that.

dignifiedquire:
I'll run it locally as well if I see the impact, that should help give a direction of how reproducible this is.

schomatis:
> I'll run it locally as well if I see the impact, that should help give a direction of how reproducible this is.

Thanks but don't bother, this is not a priority (unless you think otherwise), it's more important (at least for my work) to streamline the benchmarking process anyway so I spend less time with that than with the actual coding (and I think this can wait after that is solved to benefit from it).

porcuquine:
@schomatis I support this — especially if the outcome is a clear, reproducible process for other developers to do the same.

schomatis:
> @schomatis I support this — especially if the outcome is a clear, reproducible process for other developers to do the same.

Yes, that's the motivation behind the (WIP) Docker image: https://hub.docker.com/r/schomatis/rust-fil-proofs-profile.

schomatis:
This is already a (small but valid) win on itself, so let's get this merged and leave the MT-internals work for https://github.com/filecoin-project/rust-fil-proofs/pull/529.

dignifiedquire:
Superseeded by #532, @schomatis I think we can close this PR

PR #508, opened 2019-02-26
verify_post should be passed cfg + add FFI tests for PoSt generation/verification

laser:
Fixes go-filecoin `#2095`, which you can [read all about here](https://app.zenhub.com/workspaces/filecoin-5ab0036a12f8e82ae4ed60f0/issues/filecoin-project/go-filecoin/2095).

## Why does this PR exist?

The FFI-exposed `verify_post` function used to accept a u64 which represented sealed sector bytes. go-filecoin had no way of getting sealed sector-bytes from its sector builder, so this API wouldn't have worked without modifying the sector builder interface.

## What's in this PR?

- `verify_post` to accept config value (like `verify_seal`)
- add FFI tests which exercise PoSt verification and generation
- modify `paramcache` to accept two optional flags: `--test-only` and `--include-post` (omits Live parameters and includes PoSt parameters, respectively)

## `paramcache` mods

`paramcache` now accepts two optional flags which are used to control Groth parameter cache-populating behavior.

By default, `paramcache` will generate PoRep parameters for both Live and Test configs.

If the `--include-post` flag is passed, `paramcache` will (in addition to PoRep) generate PoSt parameters. *Once PoSt goes live in go-filecoin, we will remove this flag; we will always generate PoSt and PoRep parameters from that point onwards.*

If the `--test-only` flag is passed, `paramcache` will only generate parameters for Test config.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -96,7 +96,7 @@ jobs:
       - run:
           name: Ensure cache is hydrated with PoRep and PoSt Groth parameters (for test)
           command: |
-            cargo +stable run --release --color=always --package filecoin-proofs --bin paramcache -- --include-post --test-only
+            cargo run --release --color=always --package filecoin-proofs --bin paramcache -- --include-post --test-only
       - run:
           name: run regression tests (examples) against FFI-exposed filecoin-proofs API
           command: RUSTFLAGS="-Z sanitizer=leak" cargo run --release --package filecoin-proofs --example ffi --target x86_64-unknown-linux-gnu
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -75,6 +75,10 @@ jobs:
             - cargo-v7-{{ checksum "rust-toolchain" }}-{{ checksum "Cargo.toml" }}-{{ checksum "Cargo.lock" }}-{{ arch }}
             - parameter-cache-{{ .Revision }}
       - run:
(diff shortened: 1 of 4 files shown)


porcuquine:
@laser Did you forget to commit a file? I don't see the actual changes, just the ffi test — which is failing on CI.

laser:
I see that the PoSt parameters are lazily generating now (unsurprisingly; I doubt PoSt verify is called in any upstream test quite yet), which is likely the culprit for the test failure. 

I will investigate tomorrow AM.

porcuquine:
This reminds me that once you are actually calling `internal::generate_post` and `internal::verify_post` you will want to ensure that PoSt params are pre-generated by `paramcache` and published for `paramfetch`. The latter will require running `parampublish` *after* generating params locally.

You can do that by setting `GENERATE_POST_PARAMS` to true in `paramcache` — or probably, at this point, just getting rid of the flag and always generating.

.circleci/config.yml
```diff
@@ -93,12 +93,17 @@ jobs:
           keys:
             - cargo-v7-{{ checksum "rust-toolchain" }}-{{ checksum "Cargo.toml" }}-{{ checksum "Cargo.lock" }}-{{ arch }}
             - parameter-cache-{{ .Revision }}
-      - restore_cache:
-          keys:
-            - parameter-cache-{{ .Revision }}
```
laser: note that this line is redundant with line 95, above

Issue #505, opened 2019-02-25
FFI-exposed generate_post must accommodate a fixed VDF PoSt sectors count


laser:
### Description

For some background, [see this issue](https://github.com/filecoin-project/rust-fil-proofs/issues/493).

Subsequent discussion as to how to handle this (in code) was had [here](https://filecoinproject.slack.com/archives/CEHHJNJS3/p1551114905265400).

Until (if) PoSt can handle a dynamic "sectors count," we are stuck generating a PoSt over a fixed number of sectors (currently 2). For each PoSt, a single SNARK will be produced.

We need to modify the FFI-exposed `generate_post` function to return a variable-number of 192-byte SNARK proofs. The number of SNARK proofs returned will be equal to `ceil(count(comm_rs)/2)`.

### Acceptance criteria

1. go-filecoin can generate a proof-of-spacetime through the FFI for a number of sectors which is not equal to the rust-fil-proofs hardcoded "sectors count."

### Risks + pitfalls

1. Each pair of CommRs will correspond to one call to the `VDFPostCompound::prove` function.
1. If we have an odd-number of CommRs, the lonely CommR will be duplicated.
1. FFI-exposed `generate_post` will return a variable-length array of faults. The array should contain the indexes in the CommRs array passed as input which were faulty. Getting this right (given the CommR-duplicating behavior above) might be a little tricky.

PR #520, opened 2019-03-01
adapter layer to accommodate fixed "sectors count" in VDF PoSt

laser:
Fixes the rust-fil-proofs side of #505. Note that the go-filecoin CGO code will need to integrate with this new API and the `submitPoSt` message will need to be modified to accommodate the variable number of SNARK proof outputs. That's next on my list of things to do.

## Why does this PR exist?

Our current VDF PoSt construction allows us to prove a fixed number of sectors (that number is 2, or `POST_SECTORS_COUNT`). API consumers don't know about this; they want to provide a variable number of replica commitments when proving and receive a proof in return.

To provide the API that our consumers want, we need to adapt PoSt generation and verification FFI function calls (which accept a variable number of replica commitments) to our internal PoSt generation and verification operations (which accept exactly 2 replica commitments).

## What's in this PR?

This changeset provides adapters to map a single PoSt generation or verification call (for a variable number of sectors) into multiple PoSt generation calls or verification calls (for a fixed number of sectors). 

This changeset also includes functions which collect the result of multiple PoSt generation and verificatoin calls (for a fixed number of sectors) into a single PoSt generation or verification call (for a variable number of sectors).

The design of the adapter layer is such that:

1) We can change `POST_SECTORS_COUNT` and everything will work normally.
2) Moving from a world in which we have a fixed sectors count to one in which it is not fixed is as simple as deleting the adapter.

## API Changes

The FFI-exposed `generate_post` call used to produce a single SNARK proof. Now it produces `CEIL(COMM_RS / POST_SECTORS_COUNT)` proofs.

## Tricky bits

We need to ensure that `LEN(COMM_RS) % POST_SECTORS_COUNT == 0` due to our VDF PoSt construction ("sectors count" is part of public parameters). To achieve this when `LEN(COMM_RS) % POST_SECTORS_COUNT != 0`, we duplicate the last COMM_R. We do this n-number of times until `LEN(COMM_RS) % POST_SECTORS_COUNT == 0`.

Mapping fault indices *from* the API to faults in the fixed sectors-count world is a little tricky due because of the way that we duplicate CommRs to ensure that `LEN(COMM_RS) % POST_SECTORS_COUNT == 0`. We also need to ensure that we don't return through the FFI fault indices for replicas which we duplicated.

### DIFF
--- a/filecoin-proofs/src/api/mod.rs
+++ b/filecoin-proofs/src/api/mod.rs
@@ -136,15 +136,30 @@ pub unsafe extern "C" fn generate_post(
 ///
 #[no_mangle]
 pub unsafe extern "C" fn verify_post(
-    cfg_ptr: *const ConfiguredStore,
-    flattened_comm_rs_ptr: *const u8,
-    flattened_comm_rs_len: libc::size_t,
-    challenge_seed: &[u8; 32],
+    _cfg_ptr: *const ConfiguredStore,
+    _flattened_comm_rs_ptr: *const u8,
+    _flattened_comm_rs_len: libc::size_t,
+    _challenge_seed: &[u8; 32],
     proofs_ptr: *const [u8; 192],
     proofs_len: libc::size_t,
-    faults_ptr: *const u64,
-    faults_len: libc::size_t,
+    _faults_ptr: *const u64,
+    _faults_len: libc::size_t,
 ) -> *mut responses::VerifyPoSTResponse {
+    let mut response: responses::VerifyPoSTResponse = Default::default();
+
+    let proofs: &[[u8; 192]] = from_raw_parts(proofs_ptr, proofs_len);
+
+    if proofs[0][0] == 42 {
+        response.is_valid = true;
+    } else {
+        response.is_valid = false;
+    };
+
+    // Stay mocked for now — remove early return when ready to use.
+    Box::into_raw(Box::new(response))
+
+    /*
+
     let mut response: responses::VerifyPoSTResponse = Default::default();
 
     if let Some(cfg) = cfg_ptr.as_ref() {
@@ -196,6 +211,8 @@ pub unsafe extern "C" fn verify_post(
     }
 
     Box::into_raw(Box::new(response))
+
+    */
 }
 
 /// Initializes and returns a SectorBuilder.
--- a/filecoin-proofs/src/api/mod.rs
+++ b/filecoin-proofs/src/api/mod.rs
@@ -1,4 +1,4 @@
-use crate::api::internal::PoStOutput;
+use crate::api::post_adapter::*;
 use crate::api::responses::err_code_and_msg;
 use crate::api::responses::FCPResponseStatus;
 use crate::api::responses::FFIPieceMetadata;
@@ -7,15 +7,17 @@ use crate::api::sector_builder::metadata::SealStatus;
 use crate::api::sector_builder::SectorBuilder;
 use ffi_toolkit::rust_str_to_c_str;
 use ffi_toolkit::{c_str_to_rust_str, raw_ptr};
-use libc;
 use sector_base::api::disk_backed_storage::new_sector_config;
 use sector_base::api::disk_backed_storage::ConfiguredStore;
+
+use libc;
 use std::ffi::CString;
 use std::mem;
 use std::ptr;
 use std::slice::from_raw_parts;
 
 pub mod internal;
+pub mod post_adapter;
 pub mod responses;
 mod sector_builder;
 
@@ -94,7 +96,7 @@ pub unsafe extern "C" fn generate_post(
     flattened_comm_rs_ptr: *const u8,
     flattened_comm_rs_len: libc::size_t,
     challenge_seed: &[u8; 32],
-) -> *mut responses::GeneratePoSTResponse {
+) -> *mut responses::GeneratePoStResponse {
     let comm_rs = from_raw_parts(flattened_comm_rs_ptr, flattened_comm_rs_len)
         .iter()
         .step_by(32)
@@ -106,15 +108,13 @@ pub unsafe extern "C" fn generate_post(
             acc
         });
 
-    let mut response: responses::GeneratePoSTResponse = Default::default();
+    let mut response: responses::GeneratePoStResponse = Default::default();
 
     match (*ptr).generate_post(&comm_rs, challenge_seed) {
-        Ok(PoStOutput {
-            snark_proof,
-            faults,
-        }) => {
+        Ok(GeneratePoStDynamicSectorsCountOutput { proofs, faults }) => {
             response.status_code = FCPResponseStatus::FCPNoError;
-            response.proof = snark_proof;
+            response.proofs_len = proofs.len();
+            response.proofs_ptr = proofs.as_ptr();
 
             response.faults_len = faults.len();
             response.faults_ptr = faults.as_ptr();
@@ -136,68 +136,66 @@ pub unsafe extern "C" fn generate_post(
 ///
 #[no_mangle]
 pub unsafe extern "C" fn verify_post(
-    _cfg_ptr: *const ConfiguredStore,
-    _flattened_comm_rs_ptr: *const u8,
-    _flattened_comm_rs_len: libc::size_t,
-    _challenge_seed: &[u8; 32],
-    proof: &[u8; API_POST_PROOF_BYTES],
-    _faults_ptr: *const u64,
-    _faults_len: libc::size_t,
+    cfg_ptr: *const ConfiguredStore,
+    flattened_comm_rs_ptr: *const u8,
+    flattened_comm_rs_len: libc::size_t,
+    challenge_seed: &[u8; 32],
+    proofs_ptr: *const [u8; 192],
+    proofs_len: libc::size_t,
+    faults_ptr: *const u64,
+    faults_len: libc::size_t,
 ) -> *mut responses::VerifyPoSTResponse {
     let mut response: responses::VerifyPoSTResponse = Default::default();
 
-    if proof[0] == 42 {
-        response.is_valid = true;
+    if let Some(cfg) = cfg_ptr.as_ref() {
+        let cfg = new_sector_config(cfg);
+
+        let comm_rs = from_raw_parts(flattened_comm_rs_ptr, flattened_comm_rs_len)
+            .iter()
+            .step_by(32)
+            .fold(Default::default(), |mut acc: Vec<[u8; 32]>, item| {
+                let sliced = from_raw_parts(item, 32);
+                let mut x: [u8; 32] = Default::default();
+                x.copy_from_slice(&sliced[..32]);
+                acc.push(x);
+                acc
+            });
+
+        let faults = from_raw_parts(faults_ptr, faults_len);
+
+        let safe_challenge_seed = {
+            let mut cs = [0; 32];
+            cs.copy_from_slice(challenge_seed);
+            cs[31] &= 0b00111111;
+            cs
+        };
+
+        match internal::verify_post(VerifyPoStDynamicSectorsCountInput {
+            sector_bytes: cfg.sector_bytes(),
+            comm_rs,
+            challenge_seed: safe_challenge_seed,
+            proofs: from_raw_parts(proofs_ptr, proofs_len).to_vec(),
+            faults: faults.to_vec(),
+        }) {
+            Ok(dynamic) => {
+                response.status_code = FCPResponseStatus::FCPNoError;
+                response.is_valid = dynamic.is_valid;
+            }
+            Err(err) => {
+                let (code, ptr) = err_code_and_msg(&err);
+                response.status_code = code;
+                response.error_msg = ptr;
+            }
+        }
     } else {
-        response.is_valid = false;
-    };
+        response.status_code = FCPResponseStatus::FCPCallerError;
 
-    // Stay mocked for now — remove early return when ready to use.
+        let msg = CString::new("caller did not provide ConfiguredStore").unwrap();
+        response.error_msg = msg.as_ptr();
+        mem::forget(msg);
+    }
+
     Box::into_raw(Box::new(response))
-
-    // let comm_rs = from_raw_parts(flattened_comm_rs_ptr, flattened_comm_rs_len)
-    //     .iter()
-    //     .step_by(32)
-    //     .fold(Default::default(), |mut acc: Vec<[u8; 32]>, item| {
-    //         let sliced = from_raw_parts(item, 32);
-    //         let mut x: [u8; 32] = Default::default();
-    //         x.copy_from_slice(&sliced[..32]);
-    //         acc.push(x);
-    //         acc
-    //     });
-
-    // let faults = from_raw_parts(faults_ptr, faults_len);
-
-    // let safe_challenge_seed = {
-    //     let mut cs = [0; 32];
-    //     cs.copy_from_slice(challenge_seed);
-    //     cs[31] &= 0b00111111;
-    //     cs
-    // };
-
-    // match internal::verify_post(
-    //     sector_bytes,
-    //     &comm_rs,
-    //     &safe_challenge_seed,
-    //     proof,
-    //     faults.to_vec(),
-    // ) {
-    //     Ok(true) => {
-    //         response.status_code = FCPResponseStatus::FCPNoError;
-    //         response.is_valid = true;
-    //     }
-    //     Ok(false) => {
-    //         response.status_code = FCPResponseStatus::FCPNoError;
-    //         response.is_valid = false;
-    //     }
-    //     Err(err) => {
-    //         let (code, ptr) = err_code_and_msg(&err);
-    //         response.status_code = code;
-    //         response.error_msg = ptr;
-    //     }
-    // }
-
-    // Box::into_raw(Box::new(response))
 }
 
 /// Initializes and returns a SectorBuilder.
--- a/filecoin-proofs/src/api/post_adapter.rs
+++ b/filecoin-proofs/src/api/post_adapter.rs
@@ -95,9 +95,11 @@ pub fn generate_post_spread_input(
 }
 
 /// Collapses return values from multiple fixed sectors-count PoSt generation
-/// operations into a single return value for dynamic sector count.
+/// operations into a single return value for dynamic sector count. This
+/// function discards any faults reported for duplicated CommRs.
 ///
 pub fn generate_post_collect_output(
+    orig_comm_rs_len: usize,
     xs: Vec<error::Result<GeneratePoStFixedSectorsCountOutput>>,
 ) -> error::Result<GeneratePoStDynamicSectorsCountOutput> {
     if xs.is_empty() {
@@ -122,6 +124,8 @@ pub fn generate_post_collect_output(
             }
         }
     }
+
+    dynamic.faults = dynamic.faults.into_iter().take(orig_comm_rs_len).collect();
 
     Ok(dynamic)
 }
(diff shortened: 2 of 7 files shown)


filecoin-proofs/src/api/post_adapter.rs
```diff
@@ -0,0 +1,626 @@
...
+    pub proofs: Vec<[u8; 192]>,
+    pub faults: Vec<u64>,
+}
+
+pub struct VerifyPoStFixedSectorsCountInput {
+    pub sector_bytes: PaddedBytesAmount,
+    pub comm_rs: [Commitment; POST_SECTORS_COUNT],
+    pub challenge_seed: ChallengeSeed,
+    pub proof: [u8; 192],
+    pub faults: Vec<u64>,
```
laser: For maximum type safety, the type of this field should be `[bool; POST_SECTORS_COUNT]`. I'll change that in a future PR.

filecoin-proofs/src/api/post_adapter.rs
```diff
@@ -0,0 +1,626 @@
...
+        });
+    }
+
+    if !remainder.is_empty() {
+        let mut input_parts: [(Option<String>, Commitment); POST_SECTORS_COUNT] =
+            Default::default();
+
+        let vs = repeat_with(|| remainder[remainder.len() - 1].clone());
+        let rs = remainder.iter().cloned();
+        let it = rs.chain(vs).take(POST_SECTORS_COUNT).enumerate();
```
schomatis: I get what's going on here, but `vs`/`rs`/`it` don't carry much meaning within my context (except `it` which I suppose it's "iterator").

schomatis: Also, I'm learning a lot of Rust here :smiley:

schomatis: So, If I understand correctly, we repeat the last commitment as many times as necessary to fill `POST_SECTORS_COUNT`, that may be something worth clarifying somewhere. From the PR description I (arbitrarily) assumed that we duplicated everything, e.g., if `POST_SECTORS_COUNT` was 4 and I had `Comm_0` and `Comm_1` I would get `[Comm_0, Comm_1, Comm_0, Comm_1]` but actually it would be `[Comm_0, Comm_1, Comm_1, Comm_1]`. My assumption might be due because the example was using the basic (and current) sector count of 2, maybe a more elaborate example could help clarify that (if we indeed may change the count in the future, if not, disregard this entire comment).

filecoin-proofs/src/api/post_adapter.rs
```diff
@@ -0,0 +1,626 @@
...
+
+    for (i, x) in xs.into_iter().enumerate() {
+        match x {
+            Err(e) => {
+                return Err(e);
+            }
+            Ok(fixed) => {
+                dynamic.proofs.push(fixed.proof);
+                for fault in fixed.faults.iter() {
+                    dynamic.faults.push(((i as u64) * 2) + fault)
```
schomatis: Not sure what's going on here, where did the `2` come from?

filecoin-proofs/src/api/post_adapter.rs
```diff
@@ -0,0 +1,626 @@
...
+                }
+            }
+        }
+    }
+
+    dynamic.faults = dynamic.faults.into_iter().take(orig_comm_rs_len).collect();
```
schomatis: We are putting a lot of trust in the caller here (both in terms of `orig_comm_rs_len` and `fixed.faults` having always the same number of items) which may be acceptable, but should be more documented I think.

filecoin-proofs/src/api/post_adapter.rs
```diff
@@ -0,0 +1,626 @@
...
+    pub sector_bytes: PaddedBytesAmount,
+    pub comm_rs: Vec<Commitment>,
+    pub challenge_seed: ChallengeSeed,
+    pub proofs: Vec<[u8; 192]>,
+    pub faults: Vec<u64>,
+}
+
+pub struct VerifyPoStFixedSectorsCountInput {
+    pub sector_bytes: PaddedBytesAmount,
+    pub comm_rs: [Commitment; POST_SECTORS_COUNT],
```
schomatis: Similar as before, since there are common (unchanged) parameters here it's harder to grasp the difference between fixed and dynamic: only `comm_rs` and `faults` changing form `Vec` to fixed array.

filecoin-proofs/src/api/post_adapter.rs
```diff
@@ -0,0 +1,626 @@
...
+
+    if faults_len > commrs_len {
+        return Err(format_err!(
+            "LEN(faults) must <= LEN(comm_rs): {:?}, {:?}",
+            faults_len,
+            commrs_len
+        ));
+    }
+
+    if proofs_len != replicas_c {
```
schomatis: These types of checks (that don't trust the caller) might also be useful in the `generate_` counterparts of these functions.

filecoin-proofs/src/api/post_adapter.rs
```diff
@@ -0,0 +1,626 @@
...
+        .unwrap_or(false)
+    {
+        return Err(format_err!(
+            "MAX(faults) must <= LEN(comm_rs)-1: {:?}, {:?}",
+            faults_max,
+            commrs_len - 1
+        ));
+    }
+
+    let mut fixed: Vec<VerifyPoStFixedSectorsCountInput> = vec![];
```
schomatis: Same pre-allocation comment as above.

filecoin-proofs/src/api/post_adapter.rs
```diff
@@ -0,0 +1,626 @@
...
+
+    let remainder = chunks.remainder();
+
+    // each POST_SECTORS_COUNT grouping of commitments maps to a single fixed
+    // sectors-count PoSt verification call
+    for (i, chunk) in chunks.enumerate() {
+        let mut comm_rs: [Commitment; POST_SECTORS_COUNT] = [[0; 32]; POST_SECTORS_COUNT];
+        for (i, comm_r) in chunk.iter().enumerate() {
+            comm_rs[i] = *comm_r;
+        }
```
schomatis: (nit:) This isn't something that actually started in this PR, but my feedback as a newbie here (and keeping in mind that we're opening this repo to everyone now) is that this `comm_rs`/`comm_r` (`chunks`/`chunk`) terminology is a bit hard to reason about.

filecoin-proofs/src/api/post_adapter.rs
```diff
@@ -0,0 +1,626 @@
...
+            challenge_seed: dynamic.challenge_seed,
+            proof: dynamic.proofs[i],
+            faults: Vec::new(),
+        })
+    }
+
+    // If we receive a number of comm_rs which does not divide evenly by
+    // POST_SECTORS_COUNT, form a grouping of comm_rs which does divide
+    // evenly by POST_SECTORS_COUNT by duplicating the last comm_r in the
+    // group until LEN(comm_rs) == POST_SECTORS_COUNT.
```
schomatis: Yes, this is exactly the type of comment I had in mind that would help to understand the repetition format in `generate_`.

filecoin-proofs/src/api/post_adapter.rs
```diff
@@ -0,0 +1,626 @@
...
+            comm_rs,
+            challenge_seed: dynamic.challenge_seed,
+            proof: dynamic.proofs[proofs_len - 1],
+            faults: Vec::new(),
+        })
+    }
+
+    // Map the dynamic sectors count indices to the fixed sectors count indices.
+    for fault in dynamic.faults {
+        let i = (fault as f64 / POST_SECTORS_COUNT as f64).floor();
```
schomatis: Ok, I see now that the previous `2` was actually a reference to `POST_SECTORS_COUNT`.

filecoin-proofs/src/api/post_adapter.rs
```diff
@@ -0,0 +1,626 @@
...
+        fixed[i as usize]
```
schomatis: As in `generate_`, I can't thoroughly follow this logic because I don't know the pattern of the `fault` values (what confuses me is that we seem to be mixing `fault` values with the index we're computing to store it).

filecoin-proofs/src/api/post_adapter.rs
```diff
@@ -0,0 +1,626 @@
...
+
+    // If we had to duplicate commitments to make LEN(COMMITMENTS) ==
+    // POST_SECTORS_COUNT and the thing we duplicated was faulty, ensure that
+    // the duplicates are marked as faulty, too.
+    if !remainder.is_empty() {
+        let fixed_len = { fixed.len() - 1 };
+        let remdr_len = { remainder.len() };
+
+        let last_remainder_fault: Option<u64> =
+            { fixed[fixed_len].faults.get(remdr_len - 1).cloned() };
```
schomatis: Will the next to last `faults` always have `remdr_len - 1` items here? (just asking since I didn't quite followed the `faults` generation logic)

filecoin-proofs/src/api/post_adapter.rs
```diff
@@ -0,0 +1,626 @@
...
+
+    let mut dynamic = VerifyPoStDynamicSectorsCountOutput { is_valid: true };
+
+    for x in xs.into_iter() {
+        match x {
+            Err(e) => {
+                return Err(e);
+            }
+            Ok(VerifyPoStFixedSectorsCountOutput { is_valid }) => {
+                dynamic.is_valid = dynamic.is_valid && is_valid;
```
schomatis: Seems we can just stop here and return a `false` the first time it happens.

filecoin-proofs/src/api/post_adapter.rs
```diff
@@ -0,0 +1,626 @@
...
+
+    let chunks = dynamic.input_parts.chunks_exact(POST_SECTORS_COUNT);
+
+    let remainder = chunks.remainder();
+
+    for chunk in chunks {
+        fixed.push(GeneratePoStFixedSectorsCountInput {
+            sector_bytes: dynamic.sector_bytes,
+            challenge_seed: dynamic.challenge_seed,
+            input_parts: [chunk[0].clone(), chunk[1].clone()],
```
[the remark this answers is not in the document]
laser: Great catch. How embarrassing! Updated.

filecoin-proofs/src/api/post_adapter.rs
laser: That's a bug. The `2` should be `POST_SECTORS_COUNT`.

> Not sure what's going on here

This awkward code maps values in the fixed-output `faults` vectors to their equivalents in the dynamic-outputs `faults` vectors. 

The formula is this: `dynamic fault value = static fault value + (static output index * POST_SECTORS_COUNT`.

I dunno if this helps, but here's an attempt at a digram:

```
static_outputs = [
  GeneratePoStFixedSectorsCountOutput(faults=[]), 
  GeneratePoStFixedSectorsCountOutput(faults=[0]),
  GeneratePoStFixedSectorsCountOutput(faults=[1])
]

dynamic_outputs = generate_post_collect_output(static_outputs)
// GeneratePoStDynamicSectorsCountOutput(faults=[2, 5])
```

filecoin-proofs/src/api/post_adapter.rs
laser: You and me both!

filecoin-proofs/src/api/mod.rs
```diff
@@ -155,49 +158,61 @@ pub unsafe extern "C" fn verify_post(
...
+        response.status_code = FCPResponseStatus::FCPCallerError;
+
+        let msg = CString::new("caller did not provide ConfiguredStore").unwrap();
+        response.error_msg = msg.as_ptr();
+        mem::forget(msg);
+    }
+
+    Box::into_raw(Box::new(response))
+
+    */
```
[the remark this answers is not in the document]
laser: I'm going to make the commented-out portion real. (In my next PR)

filecoin-proofs/src/api/post_adapter.rs
laser: I rewrote the reduction in a functional style. The enclosing function now has a single exit point instead of multiple exit points, which I think makes it clearer. Thanks for calling this out.

filecoin-proofs/src/api/post_adapter.rs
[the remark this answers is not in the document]
laser: Fixed. No longer cloning; I use `into_iter()` instead.

filecoin-proofs/src/api/post_adapter.rs
```diff
@@ -0,0 +1,626 @@
...
+
+pub struct GeneratePoStDynamicSectorsCountInput {
+    pub sector_bytes: PaddedBytesAmount,
+    pub challenge_seed: ChallengeSeed,
+    pub input_parts: Vec<(Option<String>, Commitment)>,
+}
+
+pub struct GeneratePoStFixedSectorsCountInput {
+    pub sector_bytes: PaddedBytesAmount,
+    pub challenge_seed: ChallengeSeed,
```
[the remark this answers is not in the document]
laser: > do we want to encapsulate them in separate structures?

We could do that, but I found the resulting structures to be a little awkward to use. Duplicating `sector_bytes` (an unsigned 64-bit integer) and `challenge_seed` (a 32-byte array) didn't seem like a big deal to me if it resulted in a less-confusing (subjective, of course) signature.

filecoin-proofs/src/api/post_adapter.rs
laser: > These types of checks (that don't trust the caller) might also be useful in the generate_ counterparts of these functions.

I struggled to dream up invariants for the `generate_` calls that weren't already A) enforced by the type system or B) enforced at runtime by the Rust code I've already written. Can you think of any?

filecoin-proofs/src/api/post_adapter.rs
laser: Good call. I added a little example in the comments, which I'll repaste here:

```

    // For example:
    //
    // dynamic = {faults=[0, 1, 5, 6] and POST_SECTORS_COUNT = 2
    // fixed = [{faults=[0, 1]}, {faults=[]}, {faults=[1]}, {faults=[0]}]
```

filecoin-proofs/src/api/post_adapter.rs
laser: I added an example to the code (in comments), which I'll repaste here:

```
// POST_SECTORS_COUNT = 2
//
// dynamic = {
//   comm_rs=[a, b, c]
//   faults=[2]
// }
//
// fixed = [
//   {comm_rs=[a, b] faults=[]}
//   {comm_rs=[c, c] faults=[0, 1]}
// ]
```

> Will the next to last faults always have remdr_len - 1 items here?

I don't quite understand your question. Would you mind rephrasing for me, please?

filecoin-proofs/src/api/post_adapter.rs
```diff
@@ -0,0 +1,638 @@
...
+    }
+
+    if !remainder.is_empty() {
+        let mut input_parts: [(Option<String>, Commitment); POST_SECTORS_COUNT] =
+            Default::default();
+
+        let iter_with_idx = remainder
+            .iter()
+            .cloned()
+            .chain(repeat_with(|| remainder[remainder.len() - 1].clone()))
```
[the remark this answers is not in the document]
laser: Sure, will do.

filecoin-proofs/src/api/post_adapter.rs
```diff
@@ -0,0 +1,668 @@
...
+
+/// Collapses return values from multiple fixed sectors-count PoSt generation
+/// operations into a single return value for dynamic sector count. This
+/// function discards any faults reported for duplicated CommRs.
+///
+pub fn generate_post_collect_output(
+    orig_comm_rs_len: usize,
+    xs: Vec<error::Result<GeneratePoStFixedSectorsCountOutput>>,
+) -> error::Result<GeneratePoStDynamicSectorsCountOutput> {
+    let seed = if xs.is_empty() {
```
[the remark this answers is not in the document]
laser: Undestood. Will update.

filecoin-proofs/src/api/post_adapter.rs
```diff
@@ -0,0 +1,668 @@
...
+                            .map(|fault| ((i as u64) * POST_SECTORS_COUNT as u64) + fault)
+                            .collect(),
+                    ]
+                    .concat(),
+                })
+            })
+        })
+        .map(|dynamic| GeneratePoStDynamicSectorsCountOutput {
+            proofs: dynamic.proofs,
+            faults: dynamic.faults.into_iter().take(orig_comm_rs_len).collect(),
```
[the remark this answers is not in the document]
laser: Yep, you're correct!

filecoin-proofs/src/api/post_adapter.rs
```diff
@@ -0,0 +1,668 @@
...
+///
+pub fn verify_post_spread_input(
+    dynamic: VerifyPoStDynamicSectorsCountInput,
+) -> error::Result<Vec<VerifyPoStFixedSectorsCountInput>> {
+    let faults_len = dynamic.faults.len();
+    let commrs_len = dynamic.comm_rs.len();
+    let proofs_len = dynamic.proofs.len();
+    let replicas_c = (commrs_len as f64 / POST_SECTORS_COUNT as f64)
+        .ceil()
+        .round() as usize;
```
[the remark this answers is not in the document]
laser: Good catch. Dropping redundant `round`.

filecoin-proofs/src/api/post_adapter.rs
```diff
@@ -0,0 +1,668 @@
...
+        return Err(format_err!(
+            "LEN(faults) must <= LEN(comm_rs): {:?}, {:?}",
+            faults_len,
+            commrs_len
+        ));
+    }
+
+    if proofs_len != replicas_c {
+        return Err(format_err!(
+            "LEN(proofs) must == CEIL(LEN(comm_rs)/2): {:?}, {:?}",
```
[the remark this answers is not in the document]
laser: Yes. That `2` should be `POST_SECTORS_COUNT`.

filecoin-proofs/src/api/post_adapter.rs
```diff
@@ -0,0 +1,668 @@
...
+    // fixed = [
+    //   {comm_rs=[a, b] faults=[]}
+    //   {comm_rs=[c, c] faults=[0, 1]}
+    // ]
+    if !remainder.is_empty() {
+        let fixed_len = { fixed.len() };
+        let remdr_len = { remainder.len() };
+
+        let last_remainder_fault: Option<u64> =
+            { fixed[fixed_len - 1].faults.get(remdr_len - 1).cloned() };
```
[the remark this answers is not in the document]
laser: Negatory. It's a little tricky.

If we enter the `if !remainder.is_empty() {` block, the following things are true:

- The last entry in `fixed` contains at least one duplicated CommR.
- `remdr_len` equals the remainder of `LEN(COMM_RS)/POST_SECTORS_COUNT`
- The entry at `fixed[fixed.len()-1].comm_rs[remdr_len-1]` is equal to `dynamic.comm_rs[dynamic.comm_rs.len()-1]`.

So, that said:

We duplicate `fixed[fixed.len()-1].comm_rs[remdr_len-1]` until `fixed[fixed.len()-1].comm_rs.len() == POST_SECTORS_COUNT`. We need to do the same thing for that (duplicated) commitment's fault (if one exists), too.

filecoin-proofs/src/api/post_adapter.rs
```diff
@@ -0,0 +1,668 @@
...
+            challenge_seed: dynamic.challenge_seed,
+            proof: dynamic.proofs[i],
+            faults: Vec::new(),
+        })
+    }
+
+    // If we receive a number of comm_rs which does not divide evenly by
+    // POST_SECTORS_COUNT, form a grouping of comm_rs which does divide
+    // evenly by POST_SECTORS_COUNT by duplicating the last comm_r in the
+    // group until LEN(comm_rs) == POST_SECTORS_COUNT.
```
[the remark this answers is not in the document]
laser: I couldn't come up with a good way to achieve this - but you and @schomatis are both right (it should be done). I humbly submit that we move forward with what's in this PR and DRY this thing up after I have knocked out other critical path PoSt things.

filecoin-proofs/src/api/post_adapter.rs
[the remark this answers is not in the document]
porcuquine: Got it, so this is an artifact of the decision to repeat the final `comm_r`?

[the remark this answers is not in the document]
laser: Yes sir, that is correct.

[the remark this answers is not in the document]
porcuquine: Thanks.

PR #521, opened 2019-03-03
Use mmap for replication in API.

### LINKED ISSUE #482: use mmap when sealing from API
### Description

We currently read sectors into memory from disk, then pass the data to `replicate`. We should instead mmap the file and pass the result.

We know, from having initially omitted this optimization for the replication game, that doing this will decrease total memory requirements for replication.

### Acceptance criteria

Similar to #481, best would still be profiling evidence of lowered memory consumption — or indirect evidence in the form of increased replicable sector size. However, given our experience with this optimization, just ensuring it is implemented in the same way as elsewhere and still works is probably fine for a first pass.

### Risks + pitfalls

See discussion below. There are some complexities around padding and the risk of interrupted sealing.

Make sure we properly flush to disk before 'moving' (file system move) the sealed data to its final destination.

### Where to begin

Look at the `drgporep-vanilla` example to see how data is generated: https://github.com/filecoin-project/rust-proofs/blob/master/filecoin-proofs/examples/drgporep-vanilla.rs#L71-L73

Compare with `drgporep-vanilla-disk`: https://github.com/filecoin-project/rust-proofs/blob/master/filecoin-proofs/examples/drgporep-vanilla-disk.rs#L36-L47

(We won't want to use a temp file, though. Ideally, we can 'seal in place' using the original data file itself.)

Here's where we read the data when sealing from the API: https://github.com/filecoin-project/rust-proofs/blob/master/filecoin-proofs/src/api/internal.rs#L366-L367

We still probably need to perform the padding: https://github.com/filecoin-project/rust-proofs/blob/master/filecoin-proofs/src/api/internal.rs#L369-L372

If we seal in place, we won't need to write: https://github.com/filecoin-project/rust-proofs/blob/master/filecoin-proofs/src/api/internal.rs#L396

But we will need to move the data to the location contained in `outpath`. @laser is there any problem with sealing in place then moving the file?

If the original file needs to remain intact, we can perhaps copy to output first, then seal in place. In that case, what should happen if sealing is interrupted and a partial file is left?

For that matter, what if we are interrupted when sealing in place over the original data. Long-term, a strategy for persisting state so seal could be restarted would be good. However, short-term, this possibility suggests we should leave the original data intact.

The most conservative solution would be to copy the original file, seal the copy in place, then move the sealed copy to the output path. I think this is probably the right answer and leave the discussion above so you understand the issues and logic behind the suggestion.

porcuquine:
Closes #482.

This implements #482, more or less as described there.

I decided to:
  - copy the input file to the output path
  - mmap the copy
  - replicate in place
  - either
    - delete the output file (in the case of a replication error)
    - flush to ensure the data is written to the output file

### DIFF
--- a/filecoin-proofs/src/api/internal.rs
+++ b/filecoin-proofs/src/api/internal.rs
@@ -436,6 +436,29 @@ pub struct SealOutput {
     pub snark_proof: SnarkProof,
 }
 
+/// Minimal support for cleaning (deleting) a file unless it was successfully populated.
+struct FileCleanup<T: AsRef<Path>> {
+    path: T,
+    success: bool,
+}
+
+impl<'a, T: AsRef<Path>> FileCleanup<T> {
+    fn new(path: T) -> FileCleanup<T> {
+        FileCleanup {
+            path,
+            success: false,
+        }
+    }
+}
+
+impl<T: AsRef<Path>> Drop for FileCleanup<T> {
+    fn drop(&mut self) {
+        if !self.success {
+            let _ = remove_file(&self.path);
+        }
+    }
+}
+
 pub fn seal<T: Into<PathBuf> + AsRef<Path>>(
     sector_config: &SectorConfig,
     in_path: T,
@@ -444,6 +467,8 @@ pub fn seal<T: Into<PathBuf> + AsRef<Path>>(
     sector_id_in: &FrSafe,
 ) -> error::Result<SealOutput> {
     let sector_bytes = usize::from(sector_config.sector_bytes());
+
+    let mut cleanup = FileCleanup::new(&out_path);
 
     // Copy unsealed data to output location, where it will be sealed in place.
     copy(&in_path, &out_path)?;
@@ -473,15 +498,11 @@ pub fn seal<T: Into<PathBuf> + AsRef<Path>>(
         &replica_id,
         &mut data,
         None,
-    )
-    .map_err(|e| {
-        // If there was an error during replication, remove the incomplete replica.
-        let _ = remove_file(&out_path);
-        e
-    })?;
+    )?;
 
-    // If we succeeded in replicating, flush the data.
+    // If we succeeded in replicating, flush the data and protect output from being cleaned up.
     data.flush()?;
+    cleanup.success = true;
 
     let public_tau = tau.simplify();
 
--- a/filecoin-proofs/src/api/internal.rs
+++ b/filecoin-proofs/src/api/internal.rs
@@ -1,11 +1,12 @@
 use std::collections::hash_map::Entry;
 use std::collections::HashMap;
-use std::fs::File;
-use std::io::{BufWriter, Read, Write};
+use std::fs::{copy, remove_file, File, OpenOptions};
+use std::io::{BufWriter, Read};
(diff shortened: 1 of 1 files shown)


filecoin-proofs/src/api/internal.rs
```diff
@@ -474,9 +473,15 @@ pub fn seal<T: Into<PathBuf> + AsRef<Path>>(
         &replica_id,
         &mut data,
         None,
-    )?;
+    )
+    .map_err(|e| {
```
laser: The file with path `out_file` will leak under the following circumstances:

- `OpenOptions::new().read(true).write(true).open(&out_path)` is `Err`
- `f_data.set_len(sector_bytes as u64)` is `Err`
- `MmapOptions::new().map_mut(&f_data)` is `Err`

I suggest that you handle each of those cases as you do here - by passing a cleanup function to `map_err`.

filecoin-proofs/src/api/internal.rs
```diff
@@ -443,16 +444,15 @@ pub fn seal<T: Into<PathBuf> + AsRef<Path>>(
...
+    let f_data = OpenOptions::new().read(true).write(true).open(&out_path)?;
 
-    // Zero-pad the data to the requested size.
-    for _ in data.len()..sector_bytes {
-        data.push(0);
-    }
+    // Zero-pad the data to the requested size by extending the underlying file if needed.
+    f_data.set_len(sector_bytes as u64)?;
+
+    let mut data = unsafe { MmapOptions::new().map_mut(&f_data).unwrap() };
```
schomatis: This might be a good opportunity to start fleshing out a basic `mmap` API that could be leveraged in https://github.com/filecoin-project/rust-fil-proofs/issues/481 (instead of everyone calling `MmapOptions::new()` on its own, which is actually what I'm doing atm).

porcuquine: @schomatis Can you clarify what you are doing? Are you calling `MmapOtions::new()` or creating a basic `mmap` API?

filecoin-proofs/src/api/internal.rs
porcuquine: @laser Good catch. I pushed a change to handle this right. We could refine that approach to be slicker in the future, but I think what I did (add a wrapper around the path, implementing `Drop`) at least provides the semantics we want. Let me know if that works for you.

filecoin-proofs/src/api/internal.rs
```diff
@@ -476,7 +500,9 @@ pub fn seal<T: Into<PathBuf> + AsRef<Path>>(
         None,
     )?;
 
-    write_data(out_path, &data)?;
+    // If we succeeded in replicating, flush the data and protect output from being cleaned up.
+    data.flush()?;
+    cleanup.success = true;
```
laser: Interesting idea. I don't use the `Drop` trait enough.

porcuquine: Yeah, it's the Rusty way to implement RAII.

porcuquine: (The mopre fleshed-out version of this work would move initialization/allocation into the struct too.)

Issue #455, opened 2019-01-24
Cache (expansion?) parents


porcuquine:
# Cache strategy design

~We need to have a very clear idea of how is `parents` used to be able to design the cache in an efficient manner, especially since there is a memory constraint and we just can't cache everything all the time.~
We need to be able to support a full-cache strategy, which will use a lot of memory — but have this be optional. A framework for these space vs time optimization configurations should be designed as a separate issue. (TODO: create that issue.)

What follows is my current understanding based on what I've seen from the code in ZigZag example,

```
./target/release/examples/zigzag --size 1024 --no-bench
``````

The same ZigZag graph seems to be used throughout the code, not the same `ZigZagGraph` structure but rather its "reversed" (`zigzag()`'ed ) version passed from one layer to another, specifically, the same `base_graph` and `expansion_degree` is maintained, which results in always the same parents for each separate direction (`forward`/`reversed`).

The `parents` can actually be split in the parents of the base graph and of the ZigZag extension, for now this concentrate on the ZigZag extension (`expanded_parents`, although for simplicity `parents` will be used in this text to actually mean the subgroup of parents we care to cache), this is because the underlying computation is more expensive than the base. We won't cache or analyze the underlying Feistel computations (anything below `correspondent`), which is basically a series of sequential hash functions, caching a particular hash result seems of little value since it is useful only for the computation of the current node (the chances of being reused are very low), and for roughly the same amount of space we can just cache the actual result (a parent index).

The parents of this "single" (in the sense described above) ZigZag are requested in 3 places:

1. During the replication phase, in `transform_and_replicate_layers`.

2. During the prove generation in `DrgPoRep::prove`.

3. During the verification in `DrgPoRep::verify`.

I'm not sure but I'm assuming that steps 1-2 would be called separately from step 3 and wouldn't share the cache (that is, that another process would be the one verifying it under normal -not test- circumstances).

There are two different profiles for the `parents` use, the replication (1) and the prove/verify (2/3), but from the initial write-up I'm assuming that we want to optimize step 1 and don't care about 2-3 (which use `parents` orders of magnitude less than step 1).

## Step 1: replication.

[...]

PR #522, opened 2019-03-05
Larger sectors

porcuquine:
### Problem

While trying to benchmark large(r) sectors, I (eventually) discovered that `setup` was hanging. The root cause turns out to be the known issue that our Feistel implementation permutes elements of type `u32`. Since we need `expansion_degree` (currently 8) elements per node, that means we can handle only up to 2^27 nodes. This corresponds to sectors of size 2^32 (2^27 * 32), which is 4GiB .

### Solution
The simple solution is to extend to use `u64`. In a perfect world, we might be able to select alternate generic implementations for `u32` or `u64` depending on the actual sector size, but that will/would be somewhat tricky and may not be worth it. Hopefully any benefits of smaller cache keys can be implemented just at the level of the cache while leaving the Feistel implementation to use `u64`. For now, if we must accept one or the other, the priority must be on supporting larger sectors. The current work at least moves the concrete type to a constant, and handles size-related logic with `size_of`. In order to support larger elements, I swapped `blake2s` (32-byte hashes) for `blake2b` (64-byte hashes).

### Breakdown
  - add regression test for non-terminating `setup()`.
  - add support for variable-sized feistel index.
  - set default `feistel::Index` to `u64`.
  - swap in `blake2b` for `blake2s` in support of `u64` indexes.

### NOTE
 - This does change the replication algorithm, so — for example — users of this code may become out of sync with the replication game server.
 - This also, unfortunately, seems to slow replication down a small amount (x 1.08). However, with caching (#455) that speed should be recovered. Being able to replicate large sectors at all is most important.

### Status

A replication test of 8 GiB is underway and proceeding as expected. Will update on completion or otherwise, but even if we hit another roadblock, the current work seems to have patched the first.

### DIFF
--- a/storage-proofs/src/crypto/feistel.rs
+++ b/storage-proofs/src/crypto/feistel.rs
@@ -1,19 +1,21 @@
-use blake2::{Blake2s, Digest};
+use blake2b_simd::blake2b;
+use std::mem;
 
 pub const FEISTEL_ROUNDS: usize = 3;
 // 3 rounds is an acceptable value for a pseudo-random permutation,
 // see https://github.com/filecoin-project/rust-proofs/issues/425
 // (and also https://en.wikipedia.org/wiki/Feistel_cipher#Theoretical_work).
 
-pub type FeistelPrecomputed = (u32, u32, u32);
+pub type Index = u64;
+
+pub type FeistelPrecomputed = (Index, Index, Index);
 
 // Find the minimum number of even bits to represent `num_elements`
 // within a `u32` maximum. Returns the left and right masks evenly
 // distributed that together add up to that minimum number of bits.
-pub fn precompute(num_elements: u32) -> FeistelPrecomputed {
-    let mut next_pow4 = 4;
+pub fn precompute(num_elements: Index) -> FeistelPrecomputed {
+    let mut next_pow4: Index = 4;
     let mut log4 = 1;
-
     while next_pow4 < num_elements {
         next_pow4 *= 4;
         log4 += 1;
@@ -30,11 +32,11 @@ pub fn precompute(num_elements: u32) -> FeistelPrecomputed {
 // one within the `[0, num_elements)` range using a `key` that will allow
 // the reverse operation to take place.
 pub fn permute(
-    num_elements: u32,
-    index: u32,
-    keys: &[u32],
+    num_elements: Index,
+    index: Index,
+    keys: &[Index],
     precomputed: FeistelPrecomputed,
-) -> u32 {
+) -> Index {
     let mut u = encode(index, keys, precomputed);
 
     while u >= num_elements {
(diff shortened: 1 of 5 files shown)


storage-proofs/src/crypto/feistel.rs
```diff
@@ -99,27 +102,40 @@ fn decode(index: u32, keys: &[u32], precomputed: FeistelPrecomputed) -> u32 {
...
+    }
 
-    let hash = Blake2s::digest(&data);
+    let hash = FeistelHash::digest(&data);
 
-    let r = u32::from(hash[0]) << 24
-        | u32::from(hash[1]) << 16
-        | u32::from(hash[2]) << 8
-        | u32::from(hash[3]);
+    let r = (0..HALF_FEISTEL_BYTES).fold(0, |acc, i| acc | Index::from(hash[i * 8]));
```
dignifiedquire: this seems to be doing sth different than before, or at least looks much slower

porcuquine: Yes, looks like I mangled something. Thanks — will push a fix.

porcuquine: @dignifiedquire Done.

storage-proofs/src/crypto/feistel.rs
```diff
@@ -99,29 +102,49 @@ fn decode(index: u32, keys: &[u32], precomputed: FeistelPrecomputed) -> u32 {
...
-    let mut data: [u8; 8] = [0; 8];
-    data[0] = (right >> 24) as u8;
-    data[1] = (right >> 16) as u8;
-    data[2] = (right >> 8) as u8;
-    data[3] = right as u8;
-
-    data[4] = (key >> 24) as u8;
-    data[5] = (key >> 16) as u8;
-    data[6] = (key >> 8) as u8;
-    data[7] = key as u8;
```
dignifiedquire: I compared the assembly with an unrolled version, and it looks much better: https://godbolt.org/z/2rY9kd I would suggest using unrolled versions for all of these, and use two different unrolled methods if you want to make this compatible

porcuquine: Fair enough, will do that.

porcuquine: Someone once told me (cough) that iterators in Rust were so performant I should prefer them as cost-free abstractions. (Just giving you a hard time.)

dignifiedquire: they are in many cases, which is why looked at the assembly :)

porcuquine: You're the best — thanks for checking and showing your work. Godbolt is cool.

storage-proofs/src/crypto/feistel.rs
```diff
@@ -99,29 +101,58 @@ fn decode(index: u32, keys: &[u32], precomputed: FeistelPrecomputed) -> u32 {
...
+        data[6] = (key >> 8) as u8;
+        data[7] = key as u8;
+    }
+
+    let hash = blake2b(&data);
+
+    {
+        let mut r = 0;
+        let mut shift = HALF_FEISTEL_BYTES * 8;
+        for item in hash.as_ref().iter().take(HALF_FEISTEL_BYTES) {
```
dignifiedquire: what about inlining this? :P

porcuquine: Fine… done.

PR #523, opened 2019-03-05
More zigzag example tweaks

porcuquine:
This PR contains small changes I needed while benchmarking replication and proving on a high-memory machine.

Specifically:
  - Optionally allow suppression of temp file, giving more control (should be more configurable later) over where the data resides.
 - Dump vanilla proofs so they can be reused for groth proving without replicating again (expensive!). We can't yet use these dumped proofs, but if I make it through a very long replication I don't want to lose the resulting artifact. Ability to restart groth proving without replication can be added when needed.
  - Don't generate random data, just zero the mmapped data by extending the file. This will save a lot of time on expensive machines.

### DIFF
--- a/filecoin-proofs/examples/zigzag.rs
+++ b/filecoin-proofs/examples/zigzag.rs
@@ -1,4 +1,5 @@
 extern crate bellman;
+extern crate chrono;
 extern crate pairing;
 extern crate rand;
 extern crate sapling_crypto;
@@ -7,6 +8,7 @@ extern crate clap;
 #[cfg(any(feature = "cpu-profile", feature = "heap-profile"))]
 extern crate gperftools;
 extern crate memmap;
+extern crate serde_json;
 extern crate tempfile;
 #[macro_use]
 extern crate slog;
@@ -14,6 +16,7 @@ extern crate slog;
 extern crate filecoin_proofs;
 extern crate storage_proofs;
 
+use chrono::Utc;
 use clap::{App, Arg};
 #[cfg(feature = "heap-profile")]
 use gperftools::heap_profiler::HEAP_PROFILER;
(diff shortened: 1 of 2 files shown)


filecoin-proofs/examples/zigzag.rs
```diff
@@ -108,6 +137,21 @@ pub fn file_backed_mmap_from(data: &[u8]) -> MmapMut {
...
+        .create(true)
+        .open(format!("./proofs-{:?}", Utc::now()))
+        .unwrap();
+
+    info!(
+        FCP_LOG,
+        "dumping proofs ({} bytes)",
+        serialized_proofs.len(); "target" => "status"
+    );
+    let _ = file.write_all(serialized_proofs);
```
dignifiedquire: don’t hide errors

filecoin-proofs/examples/zigzag.rs
```diff
@@ -214,21 +253,13 @@ fn do_the_work<H: 'static>(
...
-    // -- print statistics
-
-    //    let serialized_proofs = proofs.iter().fold(Vec::new(), |mut acc, p| {
-    //        acc.extend(p.serialize());
-    //        acc
-    //    });
-    //    let avg_proof_size = serialized_proofs.len() / samples as usize;
-    //
-    //info!(target: "stats", "Average proof size {}", prettyb(avg_proof_size));
+    let serialized_proofs = (serde_json::to_string(&all_partition_proofs))
```
dignifiedquire: there is to_bytes method on serde json, also you can create a streaming serializer with serde json, removing the need to buffer all into memory but rather write them out immediately

filecoin-proofs/examples/zigzag.rs
```diff
@@ -108,6 +138,26 @@ pub fn file_backed_mmap_from(data: &[u8]) -> MmapMut {
...
+        .open(format!("./proofs-{:?}", Utc::now()))
+        .unwrap();
+    info!(
+        FCP_LOG,
+        "dumping {} proofs",
+        all_partition_proofs.len(); "target" => "status"
+    );
+
+    if let Err(e) = serde_json::to_writer(file, all_partition_proofs) {
+        info!(
```
dignifiedquire: maybe `warn!`?

porcuquine: Okay, merging after this.

PR #526, opened 2019-03-07
Add --bench-only option to bench circuit without replication or groth proving

porcuquine:
Exactly as the descriptions says. This will help with sector size calculations.

### DIFF
--- a/filecoin-proofs/examples/zigzag.rs
+++ b/filecoin-proofs/examples/zigzag.rs
@@ -35,8 +35,8 @@ use bellman::Circuit;
 use sapling_crypto::jubjub::JubjubBls12;
 
 use storage_proofs::circuit::test::*;
-use storage_proofs::circuit::zigzag::{ZigZagCircuit, ZigZagCompound};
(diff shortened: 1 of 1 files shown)


filecoin-proofs/examples/zigzag.rs
```diff
@@ -501,10 +514,37 @@ fn main() {
     let taper_layers = value_t!(matches, "taper-layers", usize).unwrap_or(layers);
     let use_tmp = !matches.is_present("no-tmp");
     let dump_proofs = matches.is_present("dump");
-    let groth = matches.is_present("groth");
-    let bench = !matches.is_present("no-bench");
+    let no_bench = matches.is_present("no-bench");
+    let bench_only = matches.is_present("bench-only");
+    let bench = if no_bench {
+        if bench_only {
+            panic!("Cannot specify both --bench-only and --no-bench.");
```
dignifiedquire: there is a much nicer way of specifying this: https://docs.rs/clap/2.32.0/clap/struct.Arg.html#method.conflicts_with

porcuquine: I figured as much, thanks for the reference.

PR #528, opened 2019-03-07
[WIP] deps: update `merkle_light::merkle` API

schomatis:
```
Update use of `MerkleTree` with new `Store` trait. Set for now all `Store`s to
`VecStore`s which was the implicit `Store` when the trait didn't exist to
preserve the current memory behavior.
```

---------------------------------------------

The PR itself is done, the WIP part is waiting for https://github.com/filecoin-project/merkle_light/pull/13 to get merged and return the `merkle_light` dependency `branch` to `master` (from the contingent `mmap-experiments`).

-----------------------------------------------

Local memory profiling reports same memory consumption compared to master:

```
Total: 309.4 MB
   309.0  99.9%  99.9%    309.0  99.9% ::allocate_in (inline) /rustc/14997d56a550f4aa99fe737593cd2758227afc56/src/liballoc/raw_vec.rs:109
     0.3   0.1% 100.0%      0.3   0.1% ::reserve_internal (inline) /rustc/14997d56a550f4aa99fe737593cd2758227afc56/src/liballoc/raw_vec.rs:683
     0.1   0.0% 100.0%      0.1   0.0% sapling_crypto::jubjub::JubjubBls12::new (inline) ??:0
     0.0   0.0% 100.0%      0.0   0.0% core::ptr::::is_null (inline) /rustc/14997d56a550f4aa99fe737593cd2758227afc56/src/libcore/ptr.rs:1602
     0.0   0.0% 100.0%      0.0   0.0% _dl_mcount ??:0
     0.0   0.0% 100.0%      0.0   0.0% __cxa_thread_atexit_impl ??:0
     0.0   0.0% 100.0%      0.0   0.0% ::allocate_in (inline) /rustc/14997d56a550f4aa99fe737593cd2758227afc56//src/liballoc/raw_vec.rs:109
     0.0   0.0% 100.0%      0.0   0.0% ::build /usr/local/cargo/registry/src/github.com-1ecc6299db9ec823/slog-async-2.3.0/lib.rs:592
     0.0   0.0% 100.0%      5.4   1.7% ::build::{{closure}} (inline) /usr/local/cargo/git/checkouts/merkle_light-16b81cab1b05c5b8/ff18b53/merkle/src/merkle.rs:278
```
(both runs show the same, compare with https://github.com/filecoin-project/rust-fil-proofs/pull/518).

PR #529, opened 2019-03-07
use `MmapStore` for `MerkleTree` generation

schomatis:
~Blocked on https://github.com/filecoin-project/merkle_light/pull/13, after that is merged the dependency here should point to `merkle_light/master`.~

-----------------------------------------------------------

This is now dropping the allocated memory an order of magnitude:

```
# master:

Total: 309.4 MB
   309.0  99.9%  99.9%    309.0  99.9% ::allocate_in (inline) /rustc/14997d56a550f4aa99fe737593cd2758227afc56/src/liballoc/raw_vec.rs:109
     0.3   0.1% 100.0%      0.3   0.1% ::reserve_internal (inline) /rustc/14997d56a550f4aa99fe737593cd2758227afc56/src/liballoc/raw_vec.rs:683
     0.1   0.0% 100.0%      0.1   0.0% sapling_crypto::jubjub::JubjubBls12::new (inline) ??:0

# This PR with mmap'ed MT:

Total: 19.4 MB
    13.0  67.1%  67.1%     13.0  67.1% ::allocate_in (inline) /rustc/14997d56a550f4aa99fe737593cd2758227afc56/src/liballoc/raw_vec.rs:109
     4.3  22.0%  89.1%      4.3  22.0% ::reserve_internal (inline) /rustc/14997d56a550f4aa99fe737593cd2758227afc56/src/liballoc/raw_vec.rs:683
     2.0  10.3%  99.4%      2.0  10.3% rayon::iter::collect::special_extend@1594e0 (inline) :0

```

### DIFF
--- a/storage-proofs/Cargo.toml
+++ b/storage-proofs/Cargo.toml
@@ -12,7 +12,7 @@ bitvec = "0.5"
 sapling-crypto = { git = "https://github.com/zcash-hackworks/sapling-crypto", branch = "master" }
 rand = "0.4"
 libc = "0.2"
-merkle_light = { git = "https://github.com/filecoin-project/merkle_light", rev = "63a609decf3cc552b0aa79477b828f4a493f725d" }
+merkle_light = { git = "https://github.com/filecoin-project/merkle_light", branch = "master" }
 failure = "0.1"
 bellman = "0.1"
 byteorder = "1"
--- a/storage-proofs/src/vde.rs
+++ b/storage-proofs/src/vde.rs
@@ -91,40 +91,18 @@ where
 pub fn decode_domain_block<H>(
     sloth_iter: usize,
     replica_id: &H::Domain,
-    data: &[H::Domain],
+    byte_data: &[u8],
     node: usize,
+    node_data: <H as Hasher>::Domain,
     parents: &[usize],
+    _m: usize,
 ) -> Result<H::Domain>
 where
     H: Hasher,
 {
-    let key = create_domain_key::<H>(replica_id, node, parents, data)?;
+    let key = create_key::<H>(replica_id, node, parents, &byte_data, _m)?;
 
-    Ok(H::sloth_decode(&key, &data[node], sloth_iter))
-}
-
-/// Creates the encoding key, using domain encoded data.
-fn create_domain_key<H: Hasher>(
-    id: &H::Domain,
-    node: usize,
-    parents: &[usize],
-    data: &[H::Domain],
-) -> Result<H::Domain> {
-    let mut hasher = Blake2s::new().hash_length(32).to_state();
-    hasher.update(id.as_ref());
-
-    for parent in parents.iter() {
-        // special super shitty case
-        // TODO: unsuck
-        if node == parents[0] {
-            // skip, as we would only write 0s
-        } else {
-            hasher.update(data[*parent].as_ref());
-        }
-    }
-
-    let hash = hasher.finalize();
-    Ok(bytes_into_fr_repr_safe(hash.as_ref()).into())
+    Ok(H::sloth_decode(&key, &node_data, sloth_iter))
 }
 
 /// Creates the encoding key.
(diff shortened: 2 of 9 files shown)


schomatis:
Labeling as blocked until https://github.com/filecoin-project/rust-fil-proofs/pull/528 lands (and we now the new version of `MerkleTree` is working as expected).

storage-proofs/src/batchpost.rs
```diff
@@ -9,7 +9,7 @@ use serde::ser::Serialize;
 use crate::crypto::blake2s::blake2s;
 use crate::error::Result;
 use crate::hasher::{Domain, Hasher};
-use crate::merkle::MerkleTree;
+use crate::merkle::{MerkleTree, MmapStore};
```
dignifiedquire: you could just make the `MerkleTree` be an alias to `MerkleTree<D, F, MmapStore<D>>` that way you don't have to change all the instances in the code base

schomatis: don't bother reviewing this at the moment, fyi, the issue is not concentrating all renames in a single place but whether we want to have *all* `MerkleTree`s using `mmap` instead of `Vec` (if the answer is yes, then we can do as you suggest, but then I would ask why did we implement a new `Store` trait in the first place).

dignifiedquire: I see, that makes sense 👍

schomatis:
Currently failing due to a bug I introduced in https://github.com/filecoin-project/merkle_light/pull/13/commits/ff18b53744ff08025c13979f04b1e535a861aa97 where I'm calling `new_from_slice` with the wrong length (of bytes instead of `Element`s), although fixing that still fails now with `invalid replica node` so I'll have to keep digging.

schomatis:
MT upstream branch fixed, triggering another build.

schomatis:
Currently working but has now considerable differences with the current `master`, need to do a careful rebase.

storage-proofs/src/vde.rs
```diff
@@ -91,24 +91,25 @@ where
...
+    byte_data: &[u8],
     node: usize,
+    node_data: <H as Hasher>::Domain,
     parents: &[usize],
 ) -> Result<H::Domain>
 where
     H: Hasher,
 {
-    let key = create_domain_key::<H>(replica_id, node, parents, data)?;
+    let key = create_domain_key::<H>(replica_id, node, parents, &byte_data)?;
```
dignifiedquire: this change makes `create_domain_key` the same as `create_key` I believe, so you can just use that

storage-proofs/Cargo.toml
```diff
@@ -12,7 +12,7 @@ bitvec = "0.5"
 sapling-crypto = { git = "https://github.com/zcash-hackworks/sapling-crypto", branch = "master" }
 rand = "0.4"
 libc = "0.2"
-merkle_light = { git = "https://github.com/filecoin-project/merkle_light", rev = "63a609decf3cc552b0aa79477b828f4a493f725d" }
+merkle_light = { git = "https://github.com/filecoin-project/merkle_light", branch = "mmap-experiments" }
```
dignifiedquire: BLOCKING: lets merge this before merging this.

schomatis: Not sure which "this" is *this*.

storage-proofs/src/vde.rs
schomatis: Ok, I'll take a look, I wasn't sure why there were two functions now.

schomatis: Yes, you're right, I removed `create_domain_key` which is not used anymore.

storage-proofs/Cargo.toml
porcuquine: I suspect @dignifiedquire means merge the `mmap-experiments` branch in `merkle_light` and depend on the resulting `master` commit here.

dignifiedquire: indeed

schomatis: ok, as stated in the PR [description](https://github.com/filecoin-project/rust-fil-proofs/pull/529#issue-259195810) and the ZenHub pipeline that was already the plan, so glad we're on the same page

PR #532, opened 2019-03-08
feat(vde): improve performance of create_key

dignifiedquire:
- Avoid allocations by updating the hasher directly, instead of allocating a vector.
- Use [`blake2s_simd`](https://github.com/oconnor663/blake2s_simd) to speed up the actual hashing operations.
- Avoid some conversions from `Fr` to `FrRepr` and back.

Performance Improvements

- Testcase: `32 MB - zigzag default`

| variant                  | replication time |
|--------------------------|------------------|
| master |    178.938245178s, |
| this branch    |  169.802207285s |

Note: this currently uses an unpublished version of https://github.com/oconnor663/blake2s_simd

### DIFF
--- a/storage-proofs/src/fr32.rs
+++ b/storage-proofs/src/fr32.rs
@@ -1,7 +1,8 @@
 use crate::error::*;
 
-use byteorder::{LittleEndian, WriteBytesExt};
+use byteorder::{ByteOrder, LittleEndian, WriteBytesExt};
 
+use pairing::bls12_381::FrRepr;
 use pairing::{Engine, PrimeField, PrimeFieldRepr};
 
 // Contains 32 bytes whose little-endian value represents an Fr.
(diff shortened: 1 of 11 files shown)


storage-proofs/src/fr32.rs
```diff
@@ -34,6 +35,27 @@ pub fn bytes_into_fr<E: Engine>(bytes: &[u8]) -> Result<E::Fr> {
...
+        LittleEndian::read_u64(&r[16..24]),
+        u64::from(r[31] & 0b0011_1111) << 56
+            | u64::from(r[30]) << 48
+            | u64::from(r[29]) << 40
+            | u64::from(r[28]) << 32
+            | u64::from(r[27]) << 24
+            | u64::from(r[26]) << 16
+            | u64::from(r[25]) << 8
+            | u64::from(r[24]),
+    ];
```
schomatis: Not following this format, why do we force a little endian ordering (`LittleEndian::read_u64`) here? Is this the way the hasher will always output data?

dignifiedquire: this is just doing what we were doing before, we need a canoncial way to go from [u8; 32] to [u64; 4] (the internal representation of an Fr element. And there are roughly to ways, one is little endian one is big, and we settled on little endian a while ago

porcuquine:
@dignifiedquire Given the text on the `blake2s_simd` code, I'm not convinced we should make this the default code. Reproducing here for coherence of discussion:

> EXPERIMENTAL implementation of BLAKE2s and BLAKE2sp. Much of the code is copy-pasted from blake2b_simd, which is a real project, but this is currently only for benchmarking, and it's not published on crates.io.

Can we either put this behind a feature flag for benchmarking, or else make it configurable at runtime by the very-soon-to-arrive config intended exactly for things like this? We need to be able to experiment with and find the most performant implementation, while also preserving stability and high confidence of a default configuration.

dignifiedquire:
it passes the full test vector set from the official impl, and is from the same author as the other blake2b_simd impl, so I think this is fine to use for now. I am also in talks with the author and the two crates will become one, soonish.

porcuquine:
Okay, you convinced me — as long as we make sure it gets moved forward. Still need to review content.

PR #533, opened 2019-03-08
Adding a macro to free memory in the FFI automatically

### LINKED ISSUE #303: [CHORE] Find a way to reduce boilerplate in responses.rs files
## Motivation

This type of code is repeated for every response-type which we return through the FFI:

```
///////////////////////////////////////////////////////////////////////////////
/// MaxUnsealedBytesPerSectorResponse
/////////////////////////////////////

#[repr(C)]
pub struct MaxUnsealedBytesPerSectorResponse {
    pub status_code: SBResponseStatus,
    pub error_msg: *const libc::c_char,
    pub num_bytes: u64,
}

impl Default for MaxUnsealedBytesPerSectorResponse {
    fn default() -> MaxUnsealedBytesPerSectorResponse {
        MaxUnsealedBytesPerSectorResponse {
            status_code: SBResponseStatus::SBNoError,
            error_msg: ptr::null(),
            num_bytes: 0,
        }
    }
}

impl Drop for MaxUnsealedBytesPerSectorResponse {
    fn drop(&mut self) {
        unsafe {
            drop(c_str_to_rust_str(self.error_msg));
        };
    }
}

#[no_mangle]
pub unsafe extern "C" fn destroy_max_unsealed_bytes_per_sector_response(
    ptr: *mut MaxUnsealedBytesPerSectorResponse,
) {
    let _ = Box::from_raw(ptr);
}
```

We should find a way (macros?) to reduce this duplication.

vmx:
Dropping memory is a large part of the reposonses FFI code. With
this new macro `#[derive(DropStructMacro)]` the `Drop` implementation
will be auto-generated based on the struct.

Example:

```rust
pub struct FFIStagedSectorMetadata {
    pub sector_access: *const libc::c_char,
    pub sector_id: u64,
    pub pieces_len: libc::size_t,
    pub pieces_ptr: *const FFIPieceMetadata,

    // must be one of: Pending, Failed, Sealing
    pub seal_status_code: FFISealStatus,

    // if sealing failed - here's the error
    pub seal_error_msg: *const libc::c_char,
}
```

Will automatically create:

```rust
impl Drop for FFIStagedSectorMetadata {
    fn drop(&mut self) {
        unsafe {
            free_c_str(self.sector_access as *mut libc::c_char);
            drop(Vec::from_raw_parts(
                self.pieces_ptr as *mut FFIPieceMetadata,
                self.pieces_len,
                self.pieces_len,
            ));
            free_c_str(self.seal_error_msg as *mut libc::c_char);
        };
    }
}
```

To view the generated output after the macro was applied, you can use
[cargo-expand](https://github.com/dtolnay/cargo-expand):

```console
$ cd filecoin-proofs
$ cargo expand --lib api::responses
…
```

The diff between the original code and the generated code after running
`cargo expand --lib api::responses` and re-arranging it a bit (just
moving the `Drop` to the same place) the diff is:

```diff
164c164
<                     self.faults_ptr as *mut u8,
---
>                     self.faults_ptr as *mut u64,
417c417
<             }
---
>             };
476c476
<             }
---
>             };
500c500
<             }
---
>             };
530c530
<             }
---
>             };
566c566
<             }
---
>             };
```

Closes #303.

### DIFF
--- a/drop-struct-macro-derive/src/lib.rs
+++ b/drop-struct-macro-derive/src/lib.rs
@@ -1,0 +1,107 @@
+extern crate proc_macro;
+use crate::proc_macro::TokenStream;
+use proc_macro2::{Ident, Span};
+use quote::{quote, ToTokens};
+use syn;
+
+/// A struct that contains the name of a struct field and the corresponding type
+#[derive(Debug)]
+struct FieldNameType {
+    field_name: String,
+    field_type: proc_macro2::TokenStream,
+}
+
+/// The actual code to free the *const pointers
+impl quote::ToTokens for FieldNameType {
+    fn to_tokens(&self, tokens: &mut proc_macro2::TokenStream) {
+        let field_type = &self.field_type;
+        // Convert field type to string for easy pattern matching
+        let field_type_string = self
+            .field_type
+            .clone()
+            .into_iter()
+            .map(|token| token.to_string())
+            .collect::<String>();
+        match &field_type_string[..] {
+            // Free string with `free_c_str`
+            "libc::c_char" => {
+                let field_name = Ident::new(&self.field_name, Span::call_site());
+                let gen = quote! {
+                    free_c_str(self.#field_name as *mut #field_type);
+                };
+                gen.to_tokens(tokens);
+            }
+            // Expect all others to be vectors
+            _ => {
+                if !self.field_name.ends_with("_ptr") {
+                    panic!(
+                        "Pointer needs to have field with `_ptr` as suffix. \
+                         This field is named `{}`.",
+                        self.field_name
+                    )
+                }
+                // Field ends with `_ptr` so we can re-construct the corresponding length field
+                let field_name_len = Ident::new(
+                    &format!(
+                        "{}{}",
+                        &self.field_name[..&self.field_name.len() - 4],
+                        "_len"
+                    ),
+                    Span::call_site(),
+                );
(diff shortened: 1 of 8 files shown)


laser:
@dignifiedquire - Will you review this PR? My macro foo isn't particularly strong.

dignifiedquire:
I will, very soon :)

drop-struct-macro-derive/src/lib.rs
```diff
@@ -0,0 +1,107 @@
...
+            // Free string with `free_c_str`
+            "libc::c_char" => {
+                let field_name = Ident::new(&self.field_name, Span::call_site());
+                let gen = quote! {
+                    free_c_str(self.#field_name as *mut #field_type);
+                };
+                gen.to_tokens(tokens);
+            }
+            // Expect all others to be vectors
+            _ => {
```
dignifiedquire: Can we explicitly capture `Vec`, and throw a nice error if it is not one?

vmx: The comment might be misleading. There is no `Vec` here. There's fields ending with `_len` and `_ptr`. If that's not the case there's a panic (see next line), else we expect it to be a vector which is composed by those fields.

dignifiedquire: Ah okay, that makes sense.

dignifiedquire:
rebased, will merge once CI passes

PR #538, opened 2019-03-11
perf(drgporep): decode_domain_block: avoid data allocations

### LINKED ISSUE #537: vanilla proving is way too slow
### Description

*NOTE: this issue is marked **P0** because it blocks validation of large-sector sealing.*

Vanilla proving takes far longer than it should. Here is example output from benchmarks I ran last week:

```console
➜  rust-proofs git:(zigzag-example-taper) ✗ ./target/release/examples/zigzag --m 5 --expansion 8 --layers 10 --challenges 5 --size 262144 --groth
…
Feb 22 22:49:03.378 INFO vanilla_proving_time: 80.99321579 seconds, target: stats, place: filecoin-proofs/examples/zigzag.rs:208 zigzag, root: filecoin-proofs
```

Why does it take 80 seconds to generate 750 inclusion proofs?
  - 750 = 50 challenges * 15 inclusion proofs. (15 proofs = data node + replica node + 13 parent nodes)
  - This is (more or less) 5 x 15 = 75 proofs from each of 10 layers
  - In this example (256 MiB), proofs are 24 hashes in length.
  - Each hash is 32 bytes.
  - So total proof size is 750 x 24 x 32 = 600,000 bytes.
  - The trees are already generated, so no hashing should be happening.
  - The encodings are stored with the trees, so no replication should be happening.
  - Why is it taking 80 seconds to process 600,000 bytes?

It gets worse.

When I try to generate proofs after replicating 64 GiB, the vanilla proving never terminates. Using 8 partitions, I see 8 processes pinged 100% utilization.

In this case:
 - I was generating 8,848 x 15 = 132,720 proofs.
 -  For 64 GiB, the proofs are 32 hashes (32 bytes each).
 - So total data is 132,720 x 32 x 32 = 135,905,280
 - This is ~226 times larger than the previous (slow) example.
 - So at the previous rate, this should have taken about 5 hours.
 - It did not complete in 5 hours.

So… something is wrong. Let's find the problem and fix it.

### Acceptance criteria

Vanilla proofs for 1GiB take dramatically less than 80 seconds to generate. Maybe a couple seconds?

### Risks + pitfalls

This could be annoying to debug if it requires repeated performance of slow replication.
  - Use blake2s instead of pedersen. But first test both and see whether the choice affects proving time. If it does, that's unexpected and probably means hashing is happening (which it shouldn't).
  - Try to use small sectors. However, we do know the problem seems to get worse as sizes grow, so that may be difficult.
  - It would be easier if the products of replication could be saved to disk so we could restart at the proving phase.
    - This should be part of the disk-backed merkle-tree generation story (#481).
    -  If it's not, consider adding — both in support of this but *also especially* for ongoing benchmarks/experiments on large data. (cc: @schomatis)
    -  Having the ability to restart mid-replication, or to restart an interrupted replication would be extremely valuable.
    - This will also probably be a necessary eventual production feature.
    - Finally, it might be the case that making this possible fixes the problem. In case the problem is somehow related to memory bloat incurred during replication.

### Where to begin

- Run the command in the example above and record initial vanilla proving times.
- Read the code.
- Profile.
- Acquire but do not necessarily sacrifice a chicken.

dignifiedquire:
Should fix #537 

Speed improvements on my machine for `1MB` data zigzag (`cargo run -p filecoin-proofs --release --example zigzag -- --m 5 --expansion 8 --layers 10 --challenges 5 --size 1024 --no-bench`)

- `master`:  vanilla_proving_time: `495.295015ms`
- `perf-vanilla-proving`: vanilla_proving_time: `4.295946ms`

### DIFF
--- a/storage-proofs/src/vde.rs
+++ b/storage-proofs/src/vde.rs
@@ -98,13 +98,13 @@ pub fn decode_domain_block<H>(
 where
     H: Hasher,
 {
-    let key = create_key_domain::<H>(replica_id, node, parents, data)?;
+    let key = create_domain_key::<H>(replica_id, node, parents, data)?;
 
     Ok(H::sloth_decode(&key, &data[node], sloth_iter))
 }
 
 /// Creates the encoding key, using domain encoded data.
-fn create_key_domain<H: Hasher>(
+fn create_domain_key<H: Hasher>(
     id: &H::Domain,
     node: usize,
     parents: &[usize],
--- a/storage-proofs/src/vde.rs
+++ b/storage-proofs/src/vde.rs
@@ -85,31 +85,46 @@ where
     let key = create_key::<H>(replica_id, v, &parents, &data, graph.degree())?;
     let node_data = H::Domain::try_from_bytes(&data_at_node(data, v)?)?;
 
-    // TODO: round constant
     Ok(H::sloth_decode(&key, &node_data, sloth_iter))
 }
 
-pub fn decode_domain_block<'a, H>(
-    degree: usize,
+pub fn decode_domain_block<H>(
     sloth_iter: usize,
-    replica_id: &'a H::Domain,
-    data: &'a [H::Domain],
-    v: usize,
-    parents: Vec<usize>,
+    replica_id: &H::Domain,
+    data: &[H::Domain],
+    node: usize,
+    parents: &[usize],
 ) -> Result<H::Domain>
 where
     H: Hasher,
 {
-    let byte_data = data
-        .iter()
-        .flat_map(H::Domain::into_bytes)
-        .collect::<Vec<u8>>();
+    let key = create_key_domain::<H>(replica_id, node, parents, data)?;
 
-    let key = create_key::<H>(replica_id, v, &parents, &byte_data, degree)?;
-    let node_data = data[v];
+    Ok(H::sloth_decode(&key, &data[node], sloth_iter))
+}
 
-    // TODO: round constant
-    Ok(H::sloth_decode(&key, &node_data, sloth_iter))
+/// Creates the encoding key, using domain encoded data.
+fn create_key_domain<H: Hasher>(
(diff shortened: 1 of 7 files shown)


nicola:
what is the speedup gain?

dignifiedquire:
@nicola only vanilla_proving_time gets faster as described above

storage-proofs/src/vde.rs
```diff
@@ -85,31 +85,46 @@ where
...
+) -> Result<H::Domain> {
+    let mut hasher = Blake2s::new().hash_length(32).to_state();
+    hasher.update(id.as_ref());
+
+    for parent in parents.iter() {
+        // special super shitty case
+        // TODO: unsuck
+        if node == parents[0] {
+            // skip, as we would only write 0s
+        } else {
```
porcuquine: Can't we just leave this here? This edge case doesn't matter for performance at scale, and it's a wart we'll have to keep in sync as things change. Unless 'm missing something, I think we should remove.

storage-proofs/src/vde.rs
```diff
@@ -85,31 +85,46 @@ where
...
 
-    let key = create_key::<H>(replica_id, v, &parents, &byte_data, degree)?;
-    let node_data = data[v];
+    Ok(H::sloth_decode(&key, &data[node], sloth_iter))
+}
 
-    // TODO: round constant
-    Ok(H::sloth_decode(&key, &node_data, sloth_iter))
+/// Creates the encoding key, using domain encoded data.
+fn create_key_domain<H: Hasher>(
```
porcuquine: Name is hard to parse/interpret. `create_domain_key` would be clearer.

storage-proofs/src/vde.rs
dignifiedquire: I think I tried to remove it once and things started to break, lets do this another day

Issue #531, opened 2019-03-08
Sealing with Blake2s (was: Hybrid Merkle Trees)


porcuquine:
### Description
From #77:

---
**UPDATE: note 'pending calculations' below. It may actually be the case that the optimal balance is to use ALL blake2s — in which case the cost in development time and residual complexity of the hybrid solution would be wasted.**

**Based on calculations so far, I tentatively think the next step should be to get blake2s working in circuits (~generically~). This is a more manageable next step that would have to be accomplished anyway.**

**Since this might be the right answer, we could perhaps do this not by being generic but by just swapping blake2s in for pedersen hashing. As I recall, this would sidestep the complexity and be much simpler. That would let us perform tests to validate assumptions as quickly as posible.**

---
**UPDATE 2: calculations still need to be checked more carefully, but tentative conclusions are clear.**

![image](https://user-images.githubusercontent.com/1746729/54076988-26b56e80-4267-11e9-97c4-fe70f39190a5.png)

**Fuller analysis is forthcoming.**

Newer graphs suggest Blake2s advantage kicks in around 1TB, in which case we really need to run some pure blake tests (including circuits).

![image](https://user-images.githubusercontent.com/1746729/54100276-f232d700-437a-11e9-85cd-2bea33c63f6c.png)

Obviously one issue is that for blake2s to be the better solution, we need (with current numbers) an enormous sector (> 2 TiB — with which our current rules means ~4TiB~ UPDATE: although more favorable to blake2s, new numbers indicate even larger sector sizes to justify seal times). This is too big, so it's not yet clear that blake2s is the answer — but it's now the best guess. LATER UPDATE: More calculation suggest that if we can get groth proving time down to projected time, we can make this work with smaller sectors — maybe 256GiB, which may be viable. It's currently much higher, but there's a lot of noise with parameter generation, loading, perhaps something with how we parallelize. Worst case, at that size, we can break into many more circuits and run on much smaller machines.

One thing, however, that these graphs do make clear (if accurate) is that the hybrid solution is not actually better in any circumstances. Before the intersection point at which blake2s and pedersen perform the same, pedersen is better; after, blake is.

### UPDATE conclusion:
**Therefore, we can (for now) conclude that the hybrid solution provides no advantages. We should instead focus on being able to swap blake2s in for pedersen in circuits so we can get some real performance and circuit data to improve our projections as we solve for a viable configuration.**

---
[...]

PR #543, opened 2019-03-12
feat(storage-proofs): zizag, drgporep and por abstract over the hasher in circuits

dignifiedquire:
Basic tests are passing. Needs some more checks

- make por and drgporep circuits abstract over the hasher
- implement blake2s hasher using blake2s_simd

Ref #531

### DIFF
--- a/filecoin-proofs/examples/drgporep.rs
+++ b/filecoin-proofs/examples/drgporep.rs
@@ -14,8 +14,8 @@ use sapling_crypto::jubjub::{JubjubBls12, JubjubEngine};
 
 use storage_proofs::circuit;
 use storage_proofs::circuit::variables::Root;
-
 use storage_proofs::example_helper::Example;
+use storage_proofs::hasher::PedersenHasher;
 use storage_proofs::test_helper::fake_drgpoprep_proof;
 
 struct DrgPoRepExample<'a, E: JubjubEngine> {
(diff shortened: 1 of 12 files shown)


dignifiedquire:
@porcuquine this adds the hasher to the parameter identifiers, so I am guessing we have to change parameters versions again?

porcuquine:
Hmmm, if it only changes the set of parameters generated but doesn't change the circuit/parameters for a given identifier, we don't need to bump the version. (The version is needed so we don't find old parameters by the name with which we request new ones.)

But we will need to generate and publish the parameter again. When doing that, remember to clear your cache first, so we don't keep stale parameters around in `parameters.json`.

(It's possible I'm seeing something wrong, so if this sounds mistaken, please flag.)

dignifiedquire:
This is ready to go from my side. Note this will need new params published again, as now the hasher is part of the cache identifier

porcuquine:
Can you include the param publishing in this PR? That would be the best way to do it and should be our policy. Include a comment here stating that you pinned and verified.

porcuquine:
We should see a diff to `parameters.json` in this branch as well (over-explaining because this is new process).

porcuquine:
Looking good, @dignifiedquire. Any chance of extending this to the merklepor example, so we can run that with `--hasher blake2s`?

PR #544, opened 2019-03-12
publish v10 Groth parameters + paramcache always generates PoSt parameters

### LINKED ISSUE #510: modify paramcache to generate PoSt parameters by default
### Description

Once go-filecoin starts using the real PoSt generation and verification operations, we will need to:

A) Ensure that VDF PoSt Groth parameters are generated by paramcache
B) Ensure that PoSt parameters have been published and put into parameters.json
C) Ensure that `paramcache` runs in a reasonable amount of time (for go-filecoin)

laser:
Fixes #510 

## Why does this PR exist?

- dig updated Groth param cache-keys, so we needed to regen parameters
- adding PoSt params to `parameters.json`
- enabling PoSt (soon) in go-filecoin will require PoSt Groth params

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -96,7 +96,7 @@ jobs:
       - run:
           name: Ensure cache is hydrated with PoRep and PoSt Groth parameters (for test)
           command: |
-            cargo run --release --color=always --package filecoin-proofs --bin paramcache -- --include-post --test-only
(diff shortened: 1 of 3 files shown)


porcuquine:
Something's wrong with ffi, but otherwise looks good if params were published/pinned.

PR #545, opened 2019-03-13
Don't use blank_circuit with TestConstraintSystem.

porcuquine:
Addresses #546, but in an incomplete way.

**UPDATE: As described below, this PR is supplanted by #547**

This fixes the `AssignmentMissing` error produced during synthesis of a blank circuit when running the zigzag example.

From chat:

> The issue is that if we use `Option` in `blank_circuit`, we can't use the circuit with `TestConstraintSystem`. The reason it works for circuit generation is that Bellman's `KeypairAssembly` implements a version of `alloc` which never calls the closure which returns the `AssignmentMissing` error. This is because values aren't required.

> I will push a fix which makes this error go away and makes `zigzag` usable again. However, this also removes the ability to bench circuits without first running replication (which I added in recently and is useful when replication is extremely time-consuming).

> The right answer is probably to implement a new `BenchConstraintSystem` which is like `TestConstraintSystem` combined with `KeypairAssembly`. It never needs to handle values or perform real caclulations, but it does need to keep track of inputs and constraints. I might take a quick cut at that, but I don't have too much time to spend on it now.

Unless one of us comes up with a replacement quickly, let's roll with this to unbreak things — with a plan to revert and make the `BenchConstraintSystem` later.

### DIFF
--- a/filecoin-proofs/examples/zigzag.rs
+++ b/filecoin-proofs/examples/zigzag.rs
@@ -31,7 +31,6 @@ use std::io::Write;
 use std::time::{Duration, Instant};
 use std::u32;
 
-use bellman::Circuit;
 use sapling_crypto::jubjub::JubjubBls12;
 
 use storage_proofs::circuit::test::*;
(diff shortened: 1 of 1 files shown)


porcuquine:
Addendum: Tbest part of this is that when I went to start working on this, I discovered the work was already done. I had forgotten that we worked on this exact problem with ebfull at ZCon0. He did this work there, and it's already checked in. I'll verify that it works…

porcuquine:
Whoops, not quite. BenchCS was for actual benchmarks so does need to calculate. (Our punning of 'bench' rears its head.)

porcuquine:
The new constraints system is now implemented in #547. This PR should be closed without merging once that one merges. I'm leaving it open now for context visibility and so we have it in case #547 doesn't work out. However, I think the right thing is to go with #547.

PR #557, opened 2019-03-14
Feat/gf1586 integrate real post

laser:
Fixes the rust-fil-proofs side of [this go-filecoin story](
https://github.com/filecoin-project/go-filecoin/issues/1586).

## What's in this PR?

1. Allow API consumers to call in to real PoSt generation and verification calls
1. Delete the old, fake PoSt generation function
1. Add some logging at the FFI layer using language that API consumers will grok

## Why's this PR needed?

1. API consumers are ready to generate and verify proofs!

### DIFF
--- a/filecoin-proofs/src/api/internal.rs
+++ b/filecoin-proofs/src/api/internal.rs
@@ -293,21 +293,6 @@ pub fn generate_post(
     generate_post_collect_output(n, fixed_output)
 }
 
-pub fn fake_generate_post(
-    dynamic: GeneratePoStDynamicSectorsCountInput,
-) -> error::Result<GeneratePoStDynamicSectorsCountOutput> {
-    let faults: Vec<u64> = if !dynamic.input_parts.is_empty() {
-        vec![0]
-    } else {
-        Default::default()
-    };
-
-    Ok(GeneratePoStDynamicSectorsCountOutput {
-        proofs: vec![[42; 192]],
-        faults,
-    })
-}
-
 pub fn verify_post(
     dynamic: VerifyPoStDynamicSectorsCountInput,
 ) -> error::Result<VerifyPoStDynamicSectorsCountOutput> {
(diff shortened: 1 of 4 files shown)


filecoin-proofs/src/api/internal.rs
```diff
@@ -416,7 +402,13 @@ fn verify_post_fixed_sectors_count(
     let commitments = fixed
         .comm_rs
         .iter()
-        .map(|comm_r| PedersenDomain(bytes_into_fr::<Bls12>(comm_r).unwrap().into_repr()))
+        .map(|comm_r| {
+            PedersenDomain(
+                bytes_into_fr::<Bls12>(comm_r)
+                    .expects("could not could not map comm_r to Fr")
```
dignifiedquire: @laser This is not good to panic here, as this means an invalid `comm_r` triggers a panic. Instead this should simply fail the verification.

laser: > This is not good to panic here

@porcuquine [does not agree with you](https://filecoinproject.slack.com/archives/CEGB67XJ8/p1552509403191100?thread_ts=1552509239.190000):

> This is a 100% should-panic.

dignifiedquire: @porcuquine we should talk^^

porcuquine: Nah, this can/should return an error. When this came up before I spoke too soon and made a wrong assumption about where this was coming from. My mistake — sorry for the noise.

porcuquine: Without the context, I mistakenly assumed this was caused by a bug in the challenge-seed truncating code. A`BadFrBytes` error there would have signified a programmer error (mine), and the right answer would have been to correct the code.

porcuquine: I do think this should use the error-reporting mechanism to inform the API caller that malformed input was received — since that is almost certainly the result of a bug somewhere else. If we can't return an error and can only fail verification, obviously we need to fail rather than report successful verification on bad input.

Issue #560, opened 2019-03-15
zigzag example shows log messages clobbering


sidke:
### Description

when running the zigzag example, it shows clobbered log output

`./target/release/examples/zigzag --challenges 1 --size 1 --sloth 0 --m 5 --hasher pedersen --groth`

```
...
Mar 15Mar  15:04:22.30815 15:04:22.308  INFO INFO returning tree, layer: 5replication_time: 56.167734ms, target:,  place: storage-proofs/src/layered_drgporep.rs:433 storage_proofs::layered_drgporep, root:stats , storage-proofsplace:                                                          
filecoin-proofs/examples/zigzag.rs:243 zigzag, root: filecoin-proofs                                                                                                                                                                                                                              
Mar 15 15:04:22.308 INFO setting tau/aux, layer: Mar 15 15:04:22.3080,  place: storage-proofs/src/layered_drgporep.rs:474 storage_proofs::layered_drgporep, root: storage-proofs                                                                                                                  
INFO replication_time/byte: 54.851µs, target: stats, place: filecoin-proofs/examples/zigzag.rs:257 zigzagMar 15 15:04:22.308 , INFO setting tau/auxroot, layer: 1, place: storage-proofs/src/layered_drgporep.rs:474 storage_proofs::layered_drgporep, :root: storage-proofs                      
 filecoin-proofs                                                                                                                                                                                                                                                                                  
Mar 15 15:04:22.308 Mar 15 15:04:22.INFO replication_time/GiB: 58895.812788224s, target: stats308 , INFO setting tau/auxplace: filecoin-proofs/examples/zigzag.rs:262 zigzag, root: filecoin-proofs                                                                                               
, layer: 2Mar, place: 15  15:04:22.309 INFO generating 1 partition proofs, place: filecoin-proofs/examples/zigzag.rs:268 zigzag, root: filecoin-proofs                                                                                                                                            
Mar 15 storage-proofs/src/layered_drgporep.rs:474 storage_proofs::layered_drgporep15:04, root: storage-proofs    
...
```

### Acceptance criteria

log messages do not clobber each other

PR #561, opened 2019-03-15
spawn additional loggers from single root logger

sidke:
fixes: https://github.com/filecoin-project/rust-fil-proofs/issues/560

## what's in this pr:

- uses a single root logger to make more loggers from. this allows all loggers to share the same drain instance instead of creating their own and potentially racing to the same output (stdout or file)


logging-toolkit/src/lib.rs
```diff
@@ -11,11 +13,12 @@ use slog::LevelFilter;
...
 use std::env;
 
-pub fn make_logger(
-    root_name: &'static str,
-    use_json_env_name: &str,
-    min_log_level_env_name: &str,
-) -> Logger {
+lazy_static! {
+    static ref ROOT_LOGGER: Logger =
+        make_root_logger("RUST_PROOFS_LOG_JSON", "RUST_PROOFS_MIN_LOG_LEVEL");
```
porcuquine: Should probably be `RUST_FIL_PROOFS…` now.

dignifiedquire: Maybe just `FIL_PROOFS`? it really doesn't matter that this is written in rust for env variables

PR #567, opened 2019-03-19
disk-backed merkle trees

schomatis:
Towards #481.

Creates a file-mapping to back the MT data in the `Store`.

Uses https://github.com/filecoin-project/merkle_light/pull/15.


schomatis:
Need to figure out some design issues (details embedded in the code, see only the last [commit](https://github.com/filecoin-project/rust-fil-proofs/pull/567/commits/cb775d698ef125310e898af40241d21c59fe166c), the rest comes from https://github.com/filecoin-project/rust-fil-proofs/pull/529 which needs to be merged first):

1. How do we set the directory where we store MT `mmap`-files?

2. How do we set the hash (or any other tag) to uniquely identify that file for future replication runs to load it?

3. How do we integrate this in the [`merkle_light`](https://github.com/filecoin-project/merkle_light) repository?

---------------

The output looks like

```
Mar 20 15:55:37.700 INFO encoding, layer {}: 0, root: storage-proofs, place: storage-proofs/src/layered_drgporep.rs:447 storage_proofs::layered_drgporep
Mar 20 15:55:37.700 INFO created mmap-file, path: layer-0-2857851453, root: storage-proofs, place: storage-proofs/src/merkle.rs:215 storage_proofs::merkle
Mar 20 15:55:37.701 INFO returning tree, layer: 0, root: storage-proofs, place: storage-proofs/src/layered_drgporep.rs:440 storage_proofs::layered_drgporep
Mar 20 15:55:37.703 INFO encoding, layer {}: 1, root: storage-proofs, place: storage-proofs/src/layered_drgporep.rs:447 storage_proofs::layered_drgporep
Mar 20 15:55:37.703 INFO created mmap-file, path: layer-1-2824864975, root: storage-proofs, place: storage-proofs/src/merkle.rs:215 storage_proofs::merkle
Mar 20 15:55:37.704 INFO returning tree, layer: 1, root: storage-proofs, place: storage-proofs/src/layered_drgporep.rs:440 storage_proofs::layered_drgporep
Mar 20 15:55:37.706 INFO encoding, layer {}: 2, root: storage-proofs, place: storage-proofs/src/layered_drgporep.rs:447 storage_proofs::layered_drgporep
Mar 20 15:55:37.707 INFO created mmap-file, path: layer-2-3995110150, root: storage-proofs, place: storage-proofs/src/merkle.rs:215 storage_proofs::merkle
```

dignifiedquire:
Hmm this is actually tricky, if we want to nicely integrate this into our system

porcuquine:
@schomatis

I think we will need to evolve toward a stateful sealing process in which intermediate state is stored on disk along with enough metadata to continue a paused sealing, restart an interrupted one, or replay a historical example (for benchmarking with new code).

Having thought about this a while, the simplest thing I can think of to begin with is to organize a replication/proving (sealing) job into its own directory. So all the intermediate merkle tree files would live in that directory. Because this would give us namespacing of distinct replications, we could use legible names for the layers. To avoid the confusion around replicas and original data changing in each layer, we could just number them and have something like 'tree-0', 'tree-1', … 'tree-10'.

> Need to figure out some design issues (details embedded in the code, see only the last [commit](https://github.com/filecoin-project/rust-fil-proofs/pull/567/commits/cb775d698ef125310e898af40241d21c59fe166c), the rest comes from #529 which needs to be merged first):
> 
> 1. How do we set the directory where we store MT `mmap`-files?

Based on the plan above, this should be supplied as an input on a per-replication basis. (This may not address your question, but hopefully adds a detail to the discussion.)

> 2. How do we set the hash (or any other tag) to uniquely identify that file for future replication runs to load it?

As above, I think we can use a simple naming scheme and namespace with directories.

To both this and question 1, I have answered how we should *choose* names, not what mechanism should be used to 'set' them.

I don't think there should be any global configuration of directories of logic for generating filenames at this level of abstraction. I tentatively think the whole path should be an argument to `DiskMmapStore::New`, and responsibility for deriving that path should be the responsibility of the consumer.

> 3. How do we integrate this in the [`merkle_light`](https://github.com/filecoin-project/merkle_light) repository?
> 
> The output looks like
> 
> ```
> Mar 20 15:55:37.700 INFO encoding, layer {}: 0, root: storage-proofs, place: storage-proofs/src/layered_drgporep.rs:447 storage_proofs::layered_drgporep
> Mar 20 15:55:37.700 INFO created mmap-file, path: layer-0-2857851453, root: storage-proofs, place: storage-proofs/src/merkle.rs:215 storage_proofs::merkle
> Mar 20 15:55:37.701 INFO returning tree, layer: 0, root: storage-proofs, place: storage-proofs/src/layered_drgporep.rs:440 storage_proofs::layered_drgporep
> Mar 20 15:55:37.703 INFO encoding, layer {}: 1, root: storage-proofs, place: storage-proofs/src/layered_drgporep.rs:447 storage_proofs::layered_drgporep
> Mar 20 15:55:37.703 INFO created mmap-file, path: layer-1-2824864975, root: storage-proofs, place: storage-proofs/src/merkle.rs:215 storage_proofs::merkle
> Mar 20 15:55:37.704 INFO returning tree, layer: 1, root: storage-proofs, place: storage-proofs/src/layered_drgporep.rs:440 storage_proofs::layered_drgporep
> Mar 20 15:55:37.706 INFO encoding, layer {}: 2, root: storage-proofs, place: storage-proofs/src/layered_drgporep.rs:447 storage_proofs::layered_drgporep
> Mar 20 15:55:37.707 INFO created mmap-file, path: layer-2-3995110150, root: storage-proofs, place: storage-proofs/src/merkle.rs:215 storage_proofs::merkle
> ```

schomatis:
@porcuquine Not sure I follow how this advances on the current state of this PR:

> To avoid the confusion around replicas and original data changing in each layer, we could just number them and have something like `"tree-0"`, `"tree-1"`, … `"tree-10"`.

This is already happening, see sample output.

> Because this would give us namespacing of distinct replications, we could use legible names for the layers. 

How exactly does this namespace system would work? What does it give it its distinct quality? If I run two replication process with the same parameters but different data how will I distinguish them? When I run a replication later with the *same* data as one I did before, how will I identify the saved state to reload it and avoid recomputing?

dignifiedquire:
I think the idea is to create a folder `/tmp/fil-proofs/seal-<unique id>/` for every sealing, in which all tmp files would be stored. And if successfull deleting those files again. 

For this first round I think the id could just be entirely random, later on we can start thinking about tying it to the data.

schomatis:
Ok, but then why are we storing them in the first place if we're not going to load them back again? (I'm pretty sure at this point there's a key fact about the motivation of this `mmap` effort I'm missing.)

Would the idea be to then scan all the stored files looking for the original data and see if it matches  when doing a new replication?

schomatis:
To clarify a bit more, my question is not how to give them a unique tag to the replicas so they won't overlap,(we could quickly scan the directory and find the largest index and increase it), but how do we identify them to later access them (if we indeed want to access them later)?

dignifiedquire:
I think the idea is to do reusing in a second round, when we have a more complete design, but correct me if I am wrong @porcuquine

schomatis:
> I think the idea is to do reusing in a second round, when we have a more complete design, but correct me if I am wrong @porcuquine

Ok, I'm going to go with that premise for now then (at least to unblock some work) unless I'm giving a clear indication of the contrary, but I still don't really see a benefit here, we're already `mmap`ing MTs (or soon will be when the other PR gets merged), assigning a file to that mapping to not reuse it later will consume disk considerably (that is, even worse than without this PR, saving to disk for me was a way to trade disk for CPU time, but doing it like this wins nothing and running the same test over and over will create more and more replica files -even identical ones- which this PR won't take the responsibility to clean). Just a thought. Thanks for unblocking @dignifiedquire.

porcuquine:
@schomatis There are two main reasons we would like to have the intermediate trees on disk in a recoverable way:

1. So that we can restart interrupted sealing. There are any number of reasons why a long sealing process might be interrupted. Having to restart from the beginning would be expensive and painful. However, if we have not saved the merkle trees then we will indeed have to not only regenerate them but also re-encode the data for every layer (effectively sealing all over again) when it comes time to generate proofs.

2. So we can cache generated (or partially generated) merkle trees for PoSt, or for any other reason. One reason might be because the tree was generated on a different machine, in a model where encoding, merkle-tree generation, and proving are all accomplished by machines configured for the task drawing work from a queue.

Although neither of these is absolutely pressing yet, they are the reason this will be valuable. Since the first goal of #481 may be completed with #529, we could pause work here now and add durable persistence later.

To be clear: I don't think it should be impossible to instantiate an mmapped merkle tree without a named file — so it's not obvious to me that adding the ability to persist to disk needs to negatively impact any work flow. We don't necessarily have to start using this feature for replication proper until we are ready to.

That having been said, if we *do* proceed and integrate with replication then we will have the tools we need to make replication restartable, the longer replication takes, the more important this feature will be — both in development and in production.

If you'd like to defer, let me know.

porcuquine:
@schomatis 

> To clarify a bit more, my question is not how to give them a unique tag to the replicas so they won't overlap,(we could quickly scan the directory and find the largest index and increase it), but how do we identify them to later access them (if we indeed want to access them later)?

It would be just like any other file. The creator is responsible for deciding where to put it and remembering where that was (as well as for cleaning up when no longer needed).

porcuquine:
One more thought, stimulated by the apparent out-of-memory errors I'm seeing with mmap merkle trees on master. I believe explicitly assigning a file may avoid this issue — since this guarantees enough backing store for the data, as opposed to anonymous map which depends on the size of swap. (You are closer to this, though, and can better evaluate. However, since I know you are investigating replicating larger sectors, this may be worth consideration.)

schomatis:
@dignifiedquire Need to figure out how to integrate the new `DiskMmapStore` into the MT code, particularly how to set the file name of the file being created (e.g., `tree-0`, `tree-1`, etc.) which at the moment is being handled by an environmental variable (`INTERNAL_TREE_NAME`).

schomatis:
> The creator is responsible for deciding where to put it and remembering where that was (as well as for cleaning up when no longer needed).

Also need to figure out how this should work. The current tests don't do that at the moment (so this functionality is disabled by default unless `DISK_BACKED_MMAP_TREE_DIR` is set). I'll need to create a new test than can take ownership of the directory and make sure there are no repeated files (e.g., no two files named `tree-0`, the second one having different data and potentially breaking the test, when running replications).

porcuquine:
A quick note, with apologies for brevity: please use config values and set appropriate defaults rather than directly checking environment variables.

schomatis:
@porcuquine As mentioned in the code the environmental variables are *not* the final implementation. They allow me to present you a working PoC which we can iterate on to actually discuss implementation issues which I don't see a clear solution to (in contrast with the env.var that can, when finally decided on a design direction, be easily replaced with a more appropriate configuration), such as integrating the disk-backed MT API and defining exactly *how* is the consumer going to manage the `mmap` files directory.

schomatis:
Continuing the discussion of https://github.com/filecoin-project/rust-fil-proofs/pull/567#issuecomment-477419767, from Slack,

@dignifiedquire 
> @schomatis my idea was to have an api on mt which allows you to do sth like this:
```
let mmap = make_mmap("tree-0");
let tree = MerkleTree::from_data_with_store(data, mmap);
```
> now this probably needs sth additional like `MerkleTree::calcluate_size_for_data(data)` to get the required size
> when creating the mmap

@schomatis 
>  Seems reasonable, my main question is at which layer should that API be lowered?
>  Should every `Store` support some type of file addressing?
> Even if the `Vec` and plain `Mmap` would just ignore it and keep it on memory also (and then I should make them panic when trying to force them to load from file, something that only the disk-backed version should implement).

porcuquine:
What I would ideally like to see as soon as possible is a way to set config to use disk-backed merkle trees (temp files which are cleaned when dropped under normal operation fine) so I can observe whether this allows replication of larger sectors than otherwise. If possible, let's orient this phase of work toward meeting that first goal.

storage-proofs/src/merkle.rs
```diff
@@ -153,6 +165,168 @@ fn path_index<T: Domain>(path: &[(T, bool)]) -> usize {
...
+                .write(true)
+                .create(true)
+                .open(&tree_path)
+                .unwrap()
+        } else {
+            // We don't know which layer this is (`INTERNAL_TREE_NAME`) or we
+            // don't know where to save it (`DISK_BACKED_MMAP_TREE_DIR`), just
+            // create *any* temporary file.
+            tempfile().unwrap()
+        };
```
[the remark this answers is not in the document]
schomatis: Please see the comment above, https://github.com/filecoin-project/rust-fil-proofs/pull/567#issuecomment-478653789, if you could orient me towards a concrete API implementation that would be very helpful.

schomatis:
> What I would ideally like to see as soon as possible is a way to set config to use disk-backed merkle trees (temp files which are cleaned when dropped under normal operation fine) so I can observe whether this allows replication of larger sectors than otherwise. If possible, let's orient this phase of work toward meeting that first goal.

Sure, I'll take care of that tomorrow if you feel that is blocking progress here, do note that what you are requesting is already the default behavior of this PR, see https://github.com/filecoin-project/rust-fil-proofs/pull/567/files#diff-5e638bd0999f1c7aeb3b6e0888ec8eccR30.

schomatis:
And https://github.com/filecoin-project/rust-fil-proofs/pull/567/files#diff-5e638bd0999f1c7aeb3b6e0888ec8eccR172.

storage-proofs/src/merkle.rs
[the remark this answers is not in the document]
porcuquine: Let's talk this through tomorrow if what I say here doesn't make sense, but something like:

- For `DiskMmapStore`, I would do exactly as described above (i.e. extend `new` to take an optional path.
- For `MerkleTree`, I would do as @dignifiedquire suggests:
  - Add a method which returns the required size for a given `data` input.
- Add a new type `DiskMerkleTree` which uses a `DiskMmapStore` (analogous to `MerkleTree` using `MmapStore`).
- Add a bolean config var, say, `REPLICATE-WITH-DISK-TREES` (or some better name).
- Use either `MerkleTree` or `DiskMerkleTree` when replicating, depending on the value of that config option. (NOTE: don't check for env vars directly in the code, use `config::get_config`.

porcuquine:
> Sure, I'll take care of that tomorrow if you feel that is blocking progress here, do note that what you are requesting is already the default behavior of this PR, see https://github.com/filecoin-project/rust-fil-proofs/pull/567/files#diff-5e638bd0999f1c7aeb3b6e0888ec8eccR30.

Understood, and that should make this fairly easy. I'm just suggesting a way to bring this particular design phase to a close so we can experiment with the feature and move it into use.

storage-proofs/src/merkle.rs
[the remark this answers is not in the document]
schomatis: @porcuquine 

> * For `DiskMmapStore`, I would do exactly as described above (i.e. extend `new` to take an optional path.
> * For `MerkleTree`, I would do as @dignifiedquire suggests:
>   
>   * Add a method which returns the required size for a given `data` input.

As mentioned above I still don't see how this fits with the existing `Store` trait (which doesn't have a file concept and should be extended with one, which would panic with the current implementations laying the burden on the user to be careful not to use it with the normal `MerkleTree`; I'm not saying this is a bad option just that a comment addressing this would be useful to me). I'm definitely missing something of what both you and @dignifiedquire are trying to explain me.

schomatis:
> Understood, and that should make this fairly easy. I'm just suggesting a way to bring this particular design phase to a close so we can experiment with the feature and move it into use.

I understand and agree with that, I was just pointing out that you can already check out this PR (without setting *any* environmental variable) to test if using a file-mapping version of `mmap` has an impact on the maximum allowed replicated sector; also agree that env.vars are not for production.

storage-proofs/src/merkle.rs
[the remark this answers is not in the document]
porcuquine: It's entirely possible that I'm the one missing it, since I'm only observing this work from a slight remove.

That said, here's what I *think*:
 - `Store` is very general and doesn't even know about the distinction between a `Vec` and `Mmap` i,mplementation.
 - `MmapStore` doesn't (and doesn't need to) know about files (but also, therefore, doesn't accomplish the goals of our disk-backed merkle trees).

It's not obvious to me why `Store` would need to become aware of files. That some particular `Store` implementation is backed by concrete files is an implementation detail that *shouldn't* be leaked into the more general trait. If it were, that would make instantiating an ordinary vector-based merkle tree awkward (for example).

I feel like we might be saying the same thing in some ways. It's just not clear to me why you think we need to push the details of a more specific type implementing the `Store` trait up into that trait. Doing that would seem to cause the problems you note, so why would we do it?

It's possible that @dignifiedquire (who has tracked and participated in this work a little more closely) understand the issue better, though.

In case what I'm saying does in fact make sense, the focusing question is: what would go wrong if we proceeded roughly as I described above and didn't modify the `Store` trait at all?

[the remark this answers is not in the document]
schomatis: What you're saying sounds right, maybe it's my Rust inexperience that led to assume that the only way was to modify the `Store` trait, I'll prepare a PR on the `merkle_light` side with this (even if I can't make it work it will give me a more concrete code to move forward with this discussion). Thanks for looking into this.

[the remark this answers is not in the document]
schomatis: @porcuquine Quick follow-up question to your proposed solution, how do I switch between `DiskMerkleTree` and (the non-path version) `MerkleTree`? That is, using our dynamic configuration framework how can I select between two statically different types? (Again, I might be missing some Rust related aspect of the solution.)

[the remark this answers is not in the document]
porcuquine: The way to do it is to use a trait object, i.e. `Box<merkle::MerkleTree>`. That said, this *will* probably mean propagating a lot of changes.

Given that we probably don't actually need to eventually support both, it might make morse sense to just define another type `DefaultMerkleTree` (or whatever) which can be set as appropriate to switch at compile time.

We could add a feature flag to switch between the two if we're nervous about starting out with disk-backed as the default (where we probably want to end up). Something like `[cfg(disk-trees)]` and `[cfg(not(disk-trees))]`.

The main reason not to just go with disk trees to begin with — assuming we can verify no performance loss — is the cleanup issue.

The one thing to avoid is spending a lot of time trying one way, failing, then sinking more time into the other. For that reason, and because I also because it's probably closer to where we want to end up, I think I'd probably recommend just going with the compile-time flag instead of the run-time config.

storage-proofs/src/layered_drgporep.rs
```diff
@@ -428,6 +435,42 @@ pub trait Layers {
...
+
+                                let tree_name = format!("tree-{}-{}", layer, rand::random::<u32>());
+                                // FIXME: The user of `REPLICATE_DISK_TREES_DIR` should figure out
+                                // how to manage this directory, for now we create every file with
+                                // a different random number; the problem being that tests now do
+                                // replications many times in the same run so they may end up reusing
+                                // the same files with invalid (old) data and failing.
+
+                                graph
+                                    .merkle_tree_path(&data_copy, Some(&trees_dir.join(tree_name)))
```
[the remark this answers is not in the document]
schomatis: This is the only place to exercise the `Some(path)` part of the API (although I agree it's not nice but with the directory included in the configuration it should be contained and clearly marked).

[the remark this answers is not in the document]
porcuquine: I see what you mean. What I meant is that we *could* just provide the `Some(path)` API for future use and default to using temp files in replication for now. That would let us focus on the details of how we want to manage files later as distinct work.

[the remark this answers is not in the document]
schomatis: Yes, but that will leave a significant part of the code untested (like the bug you found in the MT repo), but I think we can reach a middle ground here with the new config framework.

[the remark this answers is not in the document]
schomatis: Added `get_config` for the directory where the MTs will be stored on disk, set to empty by default to create temporary files (but added a configuration in CircleCI to test that code path).

storage-proofs/src/drgraph.rs
```diff
@@ -60,6 +71,83 @@ pub trait Graph<H: Hasher>: ::std::fmt::Debug + Clone + PartialEq + Eq {
...
+            Ok(MerkleTree::from_par_iter(
+                (0..self.size()).into_par_iter().map(f),
+            ))
+        } else {
+            Ok(MerkleTree::new((0..self.size()).map(f)))
+        }
+    }
+
     /// Returns the merkle tree depth.
     fn merkle_tree_depth(&self) -> u64 {
```
dignifiedquire: this should have `#[cfg(not(feature = "disk-trees")))]` so that only one of the two is compiled

storage-proofs/src/layered_drgporep.rs
```diff
@@ -428,6 +445,54 @@ pub trait Layers {
                             // If we panic anywhere in this closure, thread.join() below will receive an error —
                             // so it is safe to unwrap.
                             let graph = transfer_rx.recv().unwrap();
+
```
dignifiedquire: it would be nicer to just call a function here e.g `let tree_d = graph.merkle_tree()` and have that be different depending on the compile flag

dignifiedquire:
The config impl changed during this PR, could you rebase it and resolve that conflicts, that makes review a bit easier

storage-proofs/src/drgraph.rs
```diff
@@ -60,6 +71,83 @@ pub trait Graph<H: Hasher>: ::std::fmt::Debug + Clone + PartialEq + Eq {
         }
     }
 
+    /// Builds a merkle tree based on the given data and stores it in `path`
+    /// (if set).
+    // FIXME: If the `path` is set to `None` should we still create a temporary
+    //  file somewhere? (at the moment we don't)
```
porcuquine: I think the answer to this question is 'yes', but I might be missing an implication.

schomatis: Yes, those `FIXME`s have been deprecated (we already implement that), I'll remove them.

storage-proofs/src/layered_drgporep.rs
schomatis: Yes, definitely much nicer.

schomatis:
@dignifiedquire Rebased.

schomatis:
Moving the logic that uses the new decoupling of the MT (https://github.com/filecoin-project/merkle_light/pull/17) to https://github.com/filecoin-project/rust-fil-proofs/pull/642 in order to constraint the focus on this PR and get it ready to merge.

storage-proofs/src/drgraph.rs
```diff
@@ -1,15 +1,24 @@
...
 use crate::error::*;
 use crate::hasher::pedersen::PedersenHasher;
 use crate::hasher::{Domain, Hasher};
+#[cfg(feature = "disk-trees")]
+use crate::merkle::DiskMmapStore;
 use crate::merkle::MerkleTree;
 use crate::parameter_cache::ParameterSetIdentifier;
 use crate::util::{data_at_node, NODE_SIZE};
+#[cfg(feature = "disk-trees")]
+use crate::SP_LOG;
```
dignifiedquire: Maybe group all the `cfg`s together, so it is easier to understand, what that pulls in

storage-proofs/src/vde.rs
```diff
@@ -126,3 +127,28 @@ pub fn create_key<H: Hasher>(
     let hash = hasher.finalize();
     Ok(bytes_into_fr_repr_safe(hash.as_ref()).into())
 }
+
+// FIXME: CHECK THIS.
```
dignifiedquire: Seems okay to me, what exactly do you want to check?

schomatis: The existence of the function itself mainly, I introduced it to avoid accesing slices directly and wanted to raise attention to it to validate if this was in alignment with the current API we want to build here.

I'll remove the note then.

storage-proofs/src/drgraph.rs
schomatis: I got the idea that `rustfmt` was pulling them apart (respecting import name order), I'll try to group them with braces.

schomatis: It seems Rust doesn't like grouping imports with `{...}`, were you thinking in a particular arrangement?

schomatis:
@dignifiedquire Addressed last reviews, could you take a final look and approve if looks good please? After that and once https://github.com/filecoin-project/merkle_light/pull/15 is merged I'll use its new release here and merge.

.circleci/config.yml
```diff
@@ -80,6 +80,27 @@ jobs:
...
+      - image: filecoin/rust:latest
+    working_directory: /mnt/crate
+    resource_class: xlarge
+    steps:
+      - checkout
+      - attach_workspace:
+          at: "."
+      - restore_cache:
+          keys:
+            - cargo-v7-{{ checksum "rust-toolchain" }}-{{ checksum "Cargo.toml" }}-{{ checksum "Cargo.lock" }}-{{ arch }}
```
dignifiedquire: BLOCKING: this should be `v9`

storage-proofs/src/drgraph.rs
dignifiedquire: I just meant like this:

```rust
#[cfg(feature = "disk-trees")]
use a;
#[cfg(feature = "disk-trees")]
use b;

use other;
use d;
```

storage-proofs/src/layered_drgporep.rs
```diff
@@ -433,6 +451,50 @@ pub trait Layers {
...
+                                    // Signal `merkle_tree_path` to create a temporary file.
+                                    // FIXME: duplicating `merkle_tree_path` to avoid the
+                                    //  "temporary value dropped while borrowed" (because we
+                                    //  were creating a temporary `PathBuf` below).
+                                    graph.merkle_tree_path(&data_copy, None).unwrap()
+                                } else {
+                                    // Try to create `tree_dir`, ignore the error if `AlreadyExists`.
+                                    if let Some(create_error) = fs::create_dir(&tree_dir).err() {
+                                        if create_error.kind() != io::ErrorKind::AlreadyExists {
+                                            panic!(create_error);
```
dignifiedquire: Do we have to panic?

storage-proofs/src/layered_drgporep.rs
```diff
@@ -433,6 +451,50 @@ pub trait Layers {
...
+                                    let tree_d = graph
+                                        .merkle_tree_path(
+                                            &data_copy,
+                                            Some(&PathBuf::from(tree_dir).join(format!(
+                                                "tree-{}-{}",
+                                                layer,
+                                                rand::random::<u32>()
+                                            ))),
+                                        )
+                                        .unwrap();
```
dignifiedquire: is it safe to `unwrap` here?

dignifiedquire: (I see we did this before too.. )

storage-proofs/src/vde.rs
```diff
@@ -126,3 +127,26 @@ pub fn create_key<H: Hasher>(
     let hash = hasher.finalize();
     Ok(bytes_into_fr_repr_safe(hash.as_ref()).into())
 }
+
+/// Creates the encoding key from a `MerkleTree`.
+/// The algorithm for that is `Blake2s(id | encodedParentNode1 | encodedParentNode1 | ...)`.
+/// It is only public so that it can be used for benchmarking
+pub fn create_key_from_tree<H: Hasher>(
```
dignifiedquire: It would be great to have a test to ensure the two `crate_key` impls don't get out of sync

storage-proofs/src/layered_drgporep.rs
schomatis: (to this and previous comment) the `scope` is set to return `()` (empty?) errors

https://github.com/filecoin-project/rust-fil-proofs/blob/1b3b2641e52b7742fe5ec8e404bfb326b751a2a0/storage-proofs/src/layered_drgporep.rs#L421

should I change that?

dignifiedquire: that is not empty errors, that is just empty success value, so you can still return errors regularly

storage-proofs/src/vde.rs
schomatis: I don't know how to do that, the previous "check this" `FIXME` note drew attention to this function because I basically copy-pasted changing slices to trees, I don't even like having both functions.

storage-proofs/src/layered_drgporep.rs
dignifiedquire: this is a type alias defined in `errors.rs` as `type Result<T> = std::result::Result<T, Error>`

schomatis: @dignifiedquire I'm trying to return (https://github.com/filecoin-project/rust-fil-proofs/commit/616dc80619476380237074143380bf2026539708) an error but getting

```
error[E0308]: mismatched types
   --> storage-proofs/src/layered_drgporep.rs:451:59
    |
451 |                           let thread = scope.spawn(move |_| {
    |  ___________________________________________________________^
452 | |                             // If we panic anywhere in this closure, thread.join() below will receive an error —
453 | |                             // so it is safe to unwrap.
454 | |                             let graph = transfer_rx.recv().unwrap();
...   |
502 | |                             return_channel.send((layer, tree_d)).unwrap();
503 | |                         });
    | |_________________________^ expected enum `error::Error`, found ()
    |
    = note: expected type `error::Error`
               found type `()`

error: aborting due to previous error
```

I've upset Rust expectations of what is being returned but not sure how.

schomatis: Returning an `Error` at the end of the closure fixes it but I can't seem to return an `Ok` and I thought that the empty `()` signaled there's no need to return anything.

schomatis: The conditional compilation might be the problem here, I think I need to be consistent across both branches.

schomatis: Using wrong syntax, it was `Err(Error::)`.

schomatis:
@dignifiedquire Take a look at this statement

https://github.com/filecoin-project/rust-fil-proofs/blob/1b3b2641e52b7742fe5ec8e404bfb326b751a2a0/storage-proofs/src/layered_drgporep.rs#L433-L434

if it's still valid I'll revert to the previous panic/unrwap version.

storage-proofs/src/drgporep.rs
```diff
@@ -327,7 +327,7 @@ where
                 let extracted = decode_domain_block::<H>(
                     pub_params.sloth_iter,
                     &pub_inputs.replica_id.expect("missing replica_id"),
-                    tree_r.as_ref(),
+                    tree_r,
```
dignifiedquire: @schomatis thinking about this some more, this method only needs the leaves, which we could still provide as slice, avoiding the need for the create_key_from_tree method below,

schomatis: @dignifiedquire With the new offload logic I can't expose any internals of the `mmap` stores (which I see as a plus) because I need to control when they are valid and when they're not, can't let you hold a slice to something that isn't in memory. At least I didn't see a way around it without breaking too much stuff, were you thinking about a specific implementation?

dignifiedquire: Slowly starting to get worried about the implications of not being able to read slices anymore. 

What we could do that would solve that problem I believe is to provide the following method 

```rust
fn read_into(&self, &mut [u8]) {}
```

which would read into a preallocated slice, instead of allocating a new vec everytime we need to read sth. This could be then used in the `create_key_from_tree` to read into the hasher directly instead of all the allocations 

--- 

Independent of the performance implications I guess we have to have two different methods for now :(

schomatis: Sounds good, I'll implement that on the `merkle_light` side.

storage-proofs/Cargo.toml
```diff
@@ -11,7 +11,7 @@ logging-toolkit = { path = "../logging-toolkit" }
 bitvec = "0.10"
 rand = "0.4"
 libc = "0.2"
-merkletree = "=0.4.1"
+merkletree = { git = "https://github.com/filecoin-project/merkle_light", branch = "feat/store/read_into"}
```
schomatis: Updating after https://github.com/filecoin-project/merkle_light/pull/20 is merged and we cut 0.5.1.

storage-proofs/src/vde.rs
```diff
@@ -141,9 +141,11 @@ pub fn create_key_from_tree<H: Hasher>(
     hasher.update(id.as_ref());
 
     // The hash is about the parents, hence skip if a node doesn't have any parents
+    let mut scratch: [u8; NODE_SIZE] = [0; NODE_SIZE];
```
schomatis: Is this the correct size?

dignifiedquire: I believe so, thought allocation should happen inside of the `if` statement

PR #569, opened 2019-03-20
feat: switch to the latest pairing

dignifiedquire:
- use fork filecoin-project/bellman
- use fork filecoin-project/phase2
- use fork filecoin-project/sapling-crypto
- use fork filecoin-project/pairing


dignifiedquire:
Unfortunately the perf improvements did not translate to replication.

dignifiedquire:
yes, because we can more easily tweak bellman and we will need to do that and it does make operations more efficient, juznit faster in the grand overall scheme

PR #573, opened 2019-03-22
refactor(storage_proofs): make fixed node size in drgraph more explicit

vmx:
The node size in the depth robust graph was already hard-coded to 32 bytes,
but this value was still used on a few occasions as a magic number. Change
the code to rely on those 32 bytes which is defined by the NODE_SIZE constant.

I'm not sure if I should do those kind of refactors without having all the context,
please let me know.

### DIFF
--- a/storage-proofs/src/drgraph.rs
+++ b/storage-proofs/src/drgraph.rs
@@ -9,7 +9,7 @@ use crate::hasher::pedersen::PedersenHasher;
 use crate::hasher::{Domain, Hasher};
 use crate::merkle::MerkleTree;
 use crate::parameter_cache::ParameterSetIdentifier;
-use crate::util::data_at_node;
+use crate::util::{data_at_node, NODE_SIZE};
 /// The default hasher currently in use.
 pub type DefaultTreeHasher = PedersenHasher;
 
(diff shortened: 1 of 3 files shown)


porcuquine:
This is probably a clean kill. The history is that node size was once a parameter (called `lambda`, which you will still see in some places). However, since we deeply depend on this being constant, passing it everywhere added noise which we removed by hard coding. We intentionally deferred rationalizing usage of a constant in order to bound the cost of the initial change.

Finishing the job is valuable, but if we are going to do so we should go all the way (or as far as possible). If you search the code base for `lambda` and `32`, you'll probably find everything. Just be careful because there might be times that the number 32 appears but is not an instance of this constant. That's a great reason to complete this refactor, but it also means a little more discretion will be required. I mention this partially to signal that whoever reviews this should apply a little extra attention to make sure we get this right.

Finally, I don't think `util` is the right place for this constant to live. The constant exists there for the historical reasons described above — as part of an intentionally bounded initial approach and/or legacy.

I think `drgraph` would be a better location — since both graph-node and merkle-tree generation interfaces are defined there.

dignifiedquire:
While I agree with @porcuquine that we should clean up everything, I think we should still merge this as is, and do the rest in another round rather than wait for that.

porcuquine:
Sure, go for it.

PR #578, opened 2019-03-28
break out packaging from publishing script

sidke:
this is step 1ii of https://github.com/filecoin-project/go-filecoin/issues/2441

easier access to publishing `tar`s will remove the need for go-filecoin to understand the file structure of its dependencies

### DIFF
--- a/scripts/publish-release.sh
+++ b/scripts/publish-release.sh
@@ -1,14 +1,15 @@
 #!/usr/bin/env bash
 
+set -e
+
 RELEASE_BRANCH="master"
 RELEASE_NAME="$CIRCLE_PROJECT_REPONAME-`uname`"
-RELEASE_PATH="$CIRCLE_ARTIFACTS/$RELEASE_NAME"
-RELEASE_FILE="$RELEASE_PATH.tar.gz"
+RELEASE_FILE="/tmp/$RELEASE_NAME.tar.gz"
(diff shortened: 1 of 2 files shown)


scripts/publish-release.sh
```diff
@@ -20,24 +21,9 @@ fi
...
-cp target/release/libfilecoin_proofs.a $RELEASE_PATH/lib/
-cp target/release/libfilecoin_proofs.pc $RELEASE_PATH/lib/pkgconfig
-
-pushd $RELEASE_PATH
-
-tar -czf $RELEASE_FILE ./*
+`dirname $0`/package-release.sh
 
-popd
+mv /tmp/release.tar.gz $RELEASE_FILE
```
porcuquine: This should probably be a var (`TAR_FILE`, as in `package-release.sh`). As far as I can tell, this depends on these being set to the same value in both scripts. Ideally, it would be impossible for them to differ, but if we need to hard-code the relationship, let's at least document.

laser: Interesting idea. What do you think, @sidke?

porcuquine: @laser Can you approve this if/when @sidke makes a decision, or otherwise close?

sidke: i will start work back on getting single env var precompiled installs after getting PIP in

PR #579, opened 2019-03-29
refactor(storage_proofs): move condition outside of the loop

vmx:
Also adding documentation that a node without parents is indicated by
the first parent being the node itself.

I'd like to note that something here is strange though. When I was looking into "unsuckifying" the code, i first thought i could change the if statement to `node != 0` as only the first node doesn't have any parents. But that  would make the tests fail (more specifically the `zigzag_drgporep::tests::extract_all_*`ones).

It turns out that there's also parents where every element is a `2`, so it's not the first node, but it's self-referencing itself nonetheless. From reading the code at https://github.com/filecoin-project/rust-fil-proofs/blob/808dc884642acbf7b78b282eb7d933c7cc2cc3a7/storage-proofs/src/drgraph.rs#L118 I thought only the first node has no parents.

### DIFF
--- a/storage-proofs/src/drgraph.rs
+++ b/storage-proofs/src/drgraph.rs
@@ -66,6 +66,10 @@ pub trait Graph<H: Hasher>: ::std::fmt::Debug + Clone + PartialEq + Eq {
     }
 
     /// Returns a sorted list of all parents of this node. The parents may be repeated.
+    ///
+    /// If a node doesn't have any parents, then this vector needs to return a vector where
+    /// the first element is the requested node. This will be used as indicator for nodes
+    /// without parents.
     fn parents(&self, node: usize) -> Vec<usize>;
 
     /// Returns the size of the graph (number of nodes).
--- a/storage-proofs/src/vde.rs
+++ b/storage-proofs/src/vde.rs
@@ -118,12 +118,9 @@ pub fn create_key<H: Hasher>(
     let mut hasher = Blake2s::new().hash_length(NODE_SIZE).to_state();
     hasher.update(id.as_ref());
 
-    for parent in parents.iter() {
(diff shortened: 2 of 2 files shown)


schomatis:
Thanks for looking into this, your reasoning is correct but the `parents` API is not clear (for me this constitutes a bug, or at least technical debt).

Some graphs are "reversed" meaning that the parents should be nodes with indexes *bigger* than its children (in contrast with the forward case you are referencing), this makes the last node the "root" node which self-references, so we can't just check against node `0` (the 2 you saw is just some weird artifact of the test that taints the vector to check if it was modified).

(Related to that, in `create_key`, I don't follow the check against the *first* node of the parents vector though, since we sort in increasing order and a reversed graph would find the self-referencing last nodes at the end.)

For more information take a look at the `ZigZag::parents()` version which handles the reverse case, that is, we have to kinds of parents, the base type you saw (always in "forward" mode) and the ZigZag version which allows "reversing".

https://github.com/filecoin-project/rust-fil-proofs/blob/808dc884642acbf7b78b282eb7d933c7cc2cc3a7/storage-proofs/src/zigzag_graph.rs#L170-L176

How this two types of parents interact in cases like the VDE function you fixed I'm not sure of, we should standardize or at least document a bit more the `parents()` function. This goes somewhat out of the original scope of this PR so you can leave this as is and I'll take over (or you can go for the glory and I'll guide you through it :)

schomatis:
(Labeling as bug as we don't have a "technical debt" tag, but happy to create one.)

vmx:
Sorry for late reply @porcuquine. Please take it on, this is out of scope for me. After reading the related discussions, I don't feel like I'm deep enough into the subject.

PR #580, opened 2019-04-01
add piece through FFI w/out buffering piece into memory

laser:
## What's in this PR?

- `write_padded` now polymorphic over `Read`, which allows stream processing of inbound bytes
- `write_and_preprocess` also polymorphic over `Read` for same reasons
- `write_and_preprocess` consumers (e.g. FFI) can pass a `File`, which satisfies `Read`
- FFI consumers provide path, which can point to a named pipe or normal file

## Why's this PR needed?

- Buffering client pieces completely into memory for padding is bad
- We'd like to stream to `write_padded` from the go-filecoin `DagReader` (we do so by way of a named pipe)

### DIFF
--- a/filecoin-proofs/examples/ffi/main.rs
+++ b/filecoin-proofs/examples/ffi/main.rs
@@ -18,12 +18,14 @@ use ffi_toolkit::rust_str_to_c_str;
 use rand::{thread_rng, Rng};
 use std::env;
 use std::error::Error;
+use std::io::Write;
 use std::ptr;
 use std::slice::from_raw_parts;
 use std::sync::atomic::AtomicPtr;
 use std::sync::mpsc;
 use std::thread;
 use std::time::Duration;
+use tempfile::NamedTempFile;
 use tempfile::TempDir;
 
 ///////////////////////////////////////////////////////////////////////////////
@@ -48,14 +50,20 @@ unsafe fn create_and_add_piece(
     let c_piece_key = rust_str_to_c_str(piece_key.clone());
     defer!(free_c_str(c_piece_key));
 
+    // write piece bytes to a temporary file
+    let mut file = NamedTempFile::new().unwrap();
+    let p = { String::from(file.path().to_str().unwrap().clone()) };
+    let _ = file.write_all(&piece_bytes);
+    let c_piece_read_pipe = rust_str_to_c_str(p);
+
     (
         piece_bytes.clone(),
         piece_key.clone(),
         add_piece(
             sector_builder,
             c_piece_key,
-            &piece_bytes[0],
-            piece_bytes.len(),
+            piece_bytes.len() as u64,
+            c_piece_read_pipe,
         ),
     )
 }
--- a/filecoin-proofs/examples/ffi/main.rs
+++ b/filecoin-proofs/examples/ffi/main.rs
@@ -55,6 +55,7 @@ unsafe fn create_and_add_piece(
     let p = { String::from(file.path().to_str().unwrap().clone()) };
     let _ = file.write_all(&piece_bytes);
     let c_piece_path = rust_str_to_c_str(p);
+    defer!(free_c_str(c_piece_path));
 
     (
         piece_bytes.clone(),
--- a/filecoin-proofs/examples/ffi/main.rs
+++ b/filecoin-proofs/examples/ffi/main.rs
@@ -54,7 +54,7 @@ unsafe fn create_and_add_piece(
     let mut file = NamedTempFile::new().unwrap();
     let p = { String::from(file.path().to_str().unwrap().clone()) };
     let _ = file.write_all(&piece_bytes);
-    let c_piece_read_pipe = rust_str_to_c_str(p);
+    let c_piece_path = rust_str_to_c_str(p);
 
     (
         piece_bytes.clone(),
@@ -63,7 +63,7 @@ unsafe fn create_and_add_piece(
             sector_builder,
             c_piece_key,
             piece_bytes.len() as u64,
-            c_piece_read_pipe,
+            c_piece_path,
         ),
     )
 }
--- a/filecoin-proofs/examples/ffi/main.rs
+++ b/filecoin-proofs/examples/ffi/main.rs
@@ -15,6 +15,7 @@ include!(concat!(env!("OUT_DIR"), "/libfilecoin_proofs.rs"));
 use ffi_toolkit::c_str_to_rust_str;
 use ffi_toolkit::free_c_str;
 use ffi_toolkit::rust_str_to_c_str;
+use filecoin_proofs::error::ExpectWithBacktrace;
 use rand::{thread_rng, Rng};
 use std::env;
 use std::error::Error;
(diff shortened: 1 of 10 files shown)


sector-base/src/io/fr32.rs
```diff
@@ -635,20 +635,25 @@ pub fn clear_right_bits(byte: &mut u8, offset: usize) {
...
+    let mut buffer = [0; CHUNK_SIZE];
     let mut written = 0;
 
-    for chunk in source.chunks(chunk_size) {
-        written += write_padded_aux(&FR32_PADDING_MAP, chunk, target)?;
+    while let Ok(bytes_read) = source.read(&mut buffer) {
+        if bytes_read == 0 {
+            break;
+        }
+        written += write_padded_aux(&FR32_PADDING_MAP, &buffer[..bytes_read], target)?;
```
schomatis: (another one just for me) the problem with the previous implementation was that we were allocating a new (temporary) `CHUNK_SIZE` buffer in every `chunks` call of the loop, right?

filecoin-proofs/examples/ffi/main.rs
```diff
@@ -48,14 +50,21 @@ unsafe fn create_and_add_piece(
     let c_piece_key = rust_str_to_c_str(piece_key.clone());
     defer!(free_c_str(c_piece_key));
 
+    // write piece bytes to a temporary file
+    let mut file = NamedTempFile::new().unwrap();
+    let p = { String::from(file.path().to_str().unwrap().clone()) };
```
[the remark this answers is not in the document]
dignifiedquire: - this should definitely not `unwrap`, at least please use `.expect("the reason I think this will never happen")`
- you can just do  as far as I can tell
```rust
let p = file.path().to_str().to_string()
```

sector-base/src/api/disk_backed_storage.rs
```diff
@@ -121,7 +121,7 @@ impl SectorManager for DiskManager {
     fn write_and_preprocess(
         &self,
         access: &str,
-        data: &[u8],
+        data: &mut dyn Read,
```
dignifiedquire: why `dyn Read` and not `impl Read`?

sector-base/src/io/fr32.rs
```diff
@@ -1092,10 +1097,10 @@ mod tests {
     // `write_padded` for 151 bytes of 1s, check padding.
     #[test]
     fn test_write_padded() {
-        let data = vec![255u8; 151];
+        let mut data: Vec<u8> = vec![255u8; 151];
         let buf = Vec::new();
         let mut cursor = Cursor::new(buf);
-        let written = write_padded(&data, &mut cursor).unwrap();
+        let written = write_padded(&mut data[..].as_ref(), &mut cursor).unwrap();
```
dignifiedquire: You shouldn't need `.as_ref()` if you already use `&mut`, otherwise sth is off.

storage-proofs/benches/preprocessing.rs
```diff
@@ -51,16 +52,15 @@ fn preprocessing_benchmark(c: &mut Criterion) {
     );
 }
 
-fn write_padded_bench(file: &mut File, data: &[u8]) {
-    write_padded(&data, file).unwrap();
-
+fn write_padded_bench(file: &mut File, mut data: Vec<u8>) {
```
dignifiedquire: Might be nicer to pass in `&mut [u8]` instead of giving it the whole vec, that way you can avoid the clones

sector-base/src/io/fr32.rs
```diff
@@ -635,20 +635,25 @@ pub fn clear_right_bits(byte: &mut u8, offset: usize) {
...
 }
 
-pub fn write_padded<W: ?Sized>(source: &[u8], target: &mut W) -> io::Result<usize>
+const N: usize = 1000;
+const CHUNK_SIZE: usize = 127 * N;
+
+pub fn write_padded<R, W>(source: &mut R, target: &mut W) -> io::Result<usize>
 where
-    W: Read + Write + Seek,
+    R: Read + ?Sized,
```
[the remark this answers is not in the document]
laser: The `?Sized` means that the `Sized` constraint is optional. By default, all type variables are automatically constrained to `Sized`.

For more information - and a more technically correct explanation than I'm able to give - check out this thread: 

https://users.rust-lang.org/t/trait-objects-and-the-sized-trait/14410/2

sector-base/src/api/disk_backed_storage.rs
laser: > why dyn Read and not impl Read?

The reason that I used a trait object (`dyn Read`) and not `impl Read` is because I don't know what `impl Read` does or how I'd use it here.

storage-proofs/benches/preprocessing.rs
laser: I need the clone because the closure in which `write_padded_bench` is called is `FnMut` and not `FnOnce`. I'd have to use an `Rc` (or some other reference counted type) to avoid the clone. I'm not particularly concerned about memory efficiency in the benchmark setup.

dignifiedquire: well the issue is that it dilutes the benchmark, because now you are benchmarking how long it takes to clones those bytes

sector-base/src/api/disk_backed_storage.rs
dignifiedquire: Some details are here: https://internals.rust-lang.org/t/what-is-the-difference-between-impl-trait-and-dyn-trait/8823/12

Tldr `dyn Read` is only needed when you want to make runtime choices, which I believe is not needed here. 

`impl Read` is very similar to a generic type parameter, which might be more straightforward to wirte

```rust
fn write_and_preprocess<R: Read>(&self, access: &str, data: &mut R) -> Result<..> {}
```

storage-proofs/benches/preprocessing.rs
laser: Why does that matter? This benchmark is for capturing deltas in performance of the bit padding operation across (code) changes to the operation itself.

sector-base/src/io/fr32.rs
```diff
@@ -635,20 +635,25 @@ pub fn clear_right_bits(byte: &mut u8, offset: usize) {
...
+    W: Read + Write + Seek + ?Sized,
 {
     // In order to optimize alignment in the common case of writing from an aligned start,
     // we should make the chunk a multiple of 127 (4 full elements, see `PaddingMap#alignment`).
-    // n was hand-tuned to do reasonably well in the benchmarks.
```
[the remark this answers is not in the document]
laser: yep!

storage-proofs/benches/preprocessing.rs
laser: Putting that aside for a moment: 

Note that `Read#read` is of type `read(&mut self, buf: &mut [u8]) -> Result<usize>`. Given that type, I see no way around cloning the vector.

filecoin-proofs/examples/ffi/main.rs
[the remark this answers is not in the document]
laser: Okay, okay, I won't unwrap - even in test code. Will update to use `expect` and `expects` with a nice error message.

sector-base/src/api/disk_backed_storage.rs
laser: Oh, I see what you're saying.

I can't use `impl Read` here because the `write_and_preprocess` method is on a trait which I use elsewhere as a trait object. We're using a trait object so that we can leak a pointer across the FFI and back.

PR #581, opened 2019-04-03
Update dependencies

dignifiedquire:
- update rust nightly version
- update all dependencies except for rand (because of ties to bellman/pairing etc) and proptest (because of compiler issues)
- removed `blake2` crate, and made all code consistently use only the simd implementations


dignifiedquire:
Hitting https://github.com/rust-lang/rust/issues/58375 and https://github.com/xd009642/tarpaulin/issues/161 trying to go back to an older version of rustc to see if that works for now

Issue #586, opened 2019-04-04
Store non-`bool` values in config


schomatis:
I need to store a path (`String` is fine) in the configuration 

https://github.com/filecoin-project/rust-fil-proofs/blob/b429279195e08ecf3abd9879a3ef6f3b6fb2e21f/storage-proofs/src/config.rs#L14

to set the directory where file-mapping MTs will be stored.

PR #587, opened 2019-04-04
store config as strings

sidke:
fixes: https://github.com/filecoin-project/rust-fil-proofs/issues/586

changes storage-proofs config to store configuration as string values
to get a boolean value out of the config, use `get_config_bool`

### DIFF
--- a/storage-proofs/src/config.rs
+++ b/storage-proofs/src/config.rs
@@ -7,11 +7,15 @@ use std::path::Path;
 use std::sync::Mutex;
 
 type Result<T> = ::std::result::Result<T, Error>;
-type Config = HashMap<String, bool>;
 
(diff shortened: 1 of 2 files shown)


schomatis:
> I guess this kicks the can down the road to when we first want, say, an integer config value. Maybe the best plan is just to lazily implement analogous getters as we occasionally need new types?

I agree that we'll be needing integers in the config but I'll prefer to merge the support for strings as is to help move along https://github.com/filecoin-project/rust-fil-proofs/pull/567.

porcuquine:
Yep.

schomatis:
:tada:

@sidke is this ready to merge? (or would you like to add something more here?)

PR #589, opened 2019-04-05
feat(storage_proofs): Add benchmark for encoding a single node

### LINKED ISSUE #530: Write benchmark on how long it takes to encode a single node
### Description
As part of understanding the Proof of Spacetime polling time we need to have clear how long it takes TODAY to encode a single node.

We should try to estimate with a calculator with best known hardware how long would it take to calculate a single node

### Where to begin
See https://github.com/filecoin-project/research/issues/94

vmx:
You can run it via `cargo bench -- "encode single node"`.

Closes #530.

### DIFF
--- a/storage-proofs/benches/encode.rs
+++ b/storage-proofs/benches/encode.rs
@@ -1,0 +1,109 @@
+#[macro_use]
+extern crate criterion;
+extern crate storage_proofs;
+
+use criterion::{black_box, Criterion, ParameterizedBenchmark};
+use pairing::bls12_381::Bls12;
+use rand::{thread_rng, Rng};
+use storage_proofs::fr32::fr_into_bytes;
+use storage_proofs::hasher::blake2s::Blake2sHasher;
+use storage_proofs::hasher::pedersen::PedersenHasher;
+use storage_proofs::hasher::sha256::Sha256Hasher;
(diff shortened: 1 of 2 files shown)


storage-proofs/benches/encode.rs
```diff
@@ -0,0 +1,93 @@
...
+
+fn encode_single_node_benchmark(cc: &mut Criterion) {
+    let degrees = vec![3, 5, 10];
+
+    cc.bench(
+        "encode single node",
+        ParameterizedBenchmark::new(
+            "Blake2s",
+            |b, degree| {
+                let Pregenerated {
```
dignifiedquire: I think rust likes lower case here

storage-proofs/benches/encode.rs
```diff
@@ -0,0 +1,93 @@
...
+                    mut data,
+                    parents,
+                    replica_id,
+                } = pregenerate_data::<Blake2sHasher>(*degree);
+                b.iter(|| {
```
dignifiedquire: It would be good to add a call to `black_box(data)` here, to ensure that this does not get optimized away.

porcuquine:
@vmx  This is failing on CI for `rustfmt`.

Issue #585, opened 2019-04-04
Add configuration value for NUM_PROVING_THREADS


dignifiedquire:
### Description

Same default as now, should be configurable through the config

### Where to begin

https://github.com/filecoin-project/rust-fil-proofs/blob/master/storage-proofs/src/compound_proof.rs#L12

PR #591, opened 2019-04-09
make NUM_PROVING_THREADS configurable

contrun:
This should fix https://github.com/filecoin-project/rust-fil-proofs/issues/585. But this is just a temporary solution, as people may be surprised the fact that `NUM_PROVING_THREADS` specified in the toml file should be a string. To make the config work for value other than string, we may have to change the config to a struct. Should I implement that or any other ideas?


dignifiedquire:
Lets get this in, we can iterate on the config format later.

PR #600, opened 2019-04-12
support sector class + dynamic PoSt and PoRep configuration

laser:
Blocks [this downstream go-filecoin PR](https://github.com/filecoin-project/go-filecoin/pull/2549).

## What's in this PR?

- Accept config values through FFI used to control PoRep partitions, PoSt partitions, and sector size
- Plumb `PoStConfig` and `PoRep` config into `internal.rs` functions to make choice of partitions and size dynamic

## Why does this PR exist?

- Initial push to allow FFI consumers to control proofs behavior ("sector class")
- Initial work to support multiple sector sizes


filecoin-proofs/src/api/internal.rs
```diff
@@ -418,10 +427,13 @@ fn verify_post_fixed_sectors_count(
         faults: fixed.faults.clone(),
     };
 
-    let verifying_key = get_post_verifying_key(fixed.sector_bytes)?;
+    let verifying_key = get_post_verifying_key(fixed.post_config)?;
 
-    let proof =
-        MultiProof::new_from_reader(Some(POST_PARTITIONS), &fixed.proof[0..192], &verifying_key)?;
+    let proof = MultiProof::new_from_reader(
+        Some(usize::from(PoStProofPartitions::from(fixed.post_config))),
```
laser: Note that PoSt proof partitions is now dynamic here.

filecoin-proofs/src/api/internal.rs
```diff
@@ -503,9 +515,9 @@ pub fn seal<T: Into<PathBuf> + AsRef<Path>>(
     let replica_id = replica_id::<DefaultTreeHasher>(prover_id, sector_id);
 
     let compound_setup_params = compound_proof::SetupParams {
-        vanilla_params: &setup_params(sector_config.sector_bytes()),
+        vanilla_params: &setup_params(PaddedBytesAmount::from(porep_config)),
         engine_params: &(*ENGINE_PARAMS),
-        partitions: Some(POREP_PARTITIONS),
+        partitions: Some(usize::from(PoRepProofPartitions::from(porep_config))),
```
laser: Note that PoRep proof partitions is now dynamic here.

filecoin-proofs/src/api/internal.rs
```diff
@@ -546,10 +558,12 @@ pub fn seal<T: Into<PathBuf> + AsRef<Path>>(
         &groth_params,
     )?;
 
+    // TODO: POREP_PROOF_BYTES needs to be a function of PoRepPartitions
```
laser: Note that PoRep proof bytes is still fixed, below.

filecoin-proofs/src/api/internal.rs
```diff
@@ -546,10 +558,12 @@ pub fn seal<T: Into<PathBuf> + AsRef<Path>>(
...
     let mut buf = Vec::with_capacity(POREP_PROOF_BYTES);
 
     proof.write(&mut buf)?;
 
+    // TODO: POREP_PROOF_BYTES needs to be a function of PoRepPartitions
```
laser: Note that PoRep proof bytes is also fixed on the next line.

filecoin-proofs/src/api/internal.rs
```diff
@@ -158,29 +160,31 @@ fn get_post_params(
...
+    let public_params = public_params(PaddedBytesAmount::from(porep_config));
 
     let get_verifying_key =
         || ZigZagCompound::verifying_key(&public_params, &ENGINE_PARAMS).map_err(Into::into);
 
     Ok(lookup_verifying_key(
-        format!("ZIGZAG[{}]", usize::from(sector_bytes)),
+        format!(
+            "ZIGZAG[{}]",
+            usize::from(PaddedBytesAmount::from(porep_config))
```
laser: Is it true that these keys will need to be updated to include PoRep partitions?

filecoin-proofs/src/api/internal.rs
```diff
@@ -192,7 +196,10 @@ fn get_post_verifying_key(
     };
 
     Ok(lookup_verifying_key(
-        format!("POST[{}]", usize::from(sector_bytes)),
+        format!(
+            "POST[{}]",
```
laser: Is it true that this key will need to be updated to include PoSt partitions?

filecoin-proofs/src/api/mod.rs
```diff
@@ -561,3 +530,64 @@ pub unsafe extern "C" fn get_staged_sectors(
...
+    sector_size: FFISectorSize,
+    seal_proof_partitions: FFISealProofPartitions,
+    post_proof_partitions: FFIPoStProofPartitions,
+}
+
+#[allow(non_camel_case_types)]
+#[repr(C)]
+pub enum FFISectorSize {
+    SSB_OneKiB,
+    SSB_TwoHundredFiftySixMiB,
```
porcuquine: I just want to check once more that we're comfortable with a model for change in which new sizes might be inserted between existing sizes, such that — for example — the underlying enum value `1` changes significance between releases. If these values go on chain anywhere, they will be more stably interpretable through time if we use the actual sizes. (I know we've discussed and you favor this approach, just calling this aspect out explicitly one last time.)

laser: > might be inserted between existing sizes

Yep, I'm with you. The position in the enum isn't meaningful in either go-filecoin or rust-fil-proofs.

filecoin-proofs/src/api/internal.rs
porcuquine: Yes, groth parameters will need to be fetched by both sector size and number of partitions.

filecoin-proofs/src/api/internal.rs
porcuquine: Yes, verifying keys will need to be fetched by both sector size and number of partitions.

PR #601, opened 2019-04-12
perf(storage_proofs): Allocate outside the parents() call

### LINKED ISSUE #575: storage-proofs: Avoid allocations when calling `parents`
### Description
Currently we call `drgraph::parents` in a loop resulting in allocating a `Vec` over and over. Instead `drgraph::parents` sholud take as argument `&mut [u8]`, into which the resulting parents are written. This way the allocations can be avoid in both the `BucketGraph` as well as the `ZigzagGraph` implementation of `parents`.

### Where to begin
- https://github.com/filecoin-project/rust-fil-proofs/blob/master/storage-proofs/src/vde.rs#L38
- https://github.com/filecoin-project/rust-fil-proofs/blob/master/storage-proofs/src/drgraph.rs#L120
- https://github.com/filecoin-project/rust-fil-proofs/blob/master/storage-proofs/src/zigzag_graph.rs#L152

vmx:
In order to save some allocations, the vector for parents is
allocated outside the call. This is on optimisation for loops
where that vector can be re-used in several parents() calls.

It's not clear that it really improves the performance. When
running the benchmark for encoding a single node, there's no
significant difference when this change is applied:

```
group                             HEAD2                                  less-alloc2
-----                             -----                                  -----------
encode single node/Blake2s/10     1.00     13.1±0.63µs        ? B/sec    1.01     13.2±0.62µs        ? B/sec
encode single node/Blake2s/3      1.00     11.8±0.52µs        ? B/sec    1.07     12.7±0.59µs        ? B/sec
encode single node/Blake2s/5      1.00     12.3±0.92µs        ? B/sec    1.05     12.8±0.77µs        ? B/sec
encode single node/Pedersen/10    1.00     11.8±0.61µs        ? B/sec    1.00     11.8±0.59µs        ? B/sec
encode single node/Pedersen/3     1.02     11.5±0.69µs        ? B/sec    1.00     11.3±0.47µs        ? B/sec
encode single node/Pedersen/5     1.00     11.5±0.65µs        ? B/sec    1.00     11.5±0.93µs        ? B/sec
encode single node/Sha256/10      1.00     12.0±0.77µs        ? B/sec    1.02     12.2±1.22µs        ? B/sec
encode single node/Sha256/3       1.01     11.3±0.52µs        ? B/sec    1.00     11.2±0.55µs        ? B/sec
encode single node/Sha256/5       1.00     11.4±0.56µs        ? B/sec    1.00     11.4±0.63µs        ? B/sec
```

Closes #575.

### DIFF
--- a/storage-proofs/src/drgraph.rs
+++ b/storage-proofs/src/drgraph.rs
@@ -70,7 +70,10 @@ pub trait Graph<H: Hasher>: ::std::fmt::Debug + Clone + PartialEq + Eq {
     /// If a node doesn't have any parents, then this vector needs to return a vector where
     /// the first element is the requested node. This will be used as indicator for nodes
     /// without parents.
-    fn parents(&self, node: usize) -> Vec<usize>;
+    ///
+    /// The `parents` parameter is used to store the result. This is done fore performance
+    /// reasons, so that the vector can be allocated outside this call.
+    fn parents(&self, node: usize, parents: &mut [usize]);
 
     /// Returns the size of the graph (number of nodes).
     fn size(&self) -> usize;
@@ -114,14 +117,20 @@ impl<H: Hasher> ParameterSetIdentifier for BucketGraph<H> {
 
 impl<H: Hasher> Graph<H> for BucketGraph<H> {
     #[inline]
-    fn parents(&self, node: usize) -> Vec<usize> {
+    fn parents(&self, node: usize, parents: &mut [usize]) {
         let m = self.base_degree;
 
         match node {
             // Special case for the first node, it self references.
-            0 => vec![0; m as usize],
             // Special case for the second node, it references only the first one.
-            1 => vec![0; m as usize],
+            0 | 1 => {
+                // Use the degree of the curren graph (`m`), as parents.len() might be bigger
+                // than that (that's the case for ZigZag Graph)
+                #[allow(clippy::needless_range_loop)]
+                for k in 0..m {
+                    parents[k] = 0;
+                }
+            }
             _ => {
                 // seed = self.seed | node
                 let mut seed = [0u32; 8];
@@ -129,7 +138,7 @@ impl<H: Hasher> Graph<H> for BucketGraph<H> {
                 seed[7] = node as u32;
                 let mut rng = ChaChaRng::from_seed(&seed);
 
-                let mut parents = Vec::with_capacity(m);
+                #[allow(clippy::needless_range_loop)]
                 for k in 0..m {
                     // iterate over m meta nodes of the ith real node
                     // simulate the edges that we would add from previous graph nodes
@@ -142,16 +151,16 @@ impl<H: Hasher> Graph<H> for BucketGraph<H> {
 
                     // remove self references and replace with reference to previous node
                     if out == node {
-                        parents.push(node - 1);
+                        parents[k] = node - 1;
                     } else {
                         assert!(out <= node);
-                        parents.push(out);
+                        parents[k] = out;
                     }
                 }
 
-                parents.sort_unstable();
-
-                parents
+                // Use the degree of the curren graph (`m`), as parents.len() might be bigger
+                // than that (that's the case for ZigZag Graph)
+                parents[0..m].sort_unstable();
             }
         }
     }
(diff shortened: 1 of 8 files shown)


porcuquine:
Thanks, @vmx.

Given that this doesn't show a performance improvement  and makes the code harder to reason about, I'm inclined not to make the change. I don't think we should avoid straightforward value-oriented/functional style except when it carries a significant performance benefit. #575 hypothesized that this would be the case here, but you've falsified that hypothesis.

My preference would be to close #575 with a reference to your benchmarks and the relevant code here. So your effort wasn't wasted, even if the knowledge it produced doesn't make it into the code. (We can probably learn from this example and not generally worry too much about ephemeral allocation of small vectors.)

What do you think?

vmx:
I fully agree. I'm just wondering if

 - we miss anything obvious why this isn't faster
 - the benchmark is no good. One problem I see with the benchmark is that it really has a tiny runtime. I'm unsure if you can actually measure meaning full things at that scale or if basically everything is just noise.

porcuquine:
You make a good point. Specifically, it's not clear to me (after a brief look) that the benchmark is even calling the function you changed here. As far as I can tell, your speculative optimization was to `parents`, but the benchmark seems to be pre-generating that data without even calling the function.

I may be restating your second point.

So maybe the thing to do is modify the benchmark to actually exercise these changes and see whether that improves the results.

storage-proofs/src/drgraph.rs
```diff
@@ -114,22 +117,28 @@ impl<H: Hasher> ParameterSetIdentifier for BucketGraph<H> {
...
-            0 => vec![0; m as usize],
             // Special case for the second node, it references only the first one.
-            1 => vec![0; m as usize],
+            0 | 1 => {
+                // Use the degree of the curren graph (`m`), as parents.len() might be bigger
+                // than that (that's the case for ZigZag Graph)
+                #[allow(clippy::needless_range_loop)]
+                for k in 0..m {
+                    parents[k] = 0;
+                }
```
dignifiedquire: You should be able to do `for k in parents.iter_mut().take(m)` that way clippy is happy

dignifiedquire:
Agreed with what has been said, the optimized path is not actually called inside the referenced benchmark.

dignifiedquire:
Maybe simply adding a benchmark which just measures the performance of `parents` would be good

vmx:
I totally missed that I'm not actually benchmarking this thing. So I agree that adding another benchmark makes sense, perhaps one that does something similar to `vde::encode()` (or even just calls `vde::encode()`). I'm happy to work on the benchmark, though it will wait until Friday in two weeks.

porcuquine:
Sounds good. No rush.

dignifiedquire:
I have a local benchmark trying some things out, and this gives a speed increase of ~30% when using max caching for a single node encoding. So this is def worth it

porcuquine:
Nice. Let's plan to merge this once CI is happy, then.

vmx:
I've added a new benchmark for measuring `parents()` only. I don't see any difference in performance, I wonder if I did something wrong. The new numbers are ("after" is with this change, "before" is without this change):

```
    group                              after                                  before
    -----                              -----                                  ------
    parents in a loop/Blake2s/10       1.00    135.4±4.02µs        ? B/sec    1.01    136.5±8.78µs        ? B/sec
    parents in a loop/Blake2s/1000     1.00      9.1±0.27ms        ? B/sec    1.01      9.2±0.18ms        ? B/sec
    parents in a loop/Blake2s/50       1.00   563.5±31.95µs        ? B/sec    1.00   561.1±17.98µs        ? B/sec
    parents in a loop/Pedersen/10      1.00    134.6±4.32µs        ? B/sec    1.02    137.1±4.35µs        ? B/sec
    parents in a loop/Pedersen/1000    1.00      9.0±0.18ms        ? B/sec    1.01      9.2±0.19ms        ? B/sec
    parents in a loop/Pedersen/50      1.01   560.3±26.94µs        ? B/sec    1.00   554.1±19.80µs        ? B/sec
    parents in a loop/Sha256/10        1.00    133.6±3.03µs        ? B/sec    1.00    133.9±4.52µs        ? B/sec
    parents in a loop/Sha256/1000      1.00      9.1±0.22ms        ? B/sec    1.02      9.2±0.23ms        ? B/sec
    parents in a loop/Sha256/50        1.00   549.7±15.63µs        ? B/sec    1.01   553.0±15.38µs        ? B/sec
```

vmx:
@dignifiedquire pointed out that I should run with `FIL_PROOFS_MAXIMIZE_CACHING=true`.

Now there is a significant difference in performance:

```
group                              after                                   before
-----                              ------                                  -------
parents in a loop/Blake2s/10       1.00      5.1±0.51µs        ? B/sec     1.33      6.8±0.58µs        ? B/sec
parents in a loop/Blake2s/1000     1.00 1011.4±694.48µs        ? B/sec     1.10 1109.6±509.07µs        ? B/sec
parents in a loop/Blake2s/50       1.00     30.8±3.06µs        ? B/sec     1.32     40.7±5.97µs        ? B/sec
parents in a loop/Pedersen/10      1.00      5.2±0.48µs        ? B/sec     1.37      7.1±0.72µs        ? B/sec
parents in a loop/Pedersen/1000    1.00  939.9±619.75µs        ? B/sec     1.17 1101.9±517.17µs        ? B/sec
parents in a loop/Pedersen/50      1.00     31.2±3.57µs        ? B/sec     1.36     42.5±3.81µs        ? B/sec
parents in a loop/Sha256/10        1.00      4.9±0.31µs        ? B/sec     1.37      6.7±0.50µs        ? B/sec
parents in a loop/Sha256/1000      1.00  918.5±560.68µs        ? B/sec     1.25 1149.1±549.77µs        ? B/sec
parents in a loop/Sha256/50        1.00     31.0±3.35µs        ? B/sec     1.34     41.6±4.80µs        ? B/sec
```

vmx:
Changing the for loops to iterators doesn't have a performance impact.

vmx:
I can't remember your merge policy. So anyone who feels responsible, please merge (if appropriate :)

Issue #592, opened 2019-04-10
Implement Piece Inclusion Proof


porcuquine:
### Description

Names below are descriptive. Sync with integration effort before finalizing.

The first four items were taken care of in #602 

  - [x] Rename legacy piece inclusion proof to `TrustedPieceInclusionProof`. [#602 renamed but then removed this as no longer relevant.]
  - [x] Implement `compute_comm_p`. [#602 adds `compute_piece_commitment`, `generate_piece_commitment`, and `generate_piece_commitment_bytes` providing various interfaces to `comm_p` generation.]
  - [x] Implement `create_piece_inclusion_proofs` for aligned pieces. [#602 implements `piece_inclusion_proofs`.].
  - [x] Implement `verify_piece_inclusion_proof` for aligned pieces. [#602 `PieceInclusionProof` has a `verify` method.]

---
  - [ ] Extend `verify_piece_inclusion_proof` to handle un-aligned pieces.
  - [ ] Extend `create_piece_inclusion_proofs` to handle un-aligned pieces.
  - [ ] Provide support functions for packing unaligned pieces such that correct piece inclusion proofs can be generated.

Note that for testnet the plan is to only pack aligned pieces. However, we want to be able to verify unaligned pieces to handle the general case of what miners might do.

The goal is to complete the first four items above quickly. With that in place, we should be able to proceed with integration and development as required for the storage market.

The items below the line are desirable but shouldn't affect miners only using our sector-packing code.

porcuquine:
cc @sidke @whyrusleeping @acruikshank @laser

dignifiedquire:
@porcuquine are you still working on this?

porcuquine:
It's paused but will need to be picked up later. The fully aligned PIP is implemented, and the interface should be stable. However, we support neither generating nor verifying un-aligned PIPs. We theoretically intend to support at least verification of unaligned PIPs — but that unfortunately will entail most or all the work of also generating.

I'll unassign myself but probably need to at least be in the loop when the work is restarted.

laser:
This work is complete and integrated with go-filecoin. I'm closing this issue.

Note:

> Extend create_piece_inclusion_proofs to handle un-aligned pieces.

[...]

PR #602, opened 2019-04-13
Piece Inclusion Proof (PIP)

porcuquine:
Implements first part of #592.

Probably needs some interface tweaks for integration.

For now, only aligned proofs are supported. However, my goal was to provide an interface compatible with later addition of unaligned proofs.

In particular,
  - Before packing, call `PieceSpec::compute_packing` to receive a packing list and proof size.
    - Packing list provides the ordering in which arbitrary piece segments should be added to the sector.
    - Proof size allows for comparison of alternate alignments to evaluate the cost (in proof size) of padding decisions.
  - Proofs always contain size and position of the piece. (This can be two 64-bit numbers. If we want to optimize for proof size of the aligned case, we can add some kind of type tag. If we want to compress somehow, we can.) With size and position, we have all the information needed to deterministically calculate a packing order and therefore a proof shape.
  - Proofs have a variable-size component (returned by `analyze`). The variable-size component is a sequence of `H::Domain` elements, which the verifier must be able to interpret.

### NOTE (cc: @dig)
 - verification requires both sector size and piece size. Otherwise a miner could just store `CommP` in a leaf and produce a valid proof that it is contained in the sector. (cc: @sidke)

**[UPDATE: I reworked a lot since writing the below. It might be possible to reintroduce the merkle-tree personalization without too much problem now. However, if it's awkward, I still think we can do without it.]**
 - ~I removed the 'personalization', which makes merkle-tree hashes depend on height in the tree. I did this so we can use normal merkle-proof verification to verify inclusion of CommP in CommD. Otherwise, we will need a modified proof verification which takes height as a parameter.~

   ~The most general form of PIP involves roots of sub-trees, proof that these are included in piece trees, and proofs of piece trees within sectors. We can most easily use the general merkle-tree apparatus if we don't also try to encode height in the hashing (though we do need to account for length of proof paths).~

   ~I don't think we lose anything by removing the hash 'personalization' or that we ever really needed it.~

### DIFF
--- a/storage-proofs/src/piece_inclusion_proof.rs
+++ b/storage-proofs/src/piece_inclusion_proof.rs
@@ -1,255 +1,486 @@
-use std::marker::PhantomData;
-
+use itertools::Itertools;
 use merkle_light::hash::Algorithm;
-use merkle_light::proof::Proof;
+use merkle_light::merkle::{self, next_pow2, VecStore};
 
 use crate::error::*;
+use crate::fr32::Fr32Vec;
 use crate::hasher::{Domain, Hasher};
 use crate::merkle::MerkleTree;
 
-type InclusionProof<T> = Proof<T>;
-
-/// A FileInclusionProof contains a merkle inclusion proof for the first and last node
-/// of a piece. This ensures all 'edge' hashes necessary to generate a complete merkle
-/// tree are available.
-///
-/// Depending on the position of the nodes, not every hash provided will actually be needed.
-/// As a space optimization, and at the cost of greater complexity in the encoding, 'interior' nodes
-/// of either path may be omitted.
+/// Based on the alignment information (and sector size, provided during verification),
+/// the algorithm deterministically consumes the elements.
+#[derive(Clone, Debug)]
 pub struct PieceInclusionProof<H: Hasher> {
-    first_node_proof: InclusionProof<H::Domain>,
-    last_node_proof: InclusionProof<H::Domain>,
-    _h: PhantomData<H>,
+    position: usize,
+    piece_leaves: usize,
+    proof_elements: Vec<H::Domain>,
 }
 
-/// file_inclusion_proofs takes a merkle tree and a slice of piece lengths, and returns
-/// a vector of file inclusion proofs corresponding to the pieces. This assumes that the first
-/// piece begins at offset 0, and that each piece begins directly after the previous piece ends.
-/// For this method to work, the piece data used to validate pieces will need to be padded as necessary,
-/// and pieces will need to be aligned (to 128-byte chunks for Fr32 bit-padding) when written.
-pub fn file_inclusion_proofs<H: Hasher>(
+/// `position`, `length` are in H::Domain units
+#[derive(Clone, Debug)]
+pub struct PieceSpec<'a> {
+    comm_p: &'a Fr32Vec,
+    position: usize,
+    length: usize,
+}
+
+impl<'a> PieceSpec<'a> {
+    /// `compute_packing` returns a packing list and a proof size.
+    /// A packing list is a pair of (start, length) pairs, relative to the beginning of the piece,
+    /// in leaf units.
+    /// Proof size is a number of elements (size same as one leaf) provided in the variable part of a PieceInclusionProof.
+    pub fn compute_packing(&self, tree_len: usize) -> Result<(Vec<(usize, usize)>, usize)> {
+        if !self.is_aligned(tree_len) {
+            Err(Error::UnalignedPiece)
+        } else {
+            let packing_list = vec![(0, self.length)];
+            Ok((packing_list, self.proof_length(tree_len)))
+        }
+    }
+
+    pub fn is_aligned(&self, tree_len: usize) -> bool {
+        piece_is_aligned(self.position, self.length, tree_len)
+    }
+
+    fn height(&self) -> usize {
+        height_for_length(self.length)
+    }
+
+    // `proof_length` is length of proof that comm_p is in the containing root, excluding comm_p and root, which aren't needed for the proof itself.
+    fn proof_length(&self, tree_len: usize) -> usize {
+        height_for_length(tree_len) - self.height()
+    }
+}
+
+#[allow(dead_code)]
+fn create_piece_tree<H: Hasher>(
+    data: &[H::Domain],
+) -> merkle::MerkleTree<H::Domain, H::Function, VecStore<H::Domain>> {
+    let data_size = data.len();
+    // We need to compute comm_p as a merkle root over power-of-two-sized data.
+    let tree_size = next_pow2(data_size);
+
+    // If actual data is less than tree size, pad it with zeroes.
+    //if data_size < tree_size {
+    // NOTE: this assumes that `H::Domain::default()` corresponds to zeroed input.
+    // This matters because padding may have been applied at the byte level, using zero bytes.
+    merkle::MerkleTree::<H::Domain, H::Function, VecStore<H::Domain>>::new(
+        data.iter()
+            .cloned()
+            .pad_using(tree_size, |_| H::Domain::default()),
+    )
+}
+
+#[allow(dead_code)]
+/// Compute `comm_p` from a slice of Domain elements.
+/// `comm_p` is the merkle root of a piece, zero-padded to fill a complete binary sub-tree.
+fn compute_piece_commitment<H: Hasher>(data: &[H::Domain]) -> H::Domain {
+    create_piece_tree::<H>(data).root()
+}
+
+#[allow(dead_code)]
+/// Generate `comm_p` from bytes
+fn generate_piece_commitment<H: Hasher>(data: &[u8]) -> Result<H::Domain> {
+    let mut domain_data = Vec::new();
+    for d in data.chunks(32) {
+        domain_data.push(<H::Domain as Domain>::try_from_bytes(d)?)
+    }
+
+    Ok(compute_piece_commitment::<H>(&domain_data))
+}
+
+#[allow(dead_code)]
+/// Generate `comm_p` from bytes and return it as bytes.
+fn generate_piece_commitment_bytes<H: Hasher>(data: &[u8]) -> Result<Fr32Vec> {
+    let comm_p = generate_piece_commitment::<H>(data)?;
+
+    Ok(H::Domain::into_bytes(&comm_p))
+}
+
+pub fn piece_inclusion_proofs<H: Hasher>(
+    piece_specs: &[PieceSpec],
     tree: &MerkleTree<H::Domain, H::Function>,
-    piece_lengths: &[usize],
-) -> Vec<PieceInclusionProof<H>> {
-    bounds(piece_lengths)
+) -> Result<Vec<PieceInclusionProof<H>>> {
+    piece_specs
         .iter()
-        .map(|(start, end)| file_inclusion_proof(tree, *start, end - 1))
+        .map(|piece_spec| PieceInclusionProof::new(piece_spec.clone(), tree))
         .collect()
 }
 
-// Given a set of lengths, return corresponding (start, end) pairs for successive pieces.
-fn bounds(lengths: &[usize]) -> Vec<(usize, usize)> {
-    let mut start = 0;
-    let mut bounds = Vec::with_capacity(lengths.len());
-    for length in lengths {
-        let end = start + length;
-        bounds.push((start, end));
-        start = end;
-    }
-    bounds
-}
-
-/// file_inclusion_proof takes a merkle tree and the index positions of the first and last nodes
-/// of the piece whose inclusion should be proved. It returns a corresponding file_inclusion_proof.
-/// For the resulting proof to be valid, first_node must be <= last_node.
-pub fn file_inclusion_proof<H: Hasher>(
-    tree: &MerkleTree<H::Domain, H::Function>,
-    first_node: usize,
-    last_node: usize,
-) -> PieceInclusionProof<H> {
-    PieceInclusionProof {
-        first_node_proof: tree.gen_proof(first_node),
-        last_node_proof: tree.gen_proof(last_node),
-        _h: PhantomData,
-    }
-}
-
 impl<H: Hasher> PieceInclusionProof<H> {
+    pub fn new(
+        piece_spec: PieceSpec,
+        tree: &MerkleTree<H::Domain, H::Function>,
+    ) -> Result<PieceInclusionProof<H>> {
+        let PieceSpec {
+            comm_p,
+            position: first_leaf,
+            length: leaf_count,
+        } = piece_spec;
+
+        let last_leaf = first_leaf + (leaf_count - 1);
+
+        // For now, we only handled aligned pieces.
+        if !piece_spec.is_aligned(tree.len()) {
+            return Err(Error::UnalignedPiece);
+        }
+
+        let first_proof = tree.gen_proof(first_leaf);
+        let last_proof = tree.gen_proof(last_leaf);
+
+        // Including first leaf (item).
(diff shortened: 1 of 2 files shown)


storage-proofs/src/circuit/porc.rs
```diff
@@ -223,7 +223,7 @@ impl<'a, E: JubjubEngine> Circuit<E> for PoRCCircuit<'a, E> {
                 // Compute the new subtree value
                 cur = pedersen_hash::pedersen_hash(
                     cs.namespace(|| "computation of pedersen hash"),
-                    pedersen_hash::Personalization::MerkleTree(i),
+                    pedersen_hash::Personalization::NoteCommitment,
```
dignifiedquire: I always thought we need this, for the same reason we need this to avoid 2nd preimage attacks in the regular merkle tree building. Could you explain to me why that is not the case?

storage-proofs/src/merkle.rs
```diff
@@ -24,11 +24,11 @@ pub type VecMerkleTree<T, A> = merkle::MerkleTree<T, A, VecStore<T>>;
...
 pub struct MerkleProof<H: Hasher> {
     pub root: H::Domain,
-    path: Vec<(H::Domain, bool)>,
-    leaf: H::Domain,
+    pub path: Vec<(H::Domain, bool)>,
+    pub leaf: H::Domain,
 
     #[serde(skip)]
-    _h: PhantomData<H>,
+    pub _h: PhantomData<H>,
```
dignifiedquire: Instead of making this public, can we have a simple constructor that takes care of this, like we do for other structs with phantom data?

storage-proofs/src/piece_inclusion_proof.rs
```diff
@@ -1,255 +1,473 @@
...
+use merkle_light::proof::Proof;
 
-/// A FileInclusionProof contains a merkle inclusion proof for the first and last node
-/// of a piece. This ensures all 'edge' hashes necessary to generate a complete merkle
-/// tree are available.
-///
-/// Depending on the position of the nodes, not every hash provided will actually be needed.
-/// As a space optimization, and at the cost of greater complexity in the encoding, 'interior' nodes
-/// of either path may be omitted.
+// TODO: A PieceInclusionProof should be a position, a size, and a series of `H::Domain` elements.
```
dignifiedquire: Why is this a `TODO`? Seems to me that is already the case

storage-proofs/src/piece_inclusion_proof.rs
```diff
@@ -1,255 +1,473 @@
...
-    for length in lengths {
-        let end = start + length;
-        bounds.push((start, end));
-        start = end;
+impl<'a> PieceSpec<'a> {
+    /// `analyze` returns a packing list and a proof size.
+    /// A packing list is a pair of (start, length) pairs, relative to the beginning of the piece,
+    /// in leaf units.
+    /// Proof size is a number of elements (size same as one leaf) provided in the variable part of a PieceInclusionProof.
+    pub fn analyze(&self, tree_len: usize) -> Result<(Vec<(usize, usize)>, usize)> {
```
dignifiedquire: `analyze` seems to be an odd name for this, is that a predefined term for this?

storage-proofs/src/piece_inclusion_proof.rs
```diff
@@ -1,255 +1,473 @@
...
+
+    pub fn piece_is_aligned(position: usize, length: usize, tree_len: usize) -> bool {
+        let capacity_at_pos = Self::subtree_capacity(position, tree_len);
 
-    let mut hasher = H::Function::default();
+        Self::is_pow2(capacity_at_pos) && capacity_at_pos >= length
+    }
 
-    let mut last_row = Vec::new();
+    fn height_for_length(n: usize) -> usize {
```
dignifiedquire: could be a standalone method

storage-proofs/src/piece_inclusion_proof.rs
```diff
@@ -1,255 +1,473 @@
...
-    // Zip the left and right proof_vecs into one.
-    let proof_vecs = proof_vec(left_proof).zip(proof_vec(right_proof));
+    // `proof_length` is length of proof that comm_p is in the containing root, excluding comm_p and root, which aren't needed for the proof itself.
+    fn proof_length(&self, tree_len: usize) -> usize {
+        Self::height_for_length(tree_len) - self.height()
+    }
+
+    // Utility
+
+    pub fn piece_is_aligned(position: usize, length: usize, tree_len: usize) -> bool {
```
dignifiedquire: could be a standalone method

storage-proofs/src/piece_inclusion_proof.rs
```diff
@@ -1,255 +1,473 @@
...
+    }
 
-    let mut last_row = Vec::new();
+    fn height_for_length(n: usize) -> usize {
+        (n as f64).log2().ceil() as usize
+    }
 
-    for chunk in piece_data.chunks(32) {
-        last_row.push(H::Domain::try_from_bytes(chunk)?);
+    fn is_pow2(n: usize) -> bool {
```
dignifiedquire: could be a standalone method

storage-proofs/src/piece_inclusion_proof.rs
```diff
@@ -1,255 +1,473 @@
...
 
-    for (height, ((l_hash, l_is_left), (r_hash, r_is_left))) in proof_vecs.enumerate() {
-        let mut row = Vec::new();
-        if !*l_is_left {
-            row.push(*l_hash);
-        };
-        row.extend(last_row);
-        if *r_is_left {
-            row.push(*r_hash);
+    fn subtree_capacity(pos: usize, total: usize) -> usize {
```
dignifiedquire: could be a standalone method

storage-proofs/src/piece_inclusion_proof.rs
```diff
@@ -1,255 +1,473 @@
...
+        assert!(pos < total, "position must be less than tree capacity");
+
+        let mut capacity = 1;
+        // If tree is not 'full', then pos 0 will have subtree_capacity greater than size of tree.
+        let mut cursor = pos + next_pow2(total);
+
+        while cursor & 1 == 0 {
```
dignifiedquire: Could you add a commment what the shifting and anding does here, so the intent is clearer.

storage-proofs/src/piece_inclusion_proof.rs
```diff
@@ -1,255 +1,473 @@
...
+fn create_piece_tree<H: Hasher>(
+    data: &[H::Domain],
+) -> merkle::MerkleTree<H::Domain, H::Function, VecStore<H::Domain>> {
+    let data_size = data.len();
+    // We need to compute comm_p as a merkle root over power-of-two-sized data.
+    let tree_size = next_pow2(data_size);
+
+    // If actual data is less than tree size, pad it with zeroes.
+    if data_size < tree_size {
+        let zero = H::Domain::try_from_bytes(&[0; 32]).unwrap();
```
dignifiedquire: Can't you use `H::Domain::default()` for this?

storage-proofs/src/piece_inclusion_proof.rs
```diff
@@ -1,255 +1,473 @@
...
+        let mut data_copy = vec![zero; tree_size];
+        data_copy[0..data_size].copy_from_slice(data);
+        merkle::MerkleTree::<H::Domain, H::Function, VecStore<H::Domain>>::new(data_copy)
+    } else {
+        merkle::MerkleTree::<H::Domain, H::Function, VecStore<H::Domain>>::new(data.to_vec())
```
dignifiedquire: Do we not have a way to create a merkle tree, without allocating a new vec? This just needs to be iterated over afaik.

storage-proofs/src/piece_inclusion_proof.rs
```diff
@@ -1,255 +1,473 @@
...
+    }
+}
 
-    hashed
+#[allow(dead_code)]
+/// Compute comm_p from a slice of Domain elements
```
dignifiedquire: Could you add a note describing what `comm_p` actually is

dignifiedquire: Also, if we only have one kind of commitment, naming these methods `compute_commitment` and `generate_commitment` seems to be nicer

storage-proofs/src/piece_inclusion_proof.rs
```diff
@@ -1,255 +1,473 @@
...
 
-/// Return a vector of (hash, bool) pairs, where the bool indicates whether the paired hash
-/// will be the left child of the next node hashed. In order to accomplish this, we skip
-/// the first hash provided (in proof.lemma()). This is an implementation detail of how a
-/// merkle_light::proof::Proof is structured.
-fn proof_vec<T: Domain>(proof: &Proof<T>) -> impl Iterator<Item = (&T, &bool)> {
-    proof.lemma().iter().skip(1).zip(proof.path().iter())
+#[allow(dead_code)]
+/// Generate comm_p from bytes
+fn generate_comm_p<H: Hasher>(data: &[u8]) -> Result<H::Domain> {
```
dignifiedquire: The naming seems odd, maybe `commitment_from_domain` and `commitment_from_bytes`, would make the intent of these methods clearer?

storage-proofs/src/piece_inclusion_proof.rs
```diff
@@ -1,255 +1,473 @@
...
+
+pub fn piece_inclusion_proofs<H: Hasher>(
+    piece_specs: &[PieceSpec],
+    tree: &MerkleTree<H::Domain, H::Function>,
+) -> Result<Vec<PieceInclusionProof<H>>> {
+    piece_specs
         .iter()
-        .zip(pieces)
-        .all(|(proof, piece)| proof.verify(root, piece))
+        .map(|piece_spec| PieceInclusionProof::new((*piece_spec).clone(), tree))
```
dignifiedquire: No need to dereference, `clone` should do everything you need.

storage-proofs/src/piece_inclusion_proof.rs
```diff
@@ -1,255 +1,473 @@
...
+        // For an aligned piece, there is guaranteed to be a common element.
+        // That is, the merkle root of the piece must be a node of the tree (and therefore in the paths
+        // of all leaves of the piece (including the first and last).
+        for (i, (a, b)) in first_proof
+            .lemma()
+            .iter()
+            .zip(last_proof.lemma().iter())
+            .enumerate()
+        {
+            if (a == b) && (i == proof_length) {
```
dignifiedquire: why do we check for `i == proof_length` here, in addition to the match?

storage-proofs/src/piece_inclusion_proof.rs
```diff
@@ -1,255 +1,473 @@
...
+                let proof_elements = if i > first_proof.path().len() {
+                    vec![]
+                } else {
+                    let mut lemma = first_proof.lemma()[i..].to_vec();
+                    lemma.insert(0, H::Domain::try_from_bytes(&comm_p)?);
```
dignifiedquire: this could be a bit more efficient

```rust
let mut lemma = Vec::with_capacity(first_proof.lemma().len() - i + 1);
lemma.push(H::Domain::try_from_bytes(&comm_p)?);
lemma.extend_from_slice(&first_proof.lemma()[i..]);
```

storage-proofs/src/piece_inclusion_proof.rs
```diff
@@ -1,255 +1,473 @@
...
+        Err(Error::MalformedMerkleTree)
+    }
+
+    fn leaf_path(leaf: usize, tree_height: usize) -> Vec<bool> {
+        let mut height = tree_height;
+        let mut rising_leaf = leaf;
+        let mut path = Vec::new();
+
+        while height > 0 {
+            // If this leaf's 1-bit is 1, then it is a right branch, so `is_right` is true.
```
dignifiedquire: `0-bit is 1` I believe

storage-proofs/src/piece_inclusion_proof.rs
```diff
@@ -1,255 +1,473 @@
...
+        path
+    }
+
+    /// verify takes a merkle root and (pre-processed) piece data.
+    /// Iff it returns true, then PieceInclusionProof indeed proves that piece's
+    /// bytes were included in the merkle tree corresponding to root -- and at the
+    /// position encoded in the proof.
+    ///
+    /// `piece_leaves` and `sector_leaves` are in units of `Domain` (i.e. `NODE_SIZE` = 32 bytes).
+
```
dignifiedquire: empty line makes doc comments to be attached :(

storage-proofs/src/circuit/porc.rs
porcuquine: I'm going to try to reinstate it just to minimize change.

However, I think the better question is why we *do* need it. I think the 'attack' this prevents is that without it a miner can arrange for collisions of `CommD` computed over sectors of different sizes (by using hashes from an intermediate row as leaves of a smaller sector).

storage-proofs/src/piece_inclusion_proof.rs
porcuquine: Imagine a piece filled with identical leaves. Proofs of such a piece may have early matches.

storage-proofs/src/piece_inclusion_proof.rs
porcuquine: Actually, the whole song and dance can be simplified out of existence as the last stage of refactoring. Fix incoming.

storage-proofs/src/circuit/porc.rs
dignifiedquire: > I think the better question is why we _do_ need it.

I agree, I just would like to be sure that we don't need it before ripping it out.

storage-proofs/src/piece_inclusion_proof.rs
porcuquine: I meant 1-bit as in one's place (as opposed to 2-bit, 4-bit, etc.). Rewording to avoid the issue.

storage-proofs/src/circuit/porc.rs
porcuquine: After I reworked everything, the initial awkwardness became tractable so I backed this change out. In case it comes up again, I still think we can live without the personalization as long as we don't try to distinguish different-sized sectors by their data commitment (which we should not do).

storage-proofs/src/piece_inclusion_proof.rs
```diff
@@ -1,255 +1,486 @@
...
-    tree: &MerkleTree<H::Domain, H::Function>,
-    piece_lengths: &[usize],
-) -> Vec<PieceInclusionProof<H>> {
-    bounds(piece_lengths)
-        .iter()
-        .map(|(start, end)| file_inclusion_proof(tree, *start, end - 1))
-        .collect()
+/// `position`, `length` are in H::Domain units
+#[derive(Clone)]
+pub struct PieceSpec<'a> {
```
dignifiedquire: please derive `Debug` as well

storage-proofs/src/piece_inclusion_proof.rs
```diff
@@ -1,255 +1,486 @@
...
-/// A FileInclusionProof contains a merkle inclusion proof for the first and last node
-/// of a piece. This ensures all 'edge' hashes necessary to generate a complete merkle
-/// tree are available.
-///
-/// Depending on the position of the nodes, not every hash provided will actually be needed.
-/// As a space optimization, and at the cost of greater complexity in the encoding, 'interior' nodes
-/// of either path may be omitted.
+/// Based on the alignment information (and sector size, provided during verification),
+/// the algorithm deterministically consumes the elements.
 pub struct PieceInclusionProof<H: Hasher> {
```
dignifiedquire: please derive `Debug` as well

PR #603, opened 2019-04-15
feat(filecoin-proofs): implement rocksdb based kv

### LINKED ISSUE #351: replace filesystem-backed k/v store with rocksdb store
rust-fil-proofs `SectorBuilder` metadata is persisted/loaded from a simple datastore which I hacked together in a day.

dignifiedquire:
Closes #351

### DIFF
--- a/filecoin-proofs/src/api/sector_builder/kv_store/mod.rs
+++ b/filecoin-proofs/src/api/sector_builder/kv_store/mod.rs
@@ -22,7 +22,6 @@ mod tests {
     #[test]
     fn test_alpha() {
         let metadata_dir = tempfile::tempdir().unwrap();
-        println!("dir: {:?}", metadata_dir);
         let db = FileSystemKvs::initialize(metadata_dir).unwrap();
 
         let k_a = b"key-xx";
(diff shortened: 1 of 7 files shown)


filecoin-proofs/src/api/sector_builder/kv_store/mod.rs
```diff
@@ -15,7 +22,7 @@ mod tests {
     #[test]
     fn test_alpha() {
         let metadata_dir = tempfile::tempdir().unwrap();
-
+        println!("dir: {:?}", metadata_dir);
```
laser: Errant `println` here :)

laser:
Looks good. I'm glad that somebody finally got around to doing this.

filecoin-proofs/src/api/sector_builder/kv_store/mod.rs
dignifiedquire: 😵

dignifiedquire:
@laser one question from my side, this currently does not include any migration work. So existing installs will loose their data when switching to this code. Is that an issue?

filecoin-proofs/src/api/sector_builder/kv_store/mod.rs
dignifiedquire: Fixed

PR #604, opened 2019-04-15
refactor(config): typed configuration

### LINKED ISSUE #599: feat: improve configuration
### Description

Currently we can only store strings in the configuration, but we should be able to store at least `bool`, `string` and `usize`. 

### Acceptance criteria

Should use a proper config struct that parses and validates configuration options set.

dignifiedquire:
- switch to using the `config` crate
- rename config to settings, to avoid naming conflicts
- use a typed config, allowing for type safety and parsing using serde
- **Breaking Change:** prefix config variables with `FIL_PROOFS`

Closes #599


storage-proofs/src/settings.rs
```diff
@@ -0,0 +1,39 @@
...
+    }
+}
+
+impl Settings {
+    fn new() -> Result<Settings, ConfigError> {
+        let mut s = Config::new();
+
+        s.set_defaults(&Self::default())?;
+        s.merge(File::with_name(SETTINGS_PATH).required(false))?;
+        s.merge(Environment::with_prefix("FIL_PROOFS"))?;
```
schomatis: @dignifiedquire Could we add some documentation for this to make it more visible?

PR #605, opened 2019-04-16
Add CompoundProof Requirements + ChallengeRequirements

porcuquine:
This lays the groundwork for ensuring proofs meets some minimum challenge requirement, regardless of how many partitions are used. Previously we had no way of enforcing aggregate requirements. This corrects that.

It adds a general mechanism allowing `Requirements` to be specified to any `ProofScheme`. These requirements are passed as a parameter to a new `satisifies_requirements` trait method. This method also receives public parameters and number of partitions. This allows it to determine whether the proof's public parameters meet requirements when divided into a given number of partitions.

Also included is a simple `ChallengeRequirements` impelementation for layered proofs. This ensures that the total number of challenges across all partitions is at least some minimum. For now, I set the minimum for all PoReps created from the API at 1. So this should never fail. Eventually, this should prevent us from creating configurations (or somehow accidentally attempting to verify proofs) which don't meet challenge security.

Testing this will be a bit annoying, so I leave that for later when we have something meaningful to test.

### DIFF
--- a/storage-proofs/src/proof.rs
+++ b/storage-proofs/src/proof.rs
@@ -13,6 +13,7 @@ pub trait ProofScheme<'a> {
     type PublicInputs: Clone;
     type PrivateInputs;
     type Proof: Clone + Serialize + DeserializeOwned;
+    type Requirements: Default;
 
     /// setup is used to generate public parameters from setup parameters in order to specialize
     /// a ProofScheme to the specific parameters required by a consumer.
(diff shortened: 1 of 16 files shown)


storage-proofs/src/proof.rs
```diff
@@ -88,4 +89,15 @@ pub trait ProofScheme<'a> {
...
+        _pub_params: &Self::PublicParams,
+        _requirements: &Self::Requirements,
+        _partitions: usize,
+    ) -> bool {
+        true
+    }
 }
+
+#[derive(Default)]
+pub struct NoRequirements();
```
dignifiedquire: You can make this an empty struct: `pub struct NoRequirements`, that way you should be able to pass it as `my_fn(&NoRequirements)` instead of the `my_fn(&NoRequirements())`

PR #609, opened 2019-04-16
Encoding improvements

dignifiedquire:
- improve benchmarks for encoding and hashing
- pulls in #601

### DIFF
--- a/storage-proofs/benches/encode.rs
+++ b/storage-proofs/benches/encode.rs
@@ -1,10 +1,13 @@
 #[macro_use]
 extern crate criterion;
+#[cfg(feature = "cpu-profile")]
+extern crate gperftools;
 extern crate storage_proofs;
 
-use criterion::{black_box, Criterion, ParameterizedBenchmark};
+use criterion::{black_box, Criterion, ParameterizedBenchmark, Throughput};
 use pairing::bls12_381::Bls12;
 use rand::{thread_rng, Rng};
+use storage_proofs::drgraph::{new_seed, Graph};
 use storage_proofs::fr32::fr_into_bytes;
 use storage_proofs::hasher::blake2s::Blake2sHasher;
 use storage_proofs::hasher::pedersen::PedersenHasher;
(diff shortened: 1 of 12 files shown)


storage-proofs/benches/encode.rs
```diff
@@ -24,32 +56,36 @@ fn pregenerate_data<H: Hasher>(degree: usize) -> Pregenerated<H> {
...
 
     let node_data = H::Domain::try_from_bytes(&data[start..end]).unwrap();
-    let encoded = H::sloth_encode(&key, &node_data, 1);
+    let encoded = H::sloth_encode(&key, &node_data, 0);
     encoded.write_bytes(&mut data[start..end]).unwrap();
 }
 
 fn encode_single_node_benchmark(cc: &mut Criterion) {
-    let degrees = vec![3, 5, 10];
+    let degrees = vec![3, 5];
```
porcuquine: Can we include degree 13, which is our current real value. That way we can get a direct measurement of the cost of encoding one real node.

dignifiedquire: this is the base degree, and it uses an expansion degree of 8, so I believe we are already doing that

porcuquine: I see. I missed the `8` hiding on line 60.

vmx:
In https://github.com/filecoin-project/rust-fil-proofs/pull/601#issuecomment-483210583 we talk about adding a benchmark for the `parents()` call. With this PR the single node encoding benchmark now also calls `parents()` so I wonder if such a new benchmark is needed. Please advice.

porcuquine:
@vmx Yep, let's get this merged, then you can rebase and run your benchmarks again. Based on @dignifiedquire's previous comments, that should show the intended benefit.

vmx:
@porcuquine I don't have a new benchmark. Anyway I'll try to re-rerun the modified benchmark from this PR with the old code base and look if we see a difference.

porcuquine:
@vmx Yes, I think that's the right approach.

I think we want the benchmarking improvements from this PR, but if the codes changes (here) whose sole purpose is performance (with a complexity cost) don't actually improve it, I don't think we should introduce them. Same goes with yours.

vmx:
@porcuquine OK, let me then try this PR locally and see if the modified benchmark catches differences.

vmx:
@porcuquine I did three subsequent runs of both versions (with and without my change), here's the comparison from the two fastest runs:

```
group                            after3                                 before2
-----                            ------                                 -------
encode single node/Blake2s/3     1.00      8.6±0.40µs     3.6 MB/sec    1.01      8.7±0.46µs     3.5 MB/sec
encode single node/Blake2s/5     1.00      7.8±0.26µs     3.9 MB/sec    1.11      8.7±0.40µs     3.5 MB/sec
encode single node/Pedersen/3    1.00      8.5±0.39µs     3.6 MB/sec    1.01      8.7±0.35µs     3.5 MB/sec
encode single node/Pedersen/5    1.06      8.5±0.56µs     3.6 MB/sec    1.00      8.0±0.35µs     3.8 MB/sec
encode single node/Sha256/3      1.00      8.4±0.36µs     3.6 MB/sec    1.07      9.1±0.50µs     3.4 MB/sec
encode single node/Sha256/5      1.00      7.8±0.37µs     3.9 MB/sec    1.04      8.1±0.28µs     3.8 MB/sec
```

I wouldn't say that's significant. Please note that there's only a single call to `parents()`. What my change should improve is calling to parents in a loop.

My conclusion: we don't merge this PR for now and wait to have a benchmark that is calling `parents()` in a loop. I'll open a separate issue for that so we can discuss the parameters for that benchmark, as I don't know how many iterations such a `parents()` loop usually has.

dignifiedquire:
closing, as the important parts were merged a while back

Issue #422, opened 2019-01-02
Multiple Sector Sizes


porcuquine:
Spec proposal for multiple sector sizes (interfaces, circuit changes) exists and, if ratified, is implemented

 - [x] Spec Proposal
 - [x] Review
 - [x] Ratified
 - [ ] Implement Proposal

porcuquine:
Proposal work in progress: https://github.com/filecoin-project/specs/issues/175

porcuquine:
No longer WIP: in review.

laser:
Things which are done:

- a storage miner actor is pinned to a sector size when created
- the protocol allows for multiple sector sizes
- proving period duration, generation attack threshold, and collateral are now (stubbed) functions of sector size

Outstanding issues:

- [need to resolve this specs PR](https://github.com/filecoin-project/specs/pull/318), which decides whether or not the protocol defines an enumeration of supported (sector size, proof partitions)
- [this bug](https://github.com/filecoin-project/rust-fil-proofs/issues/611)

Things which are not done which I have no plans to do anytime soon:

- any new software to support "verified single tenancy"

dignifiedquire:
we are now shipping with multiple sector sizes

PR #621, opened 2019-04-18
Calculate challenges per partition, ensuring total minimum is met.

porcuquine:
Toward #422.

Included in this PR:

 - Calculate PoRep challenges based on number of partitions, such that minimum total challenges are met. The minimum is a constant across all sector sizes and proof sizes.
 - Generate parameters for all valid sector-size/partitions combinations. As of now, all combinations are valid, so all are generated.
 - Sort parameters when publishing so re-publishing without changes never produces a diff. (cc: @sidke).
 - Reduce default value of `NUM_PROVING_THREADS` to 1, since larger values may be problematic on machines with few cores. (This will begin to matter as we add partitions.)

NOTE: I published parameters here to verify that they had *not* changed. However, because `parameters.json` did not previously have a canonical ordering, this produced a diff in the changeset. I leave this in to ensure the manifest and code are in sync — but my intention is for the current work not to affect parameters or actual proofs.

### DIFF
--- a/filecoin-proofs/src/api/internal.rs
+++ b/filecoin-proofs/src/api/internal.rs
@@ -47,8 +47,6 @@ pub type ChallengeSeed = Fr32Ary;
 /// FrSafe is an array of the largest whole number of bytes guaranteed not to overflow the field.
 type FrSafe = [u8; 31];
 
-const POREP_MINIMUM_CHALLENGES: usize = 1; // FIXME: 8,000
-
 lazy_static! {
     pub static ref ENGINE_PARAMS: JubjubBls12 = JubjubBls12::new();
 }
@@ -122,7 +120,10 @@ where
 ////////////////////////////////////////////////////////////////////////////////
 
 fn get_zigzag_params(porep_config: PoRepConfig) -> error::Result<Arc<groth16::Parameters<Bls12>>> {
-    let public_params = public_params(PaddedBytesAmount::from(porep_config));
+    let public_params = public_params(
+        PaddedBytesAmount::from(porep_config),
+        usize::from(PoRepProofPartitions::from(porep_config)),
+    );
 
     let get_params =
         || ZigZagCompound::groth_params(&public_params, &ENGINE_PARAMS).map_err(Into::into);
@@ -158,7 +159,10 @@ fn get_post_params(post_config: PoStConfig) -> error::Result<Arc<groth16::Parame
 }
 
 fn get_zigzag_verifying_key(porep_config: PoRepConfig) -> error::Result<Arc<Bls12VerifyingKey>> {
-    let public_params = public_params(PaddedBytesAmount::from(porep_config));
+    let public_params = public_params(
+        PaddedBytesAmount::from(porep_config),
+        usize::from(PoRepProofPartitions::from(porep_config)),
+    );
 
     let get_verifying_key =
         || ZigZagCompound::verifying_key(&public_params, &ENGINE_PARAMS).map_err(Into::into);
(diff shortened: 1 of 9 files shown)


filecoin-proofs/src/api/internal.rs
```diff
@@ -199,17 +203,50 @@ const SLOTH_ITER: usize = 0;
...
+            taper_layers,
+        };
+    }
+    guess
 }
 
-fn setup_params(sector_bytes: PaddedBytesAmount) -> layered_drgporep::SetupParams {
+fn setup_params(
+    sector_bytes: PaddedBytesAmount,
+    partitions: usize,
```
laser: SUGGESTION: I see that this function calls `select_challenges` with `POREP_MINIMUM_CHALLENGES`. I see that it has a `partitions` parameter which is of type `usize`. I think it makes sense to use `PoRepProofPartitions` as the type for the `partitions` parameter instead.

sector-base/src/api/porep_config.rs
```diff
@@ -9,11 +9,14 @@ pub enum PoRepConfig {
     Test,
 }
 
+// When modifying, update internal::tests::partition_layer_challenges_test to reflect supported PoRepProofPartitions.
 #[derive(Clone, Copy, Debug)]
 pub enum PoRepProofPartitions {
     Two,
 }
 
+pub const POREP_PROOF_PARTITION_CHOICES: [PoRepProofPartitions; 1] = [PoRepProofPartitions::Two];
```
laser: SUGGESTION: I don't think that you should do this. I created an enum so that the compiler could enforce exhaustiveness of pattern matches - and this array defeats that. I think that instead of putting each enum member into a fixed length array and iterating over that array's elements that you should instead pattern match. This will prevent you/me from needing to maintain the `POREP_PROOF_PARTITION_CHOICES` constant.

sector-base/src/api/sector_size.rs
```diff
@@ -10,6 +10,9 @@ pub enum SectorSize {
     TwoHundredFiftySixMiB,
 }
 
+pub const SECTOR_SIZE_CHOICES: [SectorSize; 2] =
```
laser: SUGGESTION: same comment as above, with `POREP_PROOF_PARTITION_CHOICES`

filecoin-proofs/src/bin/paramcache.rs
```diff
@@ -84,10 +89,11 @@ pub fn main() {
     cache_post_params(PoStConfig::Test);
 
     if !test_only {
-        cache_porep_params(PoRepConfig::Live(
-            SectorSize::TwoHundredFiftySixMiB,
-            PoRepProofPartitions::Two,
-        ));
+        for p in &POREP_PROOF_PARTITION_CHOICES {
```
laser: SUGGESTION: See comment at `POREP_PROOF_PARTITION_CHOICES`.

filecoin-proofs/src/api/internal.rs
```diff
@@ -1142,4 +1189,24 @@ mod tests {
...
+                partitions,
+                POREP_MINIMUM_CHALLENGES,
+                LAYERS,
+                TAPER_LAYERS,
+                TAPER,
+            )
+            .all_challenges()
+        };
+        // Update to ensure all supported PoRepProofPartitions options are represented here.
+        assert_eq!(vec![1, 1, 2, 2], f(usize::from(PoRepProofPartitions::Two)));
```
laser: SUGGESTION: See comment at `POREP_PROOF_PARTITION_CHOICES`.

filecoin-proofs/src/api/internal.rs
```diff
@@ -447,7 +485,7 @@ fn make_merkle_tree<T: Into<PathBuf> + AsRef<Path>>(
     let mut data = Vec::new();
     f_in.read_to_end(&mut data)?;
 
-    public_params(bytes).graph.merkle_tree(&data)
+    public_params(bytes, 1).graph.merkle_tree(&data)
```
laser: QUESTION: Why the hard-coded 1 here? `make_merkle_tree` is called during PoSt generation. I would expect that PoSt proof partitions would be plumbed in to `make_merkle_tree` instead of this hard-coded 1.

porcuquine: I intended that as a later step. As we discussed, PoSt doesn't currently support multiple partitions so I defaulted to what it does — until such time as we need and plumb otherwise.

sector-base/src/api/porep_config.rs
porcuquine: Yes, I considered that. It's not clear to me that you can use wildcard pattern matching to enumerate the choices of an enum — which is what we'd like to be able to do. Especially when enumerating combinations, a nested loop will work, but I think we'd need to duplicate whatever manual/matching enumeration strategy was chosen — in each branch. I could be missing something, though.

laser: After thinking about this again, I retract my original objection. If your intent is to generate parameters for each combination of sector size and PoRep proof partitions, then this is a fine approach. Alternatively, you could try to generate an `Iter` impl (see [here](https://cbreeden.github.io/Macros11/)).

filecoin-proofs/src/api/internal.rs
porcuquine: `select_challenges` is purely about calculating a `LayerChallenges` that meets some challenge requirement. It could be logically housed in `filecoin_proofs` or used to reason about hypothetically (but not yet) supported partition amounts. (For example, see the unit test used to gain confidence in the correctness of its output.) If `select_challenges` were to take a`PoRepProofPartitions` we would still want another function exactly like the current one that it could call to do the real work. That seems like yet another layer of indirection I'd rather avoid.

filecoin-proofs/src/api/internal.rs
porcuquine: As the number of choices grows, we could change this test to iterate through `POREP_PROOF_PARTITION_CHOICES` if we feel that's valuable. The value of having these lines here is less to test them as to expose the calculated result explicitly in code for reference. (It does also give an opportunity to sanity check correctness — but that should be primarily handled by the explicit tests below — which are fairly sparse now but could be supplemented if necessary.)

PR #623, opened 2019-04-18
accept and return proof partitions through FFI + eliminate FFI enums + add seal verification to FFI test 

laser:
Supports go-filecoin issue `#2591`.

## Why does this PR exist?

- FFI consumers need to know how to chunk the proofs returned from `generate_post`
- FFI consumers need to tell us how to chunk the proofs provided to `verify_post`
- FFI test was not exercising `verify_seal` FFI function

## What's in this PR?

- Replace most of the `FFI*` enums, as they weren't providing any value (hat tip @porcuquine)
- Remove notion of `Live` versus `Test` `SectorClass`
- Remove notion of `Live` versus `Test` (`PoStConfig` and `PoRepConfig`)
- FFI functions which need number of proof partitions receive number of proof partitions
- FFI functions whose callers need to know number of proof partitions (for chunking) now receive number of proof partitions

### DIFF
--- a/filecoin-proofs/examples/ffi/main.rs
+++ b/filecoin-proofs/examples/ffi/main.rs
@@ -284,8 +284,7 @@ unsafe fn sector_builder_lifecycle(use_live_store: bool) -> Result<(), Box<Error
         (piece_bytes, piece_key)
     };
 
-    // poll for sealed sector metadata through the FFI and then verify that the
-    // proof was valid
+    // poll for sealed sector metadata through the FFI
     {
         let (result_tx, result_rx) = mpsc::channel();
         let (kill_tx, kill_rx) = mpsc::channel();
--- a/filecoin-proofs/examples/ffi/main.rs
+++ b/filecoin-proofs/examples/ffi/main.rs
@@ -13,6 +13,7 @@ extern crate sector_base;
 
 include!(concat!(env!("OUT_DIR"), "/libfilecoin_proofs.rs"));
 
+use byteorder::{LittleEndian, WriteBytesExt};
 use ffi_toolkit::c_str_to_rust_str;
 use ffi_toolkit::free_c_str;
 use ffi_toolkit::rust_str_to_c_str;
@@ -33,6 +34,18 @@ use tempfile::TempDir;
 ///////////////////////////////////////////////////////////////////////////////
 // SectorBuilder lifecycle test
 ///////////////////////////////
+
+fn u64_to_fr_safe(sector_id: u64) -> [u8; 31] {
+    let mut byte_vector = vec![];
+    byte_vector.write_u64::<LittleEndian>(sector_id).unwrap();
+    byte_vector.resize(31, 0);
+
+    let mut byte_array = [0; 31];
+    let bytes = &byte_vector[..byte_array.len()]; // panics if not enough data
+    byte_array.copy_from_slice(bytes);
+
+    byte_array
+}
 
 fn make_piece(num_bytes_in_piece: usize) -> (String, Vec<u8>) {
     let mut rng = thread_rng();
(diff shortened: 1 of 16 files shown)


filecoin-proofs/examples/ffi/main.rs
```diff
@@ -128,9 +141,9 @@ unsafe fn sector_builder_lifecycle(use_live_store: bool) -> Result<(), Box<Error
     let sizes = if use_live_store {
         ConfigurableSizes {
             sector_class: FFISectorClass {
-                sector_size: FFISectorSize_SSB_TwoHundredFiftySixMiB,
-                seal_proof_partitions: FFISealProofPartitions_SPP_Two,
-                post_proof_partitions: FFIPoStProofPartitions_PPP_One,
+                sector_size: 266338304,
```
porcuquine: I don't understand how this number is calculated.

I believe this should correspond to the number of bytes in 256MiB, which is 268435456.

If that's as intended, please write this in a way which makes the derivation clear. For examples `1024 * 1024 * 256` or `1 << 28`.

It almost looks as though this might be the 'raw data size' (i.e. before bit-padding), but below you use 1024 for 1KiB. Without other information, I assume `sector_size` is the sector size, not the amount of original data a sector can hold but rather the power-of-two sector size.

Please clarify.

laser: > I believe this should correspond to the number of bytes in 256MiB, which is 268435456.

Yes, you are correct. The value 266338304 is computed from 268435456 minus bit padding. This is a bug, now fixed.

> If that's as intended, please write this in a way which makes the derivation clear. For examples 1024 * 1024 * 256 or 1 << 28.

Done.

filecoin-proofs/src/api/mod.rs
```diff
@@ -535,106 +524,92 @@ pub unsafe extern "C" fn get_staged_sectors(
...
+    proof_partitions: u8,
+    flattened_proofs_ptr: *const u8,
+    flattened_proofs_len: libc::size_t,
+) -> crate::error::Result<Vec<u8>> {
+    porep_proof_partitions::try_from_u8(proof_partitions)
+        .map(PoRepProofBytesAmount::from)
+        .map(usize::from)
+        .map(|chunk| into_proof_vecs(chunk, flattened_proofs_ptr, flattened_proofs_len))
+        .and_then(|vs| {
+            vs.first()
```
porcuquine: In fact, now that I reach this point, it looks as though we are discarding all but the first of the proof vectors returned by `into_proofs_vec`. *If* `into_proofs_vec` had indeed been called with the correct number of partitions (which I now hypothesize it was *not*) — then we would be discarding partitions. Instead, we're simply discarding non-existent extra partitions. This is because we've pretended `proof_chunks` was the size of all the chunks combined, when it should have just been the (constant) size of one proof partition.

'Fortunately', this all works out 'fine' because `verify_seal` does all the work of splitting a larger proof vector into correct-sized chunks anyway. So none of these machinations bought us anything. In fact, we introduced and canceled out errors — perhaps somewhat an over-reliance on the assumption that at the end of the day what type checks must be correct.

filecoin-proofs/src/api/mod.rs
```diff
@@ -535,106 +524,92 @@ pub unsafe extern "C" fn get_staged_sectors(
...
+unsafe fn try_into_porep_proof_bytes(
+    proof_partitions: u8,
+    flattened_proofs_ptr: *const u8,
+    flattened_proofs_len: libc::size_t,
+) -> crate::error::Result<Vec<u8>> {
+    porep_proof_partitions::try_from_u8(proof_partitions)
+        .map(PoRepProofBytesAmount::from)
+        .map(usize::from)
+        .map(|chunk| into_proof_vecs(chunk, flattened_proofs_ptr, flattened_proofs_len))
+        .and_then(|vs| {
```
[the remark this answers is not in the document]
laser: > If I'm reading this right, we sequentially apply a chain of three function calls in order to find out the number of partitions (which could/should have been calculated by the caller, so doesn't provide new information) — but then only use that number to break the proofs into individual proofs.

You are not reading this right. The caller provides the number of proof partitions N as a u8, and then the dynamic array which the caller provides is chunked using that value. The `porep_proof_partitions::try_from_u8` function enforces an invariant, which is that there's only a few values of N which our system can handle.

The use case is more compelling when accepting a dynamic array of flattened PoSt proofs, but validating / chunking PoRep proofs in the same way seems sane to me.

Anyways, I'm sorry that you don't like all the types. I don't agree that the types have "no clear purpose." Quite the contrary; they allow us to enforce invariants at compile time.

The good news is that I'm going to delete almost all of those types in a coming PR, because of backwards-compatibility issues that we discussed last week.

PR #628, opened 2019-04-22
Add progress bar during parameter download

mslipper:
Closes https://github.com/filecoin-project/go-filecoin/issues/2488.

Progress bars look like this (I included an error case):

Screen Shot 2019-04-22 at 3 07 45 PM


laser:
Thanks @mslipper. I will review this by EOD Los Angeles.

PR #634, opened 2019-04-30
loosen sector size / proof partitions restrictions

laser:
## Why does this PR exist?

If two versions of rust-fil-proofs communicate with each other (i.e. a miner running V1 software is asked to verify a proof created by V2 software), we need to ensure that the older version doesn't immediately reject the proofs its been asked to verify because the associated sector size or partitions value isn't a member of its (now defunct) enum.

## What's in this PR?

- replace enums with [newtypes](https://doc.rust-lang.org/book/ch19-04-advanced-types.html#using-the-newtype-pattern-for-type-safety-and-abstraction)


filecoin-proofs/src/bin/paramcache.rs
```diff
@@ -84,16 +87,23 @@ pub fn main() {
 
     let test_only: bool = matches.is_present("test-only");
 
-    cache_porep_params(PoRepConfig(SectorSize::OneKiB, PoRepProofPartitions::Two));
-    cache_post_params(PoStConfig(SectorSize::OneKiB, PoStProofPartitions::One));
+    cache_porep_params(PoRepConfig(
+        SectorSize(TEST_SECTOR_SIZE),
```
dignifiedquire: Is there a reason that `{TEST|LIVE}_SECTOR_SIZE` are not of type `SectorSize` already?

filecoin-proofs/src/bin/paramcache.rs
```diff
@@ -84,16 +87,23 @@ pub fn main() {
...
         for p in &POREP_PROOF_PARTITION_CHOICES {
-            cache_porep_params(PoRepConfig(SectorSize::TwoHundredFiftySixMiB, *p));
+            cache_porep_params(PoRepConfig(SectorSize(LIVE_SECTOR_SIZE), *p));
         }
+
         cache_post_params(PoStConfig(
-            SectorSize::TwoHundredFiftySixMiB,
-            PoStProofPartitions::One,
+            SectorSize(LIVE_SECTOR_SIZE),
+            PoStProofPartitions(1),
```
dignifiedquire: maybe mirror with `const LIVE_POST_PROOF_PARTITIONS = PoStProofPartitions(1)`?

dignifiedquire:
ups, just realized the WIP, sorry if commenting too early

laser:
@dignifiedquire - 

> ups, just realized the WIP, sorry if commenting too early

No, you're good. I just now removed the WIP flag.

filecoin-proofs/src/api/mod.rs
```diff
@@ -530,25 +517,31 @@ unsafe fn try_into_post_proofs_bytes(
...
-        .map(usize::from)
-        .map(|chunk| into_proof_vecs(chunk, flattened_proofs_ptr, flattened_proofs_len))
+    let chunk_size = proof_partitions as usize * SINGLE_PARTITION_PROOF_LEN;
+
+    if flattened_proofs_len % chunk_size != 0 {
+        Err(format_err!(
+            "proofs array len={:?} incompatible with partitions={:?}",
+            flattened_proofs_len,
+            proof_partitions
+        ))
```
dignifiedquire: tip: I believe you can use [ensure](https://docs.rs/failure/0.1.5/failure/macro.ensure.html) to write this a bit more compact:

```rust
ensure!(a != 0, "error message... {} - {}", a, b);
```

laser: I like! Did not know.

sector-base/src/api/porep_proof_partitions.rs
```diff
@@ -1,31 +1,29 @@
...
+        x.0 as usize
     }
 }
 
 impl From<PoRepProofPartitions> for PoRepProofBytesAmount {
     fn from(x: PoRepProofPartitions) -> Self {
-        PoRepProofBytesAmount(SINGLE_PARTITION_PROOF_LEN * usize::from(x))
+        PoRepProofBytesAmount(usize::from(x) * SINGLE_PARTITION_PROOF_LEN)
     }
 }
```
porcuquine: Is this conversion path even used anymore? (I think it's not.)

PR #642, opened 2019-05-05
Provide an option to optimize either wall clock time of replication or memory

schomatis:
Based on https://github.com/filecoin-project/rust-fil-proofs/pull/567.

Closes https://github.com/filecoin-project/rust-fil-proofs/issues/647.

### DIFF
--- a/storage-proofs/src/settings.rs
+++ b/storage-proofs/src/settings.rs
@@ -16,6 +16,9 @@ pub struct Settings {
     pub merkle_tree_path: String,
     pub num_proving_threads: usize,
     pub replicated_trees_dir: String,
+    pub generate_merkle_trees_in_parallel: bool,
+    // Generating MTs in parallel optimizes for speed while generating them
+    // in sequence (`false`) optimizes for memory.
 }
 
 impl Default for Settings {
@@ -25,6 +28,12 @@ impl Default for Settings {
             merkle_tree_path: "/tmp/merkle-trees".into(),
             num_proving_threads: 1,
             replicated_trees_dir: "".into(),
+            generate_merkle_trees_in_parallel: false,
+            // FIXME: Decide on a sensible default value, setting to `false`
+            //  at the moment for testing purposes only.
+            // FIXME: If the `disk-trees` feature is off the this should be
+            //  true *always*, there's nothing to win in terms of memory
+            //  with sequential generation if we're not offlading MTs to files.
         }
     }
 }
--- a/storage-proofs/src/settings.rs
+++ b/storage-proofs/src/settings.rs
@@ -44,7 +44,9 @@ impl Settings {
 
         #[cfg(feature = "disk-trees")]
         {
-            if settings.is_ok() && settings.as_ref().unwrap().generate_merkle_trees_in_parallel == false {
+            if settings.is_ok()
+                && settings.as_ref().unwrap().generate_merkle_trees_in_parallel == false
+            {
                 return Err(ConfigError::Message("Setting GENERATE_MERKLE_TREES_IN_PARALLEL to false (sequiental generation) \ndoesn't add any value if the `disk-trees` feature is not set (no offload possible)".to_string()));
             }
         }
--- a/storage-proofs/src/settings.rs
+++ b/storage-proofs/src/settings.rs
@@ -40,6 +40,15 @@ impl Settings {
         s.merge(File::with_name(SETTINGS_PATH).required(false))?;
         s.merge(Environment::with_prefix("FIL_PROOFS"))?;
 
-        s.try_into()
+        let settings: Result<Settings, ConfigError> = s.try_into();
+
+        #[cfg(feature = "disk-trees")]
+        {
+            if settings.is_ok() && settings.as_ref().unwrap().generate_merkle_trees_in_parallel == false {
+                return Err(ConfigError::Message("Setting GENERATE_MERKLE_TREES_IN_PARALLEL to false (sequiental generation) \ndoesn't add any value if the `disk-trees` feature is not set (no offload possible)".to_string()));
+            }
+        }
+
+        settings
     }
 }
--- a/storage-proofs/src/settings.rs
+++ b/storage-proofs/src/settings.rs
@@ -28,12 +28,7 @@ impl Default for Settings {
             merkle_tree_path: "/tmp/merkle-trees".into(),
             num_proving_threads: 1,
             replicated_trees_dir: "".into(),
-            generate_merkle_trees_in_parallel: false,
-            // FIXME: Decide on a sensible default value, setting to `false`
-            //  at the moment for testing purposes only.
-            // FIXME: If the `disk-trees` feature is off the this should be
-            //  true *always*, there's nothing to win in terms of memory
-            //  with sequential generation if we're not offlading MTs to files.
+            generate_merkle_trees_in_parallel: true,
         }
     }
 }
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -97,7 +97,7 @@ jobs:
           name: Test (stable) in release profile
           command: |
             cargo +stable test --verbose --release --all --features disk-trees
-            RUSTFLAGS="-D warnings" cargo +stable build --examples --release --all --features disk-trees
+            RUSTFLAGS="-D warnings" cargo +stable build --examples --release --frozen --all --features disk-trees
           environment:
             FIL_PROOFS_REPLICATED_TREES_DIR: replicated-disk-trees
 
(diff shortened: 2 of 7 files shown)


schomatis:
Without the intermediate anonymous `mmap` (for the sequential MT generation check) the max RSS for a sector size of 8 GiB is reduced (as expected) from 40 to 32 GiB, 4 sector sizes, the replication process (`transform_and_replicate_layers` ) thus takes now (the intended) 2 sector sizes (1 MT), the `zigzag` example retains one sector size in memory during the entire run (to run the extraction at the end) and there remains a sector size to be accounted for (that would add up to the measured 32 GiB).

schomatis:
Tagging this for review but #567 should be reviewed and merged first.

schomatis:
Trying to investigate the extra sector size in the replication process (see previous [comment](https://github.com/filecoin-project/rust-fil-proofs/pull/642#issuecomment-491514472)) I'm `strace`ing the `zigzag` example and observing a pattern of anonymous `mmap`s that I'm not identifying as part of the (new) disk-backed MT generation setup. Additionally, this pattern appears only with big sector sizes and starts out with the sector size and then successively reduces the size by a factor of 1/2. I'm _speculating_ then that this is a Rust allocation mechanism and that we're still allocating local space for the MT or (most probably) the replication `data` from where it is generated. I'll try to do an in-depth profile to track its source.

The pattern observed is, from a sector size of 256 MiB:

```
mmap(NULL, 268435456, PROT_READ|PROT_WRITE, MAP_SHARED, 5, 0) = 0x7f17e8000000  # File-backed `mmap` of the leaves (one sector size)
openat(AT_FDCWD, "/home/snarky/space/rust-fil-proofs-trees-dir/tree-1-3366915306-top-half", O_RDWR|O_CREAT|O_CLOEXEC, 0666) = 6
ftruncate(6, 268435456)                 = 0
fstat(6, {st_mode=S_IFREG|0664, st_size=268435456, ...}) = 0
mmap(NULL, 268435456, PROT_READ|PROT_WRITE, MAP_SHARED, 6, 0) = 0x7f17d8000000  # File-backed `mmap` of the top half of the MT (second sector size)

mmap(NULL, 268439552, PROT_READ|PROT_WRITE, MAP_PRIVATE|MAP_ANONYMOUS, -1, 0) = 0x7f17b7fff000  # Anonymous mmap of one sector size (presumed
                                                                                                   to be thread stack storage).
mmap(NULL, 268439552, PROT_READ|PROT_WRITE, MAP_PRIVATE|MAP_ANONYMOUS, -1, 0) = 0x7f1787fff000  # Another anonymous, same size.
munmap(0x7f17b7fff000, 268439552)       = 0                                                     # Unmapping first anonymous.
mmap(NULL, 134221824, PROT_READ|PROT_WRITE, MAP_PRIVATE|MAP_ANONYMOUS, -1, 0) = 0x7f17cffff000  # Mapping _half_ sector size now.
futex(0x55d575765418, FUTEX_WAKE_PRIVATE, 2147483647) = 8
futex(0x55d5757653c0, FUTEX_WAKE_PRIVATE, 1) = 1
futex(0x55d575773cb8, FUTEX_WAIT_PRIVATE, 0, NULL) = 0
futex(0x55d5757748e0, FUTEX_WAKE_PRIVATE, 1) = 0
munmap(0x7f1787fff000, 268439552)       = 0
munmap(0x7f17cffff000, 134221824)       = 0
mmap(NULL, 134221824, PROT_READ|PROT_WRITE, MAP_PRIVATE|MAP_ANONYMOUS, -1, 0) = 0x7f17cffff000
mmap(NULL, 134221824, PROT_READ|PROT_WRITE, MAP_PRIVATE|MAP_ANONYMOUS, -1, 0) = 0x7f17bffff000
munmap(0x7f17cffff000, 134221824)       = 0
mmap(NULL, 67112960, PROT_READ|PROT_WRITE, MAP_PRIVATE|MAP_ANONYMOUS, -1, 0) = 0x7f17d3fff000   # Now 1/4th of a sector size.
futex(0x55d57576541c, FUTEX_WAKE_PRIVATE, 2147483647) = 8
futex(0x55d5757653c0, FUTEX_WAKE_PRIVATE, 1) = 1
futex(0x55d575773cb8, FUTEX_WAIT_PRIVATE, 0, NULL) = 0
futex(0x55d5757748e0, FUTEX_WAKE_PRIVATE, 1) = 0
munmap(0x7f17bffff000, 134221824)       = 0
munmap(0x7f17d3fff000, 67112960)        = 0
mmap(NULL, 67112960, PROT_READ|PROT_WRITE, MAP_PRIVATE|MAP_ANONYMOUS, -1, 0) = 0x7f17d3fff000
mmap(NULL, 67112960, PROT_READ|PROT_WRITE, MAP_PRIVATE|MAP_ANONYMOUS, -1, 0) = 0x7f17cfffe000
munmap(0x7f17d3fff000, 67112960)        = 0
mmap(NULL, 33558528, PROT_READ|PROT_WRITE, MAP_PRIVATE|MAP_ANONYMOUS, -1, 0) = 0x7f17d5fff000  # 1/8th, and so on..
futex(0x55d575765418, FUTEX_WAKE_PRIVATE, 2147483647) = 8
futex(0x55d5757653c0, FUTEX_WAKE_PRIVATE, 1) = 0
```

The command run is:

```
export TREES_DIR=/home/snarky/space/rust-fil-proofs-trees-dir                &&
FIL_PROOFS_REPLICATED_TREES_DIR="$TREES_DIR"                                  \
nohup strace  cargo run                                       \
    disk-trees                                                                \
  --                                                                          \
    --size 262144                                                            \
    --no-bench                                                                \
    --m 1  --expansion 0 --layers 1                                           \
     > zigzag-8gib.out &
```

(attaching the full `zigzag-8gib.out` which confusingly is named with the standard 8 GiB test size, this reduced 256 MiB size allows for faster tests preserving the allocation pattern)

/cc @porcuquine @dignifiedquire in case this pattern makes sense to you.

[zigzag-8gib.txt](https://github.com/filecoin-project/rust-fil-proofs/files/3170702/zigzag-8gib.txt)

schomatis:
Also note that the current implementation in this PR does a _sequential_ MT generation-encoding so I also don't follow where are all the `futex` calls coming from.

porcuquine:
> the `zigzag` example retains one sector size in memory during the entire run (to run the extraction at the end) and there remains a sector size to be accounted for (that would add up to the measured 32 GiB).

That extraction hasn't turned out to be very important and practice, and the memory usage is problematic for our benchmarks. Can we add it as command-line option defaulting to no-extraction/no-memory?

schomatis:
> That extraction hasn't turned out to be very important and practice, and the memory usage is problematic for our benchmarks. Can we add it as command-line option defaulting to no-extraction/no-memory?

@porcuquine Actually my characterization of that memory region (`data`, as it's usually called) was too simplistic; it's not retained just for extraction purposes, it's (I think) heavily embedded in the entire replication logic (some functions even use it as scratch pad). I do agree that taking it out (and absorbing it as an MT `Store`) would be a very nice win but I'd have to take a closer look at it and I'll definitely need external support to carry it through (mentioning it just so you can evaluate the cost/benefit and the estimated time, we can discuss it in more depth in a separate issue).

schomatis:
(Part of the motivation behind splitting between leaves and the rest of the MT was to provide storage options that would succeed the current overuse of `&mut [u8]` such as this one.)

storage-proofs/src/settings.rs
```diff
@@ -23,6 +27,13 @@ impl Default for Settings {
             maximize_caching: false,
             merkle_tree_path: "/tmp/merkle-trees".into(),
             num_proving_threads: 1,
+            replicated_trees_dir: "".into(),
+            generate_merkle_trees_in_parallel: false,
+            // FIXME: Decide on a sensible default value, setting to `false`
+            //  at the moment for testing purposes only.
```
schomatis: Need to decide this.

dignifiedquire: which option are you referring to with `disk-trees`?

schomatis: `generate_merkle_trees_in_parallel`

schomatis: I suppose should be `true` as it's the current behavior, but wanted to make sure.

schomatis: Misunderstood the original question I think, sorry, `disk-trees` is not an internal configuration option but the Rust feature that enables all the file-mapping related logic.

dignifiedquire: right, you could set the default depending on the build flag

dignifiedquire: You could put that logic into the config loading part, if you want

schomatis: in `Settings.new`? should that happen before or after the `try_into`?

schomatis: done

dignifiedquire:
lets get this in, it has been long enough, any further improvements we can do in follow ups :)

PR #656, opened 2019-05-21
feat(filecoin-proofs): expose public safe api

dignifiedquire:


dignifiedquire:
> I strongly believe that the FFI-exposed functions should call this functions, and that these functions should call the `internal.rs` function.

Do we actually need the indirection of `internal.rs` then?

laser:
@dignifiedquire 

> Do we actually need the indirection of internal.rs then?

It was my original intent to turn the `internal.rs` functions into the exported Rust API (which you've now got in `filecoin-proofs/src/api/safe.rs`). So, to answer your question directly: No, I don't think you need the `internal.rs` functions if you have the `safe.rs` functions. You just need to move the guts of `internal.rs` into the `safe.rs` file.

laser:
@dignifiedquire - I can take this PR over if you like, merging `safe.rs` with `internal.rs`. How's that sound?

dignifiedquire:
closing in favor of #684

schomatis:
Doing some house cleaning, going to delete this branch if nobody reclaims it tomorrow (it's showing up in my bunch because I rebased it once and has been my faithful companion ever since).

dignifiedquire:
please do

PR #657, opened 2019-05-21
fix(storage-proofs): use specific version of merkletree

dignifiedquire:


schomatis:
:-1: for me, should have squashed

PR #663, opened 2019-05-24
feat(logging-toolkit): make logging into file possible 

### LINKED ISSUE #624: configure logger to write to a non-stdout file
### Description

rust-fil-proofs is a library, but its only consumer is go-filecoin. As such, Rust logging norms (e.g. libraries shouldn't log to stdout themselves) are hard to follow. The rust-fil-proofs (library) logger can't write to the go-filecoin (daemon) logger, so instead we initialize the logger ourselves and write to stdout. 

**No coordination exists between the `libfilecoin_proofs.a` consumer's logger and the loggers configured in `libfilecoin_proofs.a`, so log messages will be interleaved if they're all sent to stdout.** If JSON logging is enabled, interleaving writes to stdout can result in malformed (e.g.) JSON.

To work around this issue, we should introduce a mechanism (environment variable, probably) which can be used to control the location to which rust-fil-proofs log entries will be written.

### Acceptance criteria

- environment variable controls log output path
- environment variable is documented somewhere
- defaults to stdout (e.g. `/dev/stdout`)

vmx:
It's now possible to write the log output into a file instead of stdout.

Just set the environment variable `FIL_PROOFS_LOG_FILE` to the file you would
like to log into. You can also log to stdout via `/dev/stdout` or stderr via
`/dev/stdout`.

Example:

    cd storage-proofs
    FIL_PROOFS_LOG_FILE=/tmp/test.log cargo test circuit::zigzag -- --nocapture

The default behaviour equals:

    cd storage-proofs
    FIL_PROOFS_LOG_FILE=/dev/stdout cargo test circuit::zigzag -- --nocapture

The `--nocapture` is only there to see the stdout output in case there is any.

Fixes #624.


laser:
Integrating this branch with go-filecoin [here](https://circleci.com/workflow-run/b78a1519-5de8-4e31-bf27-553769529a33).

PR #665, opened 2019-05-24
chore: deny unreadable literal clippy rule

vmx:
In some modules using underscores in number literals makes things indeed
more readable. Hence removing the allowance of unreadable literals.

An exception is the storage-proofs crate where hex values are used, there
it doesn't lead to more readible code.

I like the idea of following as many clippy rules as possible. Feel free to close
this if you don't agree.


laser:
Integrating PR with go-filecoin [here](https://circleci.com/gh/filecoin-project/go-filecoin/tree/laser%2Ftest-integrations-1559104765).

PR #667, opened 2019-05-24
refactor: remove unsafe impl Send/Sync

Stebalien:
Rust unsafe mantra: you are not special!

There are probably other ways to solve this, but this appears to work (unless there's some other crate that this breaks).


laser:
@Stebalien 

> Rust unsafe mantra: you are not special!

Thanks for the PR! 

Would you mind explaining what you mean by "you are not special?" Binging for the term did not result in any elucidating blog posts / docs / etc. 

Perhaps you mean that there's almost always some (superior) alternative to abominations like `unsafe impl Sync for WrappedSectorStore {}`?

Erin

Stebalien:
> Would you mind explaining what you mean by "you are not special?" Binging for the term did not result in any elucidating blog posts / docs / etc.

Sorry, that's rather hard to search for. It comes from https://doc.rust-lang.org/nomicon/print.html#transmutes. Specifically, it refers to transmuting but the same applies to any case where one tries to "fix" the type checker by using unsafe.

I intended it to be light-hearted ribbing but, re-reading it, that _definitely_ doesn't come across. I'm really sorry about that.

> Perhaps you mean that there's almost always some (superior) alternative to abominations like 

I mean this type violates the contract of unsafe. If one doesn't use unsafe, one isn't supposed to be able to invoke undefined behavior. However, this type provides a "safe" escape hatch that allows some types to be sent across threads, even if they're not `Send`. The current uses of this type may not actually _invoke_ undefined behavior, but the fact that this type exists makes it impossible to rely on the unsafe contract when auditing the code.

laser:
@Stebalien 

> I'm really sorry about that.

No offense taken at all :)

> The current uses of this type may not actually invoke undefined behavior, but the fact that this type exists makes it impossible to rely on the unsafe contract when auditing the code.

This makes sense to me. Thanks for the explanation!

laser:
@Stebalien - PR integrates just fine. Would you mind rebasing? I'll merge shortly thereafter.

laser:
Candidate branch has successfully built with go-filecoin. Build output can be viewed [here](https://circleci.com/workflow-run/6c9abaa4-7099-45f2-bfe5-75511eaf0e22).

Issue #390, opened 2018-12-13
add clippy rules preventing `Result#unwrap` and `Option#unwrap`


laser:
## Acceptance Criteria

1. replace all calls to `unwrap` with `expects`
1. add clippy rules which fail builds if `unwrap` is used

## Helpful Links

1. https://rust-lang.github.io/rust-clippy/master/index.html#option_unwrap_used
1. https://rust-lang.github.io/rust-clippy/master/index.html#result_unwrap_used

porcuquine:
I think `expects_with_backtrace` is now `expects`.

schomatis:
Potential 3rd point in the acceptance criteria:

* Add a note in the [Coding Standards](https://github.com/filecoin-project/rust-fil-proofs/blob/master/CONTRIBUTING.md#coding-standards) suggesting an alternative, see https://github.com/filecoin-project/rust-fil-proofs/pull/671#discussion_r288795873.

dignifiedquire:
@DrPeterVanNostrand after your PR is merged, can you please add the warnings, and copy the output into a gist, so we can have a better understanding how much is left on this? Thank you

DrPeterVanNostrand:
@dignifiedquire here are the clippy warnings
https://gist.github.com/DrPeterVanNostrand/1465e93eda049995e4c49d1d972a74fa

laser:
@DrPeterVanNostrand - Are you working on this issue? If not, do you know what's remaining? I'm having some trouble determining the state of this item from reading the GitHub issue-comments.

porcuquine:
800 commits later, I'm not sure if this issue can be closed or not. Please advise.

vmx:
There's even a PR open for it: https://github.com/filecoin-project/rust-fil-proofs/pull/1122

PR #671, opened 2019-05-29
refactor(storage-proofs): remove panic-able unwraps

DrPeterVanNostrand:
Relates to Issue #390 
- Removes `panic`-able `unwrap`s from the `storage-proofs` subcrate.
- Does not add clippy lint for disallowing unwrapping because many `unwrap`s can't `panic`.

### DIFF
--- a/storage-proofs/src/batchpost.rs
+++ b/storage-proofs/src/batchpost.rs
@@ -246,21 +246,23 @@ mod tests {
 
         let priv_inputs = PrivateInputs::<H>::new(data.as_slice(), &tree);
 
-        let proof = BatchPoST::<H>::prove(&pub_params, &pub_inputs, &priv_inputs).unwrap();
+        let proof = BatchPoST::<H>::prove(&pub_params, &pub_inputs, &priv_inputs)
+            .expect("failed to create proof");
 
-        assert!(
-            BatchPoST::<H>::verify(&pub_params, &pub_inputs, &proof).unwrap(),
-            "failed to verify"
-        );
+        let proof_is_valid = BatchPoST::<H>::verify(&pub_params, &pub_inputs, &proof)
+            .expect("failed to verify proof");
+
+        assert!(proof_is_valid, "failed to verify");
 
         // mess with a single part of the proof
         {
             let mut proof = proof;
             proof.challenges[0] = proof.challenges[0] + 1;
-            assert!(
-                !BatchPoST::<H>::verify(&pub_params, &pub_inputs, &proof).unwrap(),
-                "verified invalid proof"
-            );
+
+            let proof_is_valid = BatchPoST::<H>::verify(&pub_params, &pub_inputs, &proof)
+                .expect("failed to verify proof");
+
+            assert!(!proof_is_valid, "verified invalid proof");
         }
     }
 
--- a/storage-proofs/src/circuit/por.rs
+++ b/storage-proofs/src/circuit/por.rs
@@ -301,7 +301,8 @@ mod tests {
                 &public_params.vanilla_params,
                 setup_params.engine_params,
             )
-            .unwrap();
+            .expect("failed to generate groth params");
+
             let proof = PoRCompound::<PedersenHasher>::prove(
                 &public_params,
                 &public_inputs,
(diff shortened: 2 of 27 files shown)


storage-proofs/src/circuit/por.rs
```diff
@@ -557,19 +574,22 @@ mod tests {
...
-            let proof = merklepor::MerklePoR::<PedersenHasher>::prove(
+            let proving_res = merklepor::MerklePoR::<PedersenHasher>::prove(
                 &pub_params,
                 &pub_inputs,
                 &priv_inputs,
-            )
-            .unwrap();
+            );
+            let proof = proving_res.unwrap_or_else(|e| {
+                panic!("Proving faile: {}", e);
```
laser: Looks like the trailing `d` got away from you! :)

laser:
@dignifiedquire - Would you mind reviewing this? Thanks for the PR, @DrPeterVanNostrand.

storage-proofs/src/example_helper.rs
```diff
@@ -337,10 +337,14 @@ pub trait Example<'a, C: Circuit<Bls12>>: Default {
         let (data_size, challenge_count, m, sloth_iter, typ) = {
             let matches = instance.clap();
 
-            let data_size = value_t!(matches, "size", usize).unwrap() * 1024;
-            let challenge_count = value_t!(matches, "challenges", usize).unwrap();
-            let m = value_t!(matches, "m", usize).unwrap();
-            let sloth_iter = value_t!(matches, "sloth", usize).unwrap();
+            let data_size = value_t!(matches, "size", usize)
+                .map(|size| size * 1024)
+                .expect("Failed to parse `size` CLI arg as `usize`");
```
schomatis: Question out of Rust inexperience: is it the same to have the check after the `map`?

storage-proofs/src/batchpost.rs
```diff
@@ -246,21 +246,29 @@ mod tests {
 
         let priv_inputs = PrivateInputs::<H>::new(data.as_slice(), &tree);
 
-        let proof = BatchPoST::<H>::prove(&pub_params, &pub_inputs, &priv_inputs).unwrap();
+        let proof =
+            BatchPoST::<H>::prove(&pub_params, &pub_inputs, &priv_inputs).unwrap_or_else(|e| {
+                panic!("Failed to create proof: {}", e);
+            });
```
dignifiedquire: The idea was to use [`.expect`](https://doc.rust-lang.org/std/result/enum.Result.html#method.expect) for these. 

that lets you write this as 

```rust
do_fail_thing().expect("do fail thing failed");
```

schomatis: Isn't the information in the error worth it?

dignifiedquire: using `expect` will give you that information, take a look at this playground example: https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=8e81bc6b0ae9dee0102e196dc3e83967

storage-proofs/src/example_helper.rs
dignifiedquire: It is, as `map` only gets executed when no error occurs

storage-proofs/src/batchpost.rs
DrPeterVanNostrand: `expect` prints the `Debug` info, whereas `panic!("{}", e)` prints the `Display`, I used the later because `error::Error` has the nice `Fail` display attributes on its variants (e.g. https://github.com/filecoin-project/rust-fil-proofs/blob/master/storage-proofs/src/error.rs#L8) and imo gives better context.

I'm happy to use either `expect` or `panic`.

schomatis: Good point, I think both options serve the purpose of #390 quite nicely so I'd favor keeping the PR as is and avoid duplicating work.

DrPeterVanNostrand: @dignifiedquire are you in favor of using `expect` everywhere?

dignifiedquire: I think expect is much more idiomatic rust, and much easier to read, so lets use that. If we need more detailed views on panic, we should try to improve our Error type.

storage-proofs/src/example_helper.rs
```diff
@@ -31,7 +31,7 @@ pub fn prettyb(num: usize) -> String {
     );
     let pretty_bytes = format!("{:.2}", num / delimiter.powi(exponent))
         .parse::<f64>()
-        .unwrap()
+        .expect("Failed to parse `numer` as `u64`")
```
dignifiedquire: I think there is a b missing

dignifiedquire:
ups just read your not around clippy, I think we could write the explanation into the „expect“ instead, or a generic „unfaillable“ to use expects everywhere, not a 100% sure though on the best strategy

DrPeterVanNostrand:
@dignifiedquire I could add the following to `storage-proofs/src/lib.rs`:
```
#![warn(clippy::option_unwrap_used)]
#![warn(clippy::result_unwrap_used)]
```
but that produces a lot of clippy warnings

dignifiedquire:
Lets figure out the clippy settings later, and get these improvements in :)

PR #673, opened 2019-05-31
WIP: Catch panics

### LINKED ISSUE #553: a panic should never unwind the stack past the FFI
### Description

go-filecoin depends on rust-fil-proofs via CGO. A panic in rust-fil-proofs should not propagate to go-filecoin where it cannot be handled.

### Acceptance criteria

Trap all panics at the FFI boundary. Transmute panic to an error message, and send that error message back to FFI consumer.

### Where to start

[This](https://doc.rust-lang.org/nomicon/ffi.html) portion of the Rust docs outline an approach for catching panics before propagating back to FFI consumers. [This](https://blog.datalust.co/rust-at-datalust-how-we-integrate-rust-with-csharp/) blog post gives some examples of how we might use the `UnwindSafe` trait.

vmx:
This code currently doesn't even compile. But I upload it though, in case
someone else would like to push that over the finish line.

Closes #553.

### DIFF
--- a/filecoin-proofs/src/api/mod.rs
+++ b/filecoin-proofs/src/api/mod.rs
@@ -1,6 +1,7 @@
 use libc;
 use slog::*;
 use std::mem;
+use std::panic;
 use std::ptr;
 use std::slice::from_raw_parts;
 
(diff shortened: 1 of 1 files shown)


vmx:
Please not that I won't be able to work on it next week. So feel free to take this as a starting point if it's an important thing to get fixed. If not, then I'll continue that on in two weeks.

dignifiedquire:
@vmx thank you for pushing this forward. I am wondering if we need to wrap each method, or if we could have a simple setup call do the work and reuse the structure that https://github.com/rust-cli/human-panic uses

vmx:
@dignifiedquire Oh nice. I didn't know about https://doc.rust-lang.org/std/panic/fn.set_hook.html. When I looked into this issue I only looked into https://github.com/jrobsonchase/easy_ffi and https://github.com/Michael-F-Bryan/ffi_helpers.

I'll try again in two weeks with `set_hook` as it really sounds like what we actually need.

laser:
@dignifiedquire (cc @vmx) Any objections to my taking this work over?

Note: When the sector builder is dropped (this happens when the node shuts down), we [`join`](https://github.com/filecoin-project/rust-fil-proofs/blob/master/filecoin-proofs/src/api/sector_builder/mod.rs#L207) the child (worker) threads. If a [child thread panicked](https://doc.rust-lang.org/1.2.0/std/thread/struct.JoinGuard.html#method.join), it will be re-raised in the parent at that point.

dignifiedquire:
@laser lets wait until repo splitting + cli investigations are through?

laser:
roger that

vmx:
What's the current state? Should I get back to this or pick another issue on Friday?

laser:
> What's the current state? Should I get back to this or pick another issue on Friday?

@vmx - I have been pulled over to a different thing. Please do pick this up when you are able.

vmx:
I think using `std::panic::set_hook()` won't really work in our case. If I understand it correctly, what we really like is that a call into the FFI does never panic, but return an error response. Here's the catch. with `std::panic::set_hook()` we lose all context and also can't return anything. You could print some error, but you can't return a proper error message.

Does this mean that I should pursue the approach I started here (next week)?

laser:
> If I understand it correctly, what we really like is that a call into the FFI does never panic, but return an error response.

@vmx You do indeed understand it correctly!

> You could print some error, but you can't return a proper error message.

I agree with your analysis. We do want to return to the FFI consumer an error message and non-zero status code. With those constraints in mind, `std::panic::set_hook` doesn't look like a good fit.

vmx:
I'm closing this one, it's superseded by https://github.com/filecoin-project/rust-fil-sector-builder/pull/9.

PR #684, opened 2019-06-04
separate sector builder/non-sector builder FFI code + provide idiomatic Rust API to "primitive proofs operations"

laser:
Sorry about the size of this PR. If you want to make an omelette, you gotta break some eggs.

## Why does this PR exist?

We're about to split rust-fil-proofs into a few different repositories. rust-fil-proofs must provide an idiomatic Rust interface to the primitive proofs operations. FFI code needs to be split into sector builder and non-sector builder-specific portions so that we can move each chunk into its own repo.

## What's in this PR?

- create new `safe.rs` file, which provides an idiomatic Rust interface to the primitive proofs operations
  - seal
  - verify seal
  - verify post
  - generate post
- create new `ffi_sector_builder.rs` file, which contains FFI code specific to interacting with the sector builder
  - add piece by key
  - read piece by key
  - seal everything
  - get sealed sectors
  - get staged sectors
  - generate a PoSt
- create new `ffi_stateless.rs` file, which contains sector builder-nonspecific code
  - verify a PoSt
  - verify a seal
  - get max user bytes which fit into a (post-padding) sector


laser:
@dignifiedquire 

> why not split the safe api into statefull and stateless?

It occurred to me that there's no such thing as a standalone, stateful, safe API. That's just the `SectorBuilder` API, right? e.g. `SectorBuilder#add_piece`, `SectorBuilder#read_piece_from_sealed_sector` etc.

dignifiedquire:
that is true

laser:
@dignifiedquire 

> why keep the api indirection?

I just removed it. Good call.

Issue #682, opened 2019-06-04
Extract sector management and ffi related code into a new repo


dignifiedquire:
in this repo:
- [crate] storage-proofs
- [crate] filecoin-proofs
  - will contain previous `sector-base/io/fr32.rs`
- [crate] logging-toolkit

new repo: rust-fil-sector-builder
- [crate] sector-builder
  - Previous sector base
  - sector-builder 
- [crate] drop-struct-..
- [crate] sector-builder-ffi
  - produces libsectorbuilder
  - previous ffi api
  - Ffi-toolkit

dignifiedquire:
New project is here: https://github.com/filecoin-project/rust-fil-sector-builder

PR #687, opened 2019-06-05
refactor: split code into new crates, to prepare for repo splitting

dignifiedquire:
This does the splitting of code as described in #682, only without moving the code out of this repo. That will happen in a later step

Ref #682


laser:
@dignifiedquire Pushed update to packaging script. Will squash-merge once built.

PR #689, opened 2019-06-07
fix(storage-proofs): invert compilation option in generate_merkle_tre…

schomatis:
…es_in_parallel


schomatis:
> lets just warn instead of an error

I can't see a scenario where that combination of configurations make sense and my fear is that the user may miss it and pay a considerable performance penalty for it. What about, in that case, resetting `GENERATE_MERKLE_TREES_IN_PARALLEL` to `true` and warn the user about that?

dignifiedquire:
I am thinking about benchmarks that need to run on single core only

schomatis:
Good point, fixing.

PR #706, opened 2019-06-17
fix(storage-proofs): avoid mut warning in merkletree bench

schomatis:


storage-proofs/benches/merkle.rs
```diff
@@ -9,7 +9,10 @@ use storage_proofs::hasher::pedersen::PedersenHasher;
 use storage_proofs::zigzag_graph::{ZigZag, ZigZagBucketGraph, DEFAULT_EXPANSION_DEGREE};
 
 fn merkle_benchmark(c: &mut Criterion) {
+    #[cfg(feature = "big-sector-sizes-bench")]
     let mut params = vec![128, 1024];
+    #[cfg(not(feature = "big-sector-sizes-bench"))]
+    let params = vec![128, 1024];
```
schomatis: I really hope there's a cleaner way to do this _without_ repeating the parameter values.

dignifiedquire: there is, but I think for now this works well enough

schomatis: out of curiosity (I'm merging this) what other alternatives were you thinking of?

dignifiedquire: probably sth like this

```rust
let mut params = vec![];
params.push(1);
params.push(2);

#[cfg(not(feature = "bla"))]
params.push(3);
```

schomatis: nice

Issue #677, opened 2019-06-04
ensure direct parent is first parent input to KDF


porcuquine:
### Description

When parents are passed to KDF, they are sorted by index. The sort order should be reversed for every other layer.

Recall that the 'direct parent' i.e. the one with index directly preceding the current node for the direction of this layer's replication -- is always include as a parent of each node. This 'direct parent' should be the one hashed first. This means that when replicating in the 'forward' direction (as in the first layer, with index 0), the parents should be sorted in reverse index order.

Otherwise, it's possible to pre-hash all the parents except the direct parent, minimizing the amount of required sequential hashing.

### Acceptance criteria

KDF hashing cannot begin until the previous node has been encoded.

### Where to begin

Most simply:
 - Modify the `parents` method in `drgraph.rs` to sort in descending order. We need to change the value returned here so `DrgPoRep` behaves correctly.
 - Modify the `parents` method in `zigzag_graph.rs` to reverse the order of the calculated parents if the graph is reversed.

Ideally, instead of reversing, add a parameter to the base parents method, allowing callers to control the sort order.

dignifiedquire:
@schomatis this should be unblocked, let me know if you need anything to finish this

schomatis:
unblocked, will resolve this week

PR #712, opened 2019-06-25
fix(storage-proofs): ensure direct parent is first parent input to KDF

schomatis:
*Not working.*

Fixes https://github.com/filecoin-project/rust-fil-proofs/issues/677.

I don't think `parents` (and related) functions should be concerned with the order of its nodes, that seems like the responsibility of the functions that use them for KDF purposes (e.g., `encode` and similar).

### DIFF
--- a/storage-proofs/src/drgporep.rs
+++ b/storage-proofs/src/drgporep.rs
@@ -325,6 +325,12 @@ where
                 //     challenge,
                 // )?;
 
+                // The direct parent should be the one hashed first, see `vde::encode`.
+                parents.sort();
+                if pub_params.graph.forward() {
+                    parents.reverse();
+                }
+
                 let extracted = decode_domain_block::<H>(
                     pub_params.sloth_iter,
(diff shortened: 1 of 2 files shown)


porcuquine:
There needs to be a function available to both circuit and non-circuit proofs, which specifies the canonical ordering of the parents. This is necessary so that public inputs for proof generation/verification will agree. The canonical ordering is currently provided by the `parents` method, and I don't think there's a good reason to change it. Anything else is likely to add a lot of gratuitous complexity. The actual fix required here is small; let's keep it simple.

For reference, here's where `parents` is called when generating public inputs for the DrgPoRep circuit: https://github.com/filecoin-project/rust-fil-proofs/blob/master/storage-proofs/src/circuit/drgporep.rs#L194

dignifiedquire:
@schomatis what is the state of this?

schomatis:
Interrupted by work with more priority, will resume this week.

Issue #629, opened 2019-04-23
paramfetch should allow operators to download only the parameters which they'll use


laser:
### Description

Developers today use 256MiB and 1024KiB sectors when sealing. The devnet/user cluster uses 256MiB sectors when sealing. Soon, we will publish Groth parameters for 1GiB sectors. 

Developers will likely not use this sector size during their normal development flow. They should not have to download Groth parameters for 1GiB sectors if they do not plan to seal 1GiB sectors.

Mining rig operators will use only one sector size. They should not have to download Groth parameters for sector sizes they'll never use when sealing.

### Acceptance criteria

- prevent go-filecoin operators and developers from needing to download all published Groth parameters
- ensure that all verifying keys are downloaded regardless of which Groth parameters are downloaded

### Risks + pitfalls

- will likely need to modify `parampublish` such that `parameters.json` includes sector size

--------

# Proposal: Supporting 1GiB Sectors

For additional background, see [this](https://github.com/filecoin-project/rust-fil-proofs/issues/629) and [this](https://github.com/filecoin-project/rust-fil-proofs/issues/625).

## Problem

Developers use 256MiB and 1KiB sectors today. Cryptolab envisions publishing 1GiB sectors sometime in the near future, and I think that we should start running large sectors in the devnet/user cluster within the next month or two so that we can see where our system fails.

We don't want to require go-filecoin developers or mining operators to download Groth parameters for sectors which they never intend to seal. Our existing `paramfetch` program allows an operator to either download all parameters or download parameters on a per filename basis. The file names don't communicate the _meaning_ of the Groth parameter-file (e.g. the filename `v10-zigzag-proof-of-replication-6648cd509799c8b76f944bbc971692c69d1480fa60c569bfc1284f0bd8ead988` doesn't communicate the size of the sector), so an operator doesn't have the information required to opt out of downloading Groth parameters for sectors which they'll never seal.

### Additional Metadata in `parameters.json`

The existing `parameters.json` file should be modified to include additional parameter and verifying key-related metadata. `parampublish` will need to be modified in order to achieve this. We will augment each JSON object in `parameters.json` to include the number of bytes in the associated (sealed) sector:

```json
{
  "v10-vdf-post-c8ddc8e74f9c2ed6a4bc83df171a9e9e84e242cf6e48e20f71585106fa8a3ce6.vk": {
    "sector_size": 266338304,
[...]

PR #714, opened 2019-06-30
generate Groth parameter cache metadata file

laser:
Makes incremental progress towards #629 

## Why does this PR exist?

We need to give miners a way to download only the Groth parameters that plan to use when sealing. The `parameters.json` file, which we distribute with the go-filecoin binary, includes (today) only the name of each published file (Groth parameter or VK), its checksum, and IPFS cid. No metadata is published, so `paramfetch` cannot preset a miner with a choice of downloading (or not) assets based on e.g. sector size.

## What's in this PR?

This PR exposes sector size as a property of a parameter set. VDF PoSt and ZigZag PoRep (the two circuits for which we currently publish parameters and verifying keys) both produce non-`None` values. This sector size is written to a struct which in turn is JSON encoded and written to disk along with the verifying key and/or Groth parameter. In the future, `parampublish` will read these metadata files and insert their contents into the `parameters.json` file.

## Demo

Before this changeset:

```shell
06:51 $ ls /tmp/filecoin-proof-parameters/
v10-vdf-post-c8ddc8e74f9c2ed6a4bc83df171a9e9e84e242cf6e48e20f71585106fa8a3ce6
v10-vdf-post-c8ddc8e74f9c2ed6a4bc83df171a9e9e84e242cf6e48e20f71585106fa8a3ce6.vk
v10-zigzag-proof-of-replication-6648cd509799c8b76f944bbc971692c69d1480fa60c569bfc1284f0bd8ead988
v10-zigzag-proof-of-replication-6648cd509799c8b76f944bbc971692c69d1480fa60c569bfc1284f0bd8ead988.vk
```

After:

```shell
06:52 $ ls /tmp/filecoin-proof-parameters/
v10-vdf-post-c8ddc8e74f9c2ed6a4bc83df171a9e9e84e242cf6e48e20f71585106fa8a3ce6
v10-vdf-post-c8ddc8e74f9c2ed6a4bc83df171a9e9e84e242cf6e48e20f71585106fa8a3ce6.meta
v10-vdf-post-c8ddc8e74f9c2ed6a4bc83df171a9e9e84e242cf6e48e20f71585106fa8a3ce6.vk
v10-zigzag-proof-of-replication-6648cd509799c8b76f944bbc971692c69d1480fa60c569bfc1284f0bd8ead988
v10-zigzag-proof-of-replication-6648cd509799c8b76f944bbc971692c69d1480fa60c569bfc1284f0bd8ead988.meta
v10-zigzag-proof-of-replication-6648cd509799c8b76f944bbc971692c69d1480fa60c569bfc1284f0bd8ead988.vk
```

Contents of a `.meta` file:

```shell
06:52 $ cat v10-zigzag-proof-of-replication-
v10-vdf-post-c8ddc8e74f9c2ed6a4bc83df171a9e9e84e242cf6e48e20f71585106fa8a3ce6.meta
{"sector_size":268435456}
```

### DIFF
--- a/storage-proofs/src/drgporep.rs
+++ b/storage-proofs/src/drgporep.rs
@@ -94,6 +94,7 @@ where
     }
 
     fn sector_size(&self) -> Option<u64> {
+        // TODO: Expose sector size, if available.
         None
     }
 }
--- a/storage-proofs/src/drgporep.rs
+++ b/storage-proofs/src/drgporep.rs
@@ -10,7 +10,7 @@ use crate::error::Result;
 use crate::fr32::bytes_into_fr_repr_safe;
 use crate::hasher::{Domain, Hasher};
 use crate::merkle::{MerkleProof, MerkleTree};
-use crate::parameter_cache::ParameterSetIdentifier;
+use crate::parameter_cache::ParameterSetMetadata;
 use crate::porep::{self, PoRep};
 use crate::proof::{NoRequirements, ProofScheme};
 use crate::vde::{self, decode_block, decode_domain_block};
@@ -54,7 +54,7 @@ pub struct DrgParams {
 pub struct PublicParams<H, G>
 where
     H: Hasher,
-    G: Graph<H> + ParameterSetIdentifier,
+    G: Graph<H> + ParameterSetMetadata,
 {
     pub graph: G,
     pub sloth_iter: usize,
@@ -67,7 +67,7 @@ where
 impl<H, G> PublicParams<H, G>
 where
     H: Hasher,
-    G: Graph<H> + ParameterSetIdentifier,
+    G: Graph<H> + ParameterSetMetadata,
 {
     pub fn new(graph: G, sloth_iter: usize, private: bool, challenges_count: usize) -> Self {
         PublicParams {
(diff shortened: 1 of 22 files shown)


storage-proofs/src/drgporep.rs
```diff
@@ -80,18 +80,23 @@ where
...
         format!(
             "drgporep::PublicParams{{graph: {}; sloth_iter: {}}}",
-            self.graph.parameter_set_identifier(),
+            self.graph.identifier(),
             self.sloth_iter
         )
     }
+
+    fn sector_size(&self) -> Option<u64> {
+        // TODO: Expose sector size, if available.
```
dignifiedquire: in which case is this needed?

laser: @dignifiedquire 

> in which case is this needed?

Sector size is required if we wish to publish Groth parameters for the circuit. If we don't, we can leave the return value `None`. I am not sure which PoSt construction we're going to ultimately publish, so I went through the effort of returning a non-`None` value for only VDF PoSt. 

> what about all the TODOs?

VDF PoSt is the only PoSt constructor which is used by the go-filecoin node today. Same thing goes for ZigZag PoRep. If you'd like me to go through the effort of figuring out how to expose sector size for circuits for which we aren't going to publish parameters, I will - but I don't think it's a great use of time.

laser:
@dignifiedquire - I'm going to close this PR. I'll create a new PR when:

1. I have `parampublish` creating the `parameters.json` file which includes sector size
1. I have fixed the remaining TODOs

PR #737, opened 2019-07-12
feat(doc): add optimization strategies in README

### LINKED ISSUE #703: Document usage for the memory reduction use case with the `disk-trees` feature

schomatis:
Now that I read it altogether this is not at all a straightforward optimization process, @porcuquine we're still missing a general profile configuration to optimize either for memory or speed that basically says "I don't care about the details, I have enough X resources, just make sure Y is as small as possible".

Closes #703.

### DIFF
--- a/README.md
+++ b/README.md
@@ -158,6 +158,54 @@ docker build -t foo -f ./Dockerfile-ci . && \
     cargo run --release --package filecoin-proofs --example ffi --target x86_64-unknown-linux-gnu
 ```
 
+## Optimizing for either speed or memory during replication
+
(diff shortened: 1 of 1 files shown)


porcuquine:
Agreed. I think `MAXIMIZE_CACHING` was originally intended to be such a single-bit statement of intention. We just used it as the single handle on parent caching when that was (and still is, I think) the only available option.

The idea would be to have a small number of such master settings which together define a 'strategy' and can easily be set to optimize completely for memory, completely for speed, and perhaps for some reasonable intermediate compromise.

Users wanting more fine-grained control can tune individual settings manually. As yet there are relatively few options, but as these grow we should try to converge on something like the design described above (or some alternative to be proposed).

Issue #566, opened 2019-03-18
verify_post or generate_post should produce a validation error if provided a CommR which cannot be transformed into Fr


laser:
### Description

If `verify_post` or `generate_post` are passed 32 bytes which cannot be converted into a Fr, we currently panic:

```
let commitments = fixed
    .comm_rs
    .iter()
    .map(|comm_r| {
        PedersenDomain(
            bytes_into_fr::<Bls12>(comm_r)
                .expects("could not could not map comm_r to Fr")
                .into_repr(),
        )
    })
    .collect::<Vec<PedersenDomain>>();
```

We should instead return an `Error` and let the caller (`mod.rs`) transmute this to an FFI error-response.

### Acceptance criteria

1. Caller does not see panic when passing CommR which cannot be transformed into an Fr
1. rust-fil-proofs return value includes some information about the failure, e.g. a string containing "could not could not map comm_r to Fr"

porcuquine:
The one point I would make is that we don't want to lose the inner error information. (In this case, `BadFrBytesError`.)

If/when we implement a general solution to unwind and catch panics at the API border, then using `expects` in this way will be a reasonable way to capture both the inner error and the outer context.

Otherwise, we have to get into error wrapping, which seems needlessly complicated.

Since it's simplest for now, let's go with returning and error here — but I think we should avoid a policy of flattening errors this way, since it will make debugging actual internal error conditions much harder.

We should instead implement a general solution to internal panics and use that (in my opinion).

laser:
> We should instead implement a general solution to internal panics and use that (in my opinion).

I think that we should implement a general solution to internal panics.

I think we should also work hard to not panic. I think that if the FFI consumer provides us with a bad CommR (as in: cannot be transformed into an Fr), and we can detect that it is a bad CommR without panicking that we should not panic.

porcuquine:
That's fine in this case — since there can only really be one error. In the general case, we should not throw away internal errors. If we feel it important to add context from the API layer, we should either wrap the errors (which seems like overkill) or consider using a panicking wrapper (like `expects`) to do accomplish the goal. The point is that if we have a defined semantics around panicking such that it is handled as we intend, then we don't necessarily need to avoid it at all costs.

As it stands, in other places where internal errors can occur, we're just returning them as-is. I'm fine with that.

laser:
> we should not throw away internal errors.

I agree that we should not throw away internal errors.

> we should either wrap the errors

[...]

PR #755, opened 2019-07-23
pass BadFrBytes error up instead of panicking

sidke:
# What's in this pr?
- swap out functional comm_r generation for imperative to easily catch on errors and pass their information up
- incomplete test for `verify_post_fixed_sectors_count` to check the unhappy paths. would like to pull @laser in to have a more robust test for this function at this layer

resolves: https://github.com/filecoin-project/rust-fil-proofs/issues/566

### DIFF
--- a/filecoin-proofs/src/api.rs
+++ b/filecoin-proofs/src/api.rs
@@ -8,13 +8,13 @@ use ff::PrimeField;
 use memmap::MmapOptions;
 use paired::bls12_381::Bls12;
 use paired::Engine;
-use tempfile::tempfile;
 
 use storage_proofs::circuit::multi_proof::MultiProof;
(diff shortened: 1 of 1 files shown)


filecoin-proofs/src/api.rs
```diff
@@ -759,6 +774,61 @@ mod tests {
         Ok(())
     }
 
+    #[test]
+    fn test_verify_seal_fr32_validation() {
+        //        verify_seal() // comm_r, comm_d, comm_r_star
```
dignifiedquire: why is this commented outd?

laser: @sidke It should be deleted, actually.

laser: ...or rewritten to actually test `verify_seal` instead of `verify_post`. :)

PR #762, opened 2019-07-27
feat(change POST_SECTORS_COUNT from 2 to 262144 as per spec)

laser:
## Why does this change exist?

The existing implementation sets POST_SECTORS_COUNT to 2, and runs a single PoSt per pair of replica commitments passed to the `generate_post` call. This is fine when len(replicas) is a small number, but causes timeouts when it's a large number. To conform to the spec, we should change 2 to be 262144.

## What's in this PR?

- delete the adapter layer
- generate a (bogus) PoSt proof in a single shot over all replicas
- if not len(replicas) < POST_SECTORS_COUNT, append to the replicas vector a clone of the head of the vector until len(replicas) == POST_SECTORS_COUNT
- breaking API change: generate_post now produces one proof (and uses new type-names)
- breaking API change: verify_post now accepts one proof (and uses new type-names)


dignifiedquire:
@laser looks like some ci jobs are unhappy

laser:
Closing this PR and rethinking things.

PR #763, opened 2019-07-29
feat(storage-proofs): implement rational post

dignifiedquire:
Follows the definition in https://github.com/filecoin-project/specs/pull/332

PRs for other repos

- https://github.com/filecoin-project/rust-fil-proofs-ffi/pull/12
- https://github.com/filecoin-project/rust-fil-sector-builder/pull/23


storage-proofs/src/rational_post.rs
```diff
@@ -0,0 +1,461 @@
...
+            "missmatching challenges and commitments"
+        );
+        let challenges = pub_inputs.challenges;
+
+        let proofs = challenges
+            .iter()
+            .zip(pub_inputs.commitments.iter())
+            .map(|(challenge, commitment)| {
+                let challenged_sector = *challenge % sector_count;
+                let challenged_leaf = *challenge % (sector_size / 32);
```
porcuquine: See https://github.com/filecoin-project/specs/pull/332/files#r308409694.

Given the complexity noted in the issue, I'd like to see derivation of challenged sectors and challenged leaves abstracted and shared between proof and verification — in such a way that we can carefully examine the method in one place to feel confident its safe.

storage-proofs/src/rational_post.rs
```diff
@@ -0,0 +1,461 @@
...
+        pub_inputs: &'b Self::PublicInputs,
+        priv_inputs: &'b Self::PrivateInputs,
+    ) -> Result<Self::Proof> {
+        let sector_size = pub_params.sector_size;
+        let sector_count = priv_inputs.trees.len() as u64;
+
+        assert_eq!(
+            pub_inputs.challenges.len(),
+            pub_inputs.commitments.len(),
+            "missmatching challenges and commitments"
```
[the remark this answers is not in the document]
dignifiedquire: 😱

filecoin-proofs/src/types/sector_class.rs
```diff
@@ -10,7 +10,7 @@ pub struct SectorClass(
 impl From<SectorClass> for PoStConfig {
     fn from(x: SectorClass) -> Self {
         match x {
-            SectorClass(ss, _, ppp) => PoStConfig(ss, ppp),
+            SectorClass(ss, _, _) => PoStConfig(ss),
```
laser: BLOCKING: If `PoStConfig` no longer takes a proof partitions-value, then you should drop it from `SectorClass` too.

storage-proofs/src/circuit/multi_proof.rs
```diff
@@ -42,4 +42,10 @@ impl<'a, E: Engine> MultiProof<'a, E> {
         }
         Ok(())
     }
+
+    pub fn to_vec(&self) -> Vec<u8> {
+        let mut out = Vec::new();
+        self.write(&mut out).expect("known allocation target");
```
laser: QUESTIONS:

1. Are you sure you want to panic here?
2. Instead of adding this method, could you instead satisfy the `From` trait?

storage-proofs/src/circuit/rational_post.rs
```diff
@@ -0,0 +1,377 @@
...
+        _partition_k: Option<usize>,
+    ) -> Vec<Fr> {
+        let mut inputs = Vec::new();
+
+        let por_pub_params = merklepor::PublicParams {
+            leaves: (pub_params.sector_size as usize / NODE_SIZE),
+            private: false,
+        };
+
+        for (challenge, commitment) in pub_in.challenges.iter().zip(pub_in.commitments) {
```
laser: SUGGESTION: I think it would be wise to assert that `pub_in.challenges` and `pub_in.commitments` have equal lengths.

storage-proofs/src/circuit/rational_post.rs
```diff
@@ -0,0 +1,377 @@
...
+        assert_eq!(cs.num_inputs(), 5, "wrong number of inputs");
+        assert_eq!(cs.num_constraints(), 13746, "wrong number of constraints");
+        assert_eq!(cs.get_input(0, "ONE"), Fr::one());
+    }
+
+    #[ignore] // Slow test – run only when compiled for release.
+    #[test]
+    fn rational_post_test_compound() {
+        let params = &JubjubBls12::new();
+        let rng = &mut XorShiftRng::from_seed([0x3dbe6259, 0x8d313d76, 0x3237db17, 0xe5bc0654]);
```
laser: QUESTION: What are these magic values? Where do they come from?

storage-proofs/src/circuit/test/mod.rs
```diff
@@ -314,6 +321,7 @@ impl<E: Engine> TestConstraintSystem<E> {
         assert_eq!(expected.len() + 1, self.inputs.len());
         for (a, b) in self.inputs.iter().skip(1).zip(expected.iter()) {
             if &a.0 != b {
+                println!("failed {:?} - {:?}", a, b);
```
laser: Errant `println!`

storage-proofs/src/rational_post.rs
```diff
@@ -0,0 +1,473 @@
...
+            challenges_count: sp.challenges_count,
+        })
+    }
+
+    fn prove<'b>(
+        _pub_params: &'b Self::PublicParams,
+        pub_inputs: &'b Self::PublicInputs,
+        priv_inputs: &'b Self::PrivateInputs,
+    ) -> Result<Self::Proof> {
+        assert_eq!(
```
laser: nice

storage-proofs/src/rational_post.rs
```diff
@@ -0,0 +1,473 @@
...
+        test_rational_post_validates::<Blake2sHasher>();
+    }
+
+    #[test]
+    fn rational_post_actually_validates_pedersen() {
+        test_rational_post_validates::<PedersenHasher>();
+    }
+
+    fn test_rational_post_validates_challenge_identity<H: Hasher>() {
+        let rng = &mut XorShiftRng::from_seed([0x3dbe6259, 0x8d313d76, 0x3237db17, 0xe5bc0654]);
```
laser: More magic values. Where do these things come from? What do they represent?

filecoin-proofs/src/api.rs
```diff
@@ -67,19 +65,93 @@ pub struct SealOutput {
     pub piece_inclusion_proofs: Vec<PieceInclusionProof<PedersenHasher>>,
 }
 
+#[derive(Clone, Debug)]
+pub struct GeneratePoStOutput {
+    pub proof: Vec<u8>,
+}
+
 /// Generates a proof-of-spacetime, returning and detected storage faults.
```
laser: This comment no longer tells the truth.

filecoin-proofs/src/api.rs
```diff
@@ -67,19 +65,93 @@ pub struct SealOutput {
...
-///
 pub fn generate_post(
     post_config: PoStConfig,
-    challenge_seed: ChallengeSeed,
+    challenge_seed: &ChallengeSeed,
     input_parts: Vec<(Option<String>, Commitment)>,
-) -> error::Result<GeneratePoStDynamicSectorsCountOutput> {
-    generate_post_dynamic(GeneratePoStDynamicSectorsCountInput {
-        post_config,
+    faults: &[u64],
```
laser: QUESTION: What does one of these `u64`s represent? An index in `input_parts`? Or a sector id?

storage-proofs/src/rational_post.rs
```diff
@@ -0,0 +1,473 @@
...
+            }
+        }
+
+        Ok(true)
+    }
+}
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct Challenge {
+    pub sector: u64,
```
laser: BLOCKING: Is this a sector id? An index of some sort? You either need a better name for this field, a better type than `u64`, or some comments. Or maybe all three.

storage-proofs/src/rational_post.rs
```diff
@@ -0,0 +1,473 @@
...
+
+/// Rational PoSt specific challenge derivation.
+pub fn derive_challenges(
+    challenge_count: usize,
+    sector_size: u64,
+    sector_count: u64,
+    seed: &[u8],
+    faults: &[u64],
+) -> Vec<Challenge> {
+    // TODO: ensure sorting of faults
```
laser: SUGGESTION: Since your code requires sorted stuff, I _highly_ suggest that you introduce a new type to replace `&[u64]` which represents a set of sorted unsigned, 64-bit integers. Something like:

```
pub struct SortedSectorIds<'a> {
    ids: &'a [u64] // not pub
}
```

also, create

```
impl From<&[u64]> for SortedSectorIds {
    fn from(ns: &[u64]) -> Self {
        // sort ns
    }
}

impl From<SortedSectorIds> for &[u64] {
    fn from(x: SortedSectorIds) -> Self {
        x.ids
    }
}
```

storage-proofs/src/rational_post.rs
```diff
@@ -0,0 +1,473 @@
...
+    pub leaf: u64,
+}
+
+/// Rational PoSt specific challenge derivation.
+pub fn derive_challenges(
+    challenge_count: usize,
+    sector_size: u64,
+    sector_count: u64,
+    seed: &[u8],
+    faults: &[u64],
```
laser: QUESTION: What does one of these `u64` represent? A sector id?

filecoin-proofs/src/api.rs
```diff
@@ -67,19 +65,93 @@ pub struct SealOutput {
     pub piece_inclusion_proofs: Vec<PieceInclusionProof<PedersenHasher>>,
 }
 
+#[derive(Clone, Debug)]
+pub struct GeneratePoStOutput {
```
laser: BLOCKING: The `verify_post` function returns a `Result<bool>` (no wrapping type). I think that the `generate_post` function should return a `Result<Vec<u8>>` instead of this struct type.

The `generate_post` function used to return a struct because I wanted to return something more descriptive than a tuple (of faults and proof-vec).

filecoin-proofs/src/api.rs
```diff
@@ -67,19 +65,93 @@ pub struct SealOutput {
...
+        challenges: &challenges,
+        commitments: &commitments,
+        faults,
+    };
+
+    let trees: Vec<Tree> = input_parts
+        .iter()
+        .map(|(access, _)| {
+            if let Some(s) = &access {
+                make_merkle_tree(
```
laser: BLOCKING: Are you sure you want to hold all of these trees in memory?

dignifiedquire:
> BLOCKING: This PR needs to generate and publish new Groth parameters.

yes, will do once this PR is fully settled and approved

filecoin-proofs/src/bin/paramcache.rs
```diff
@@ -54,32 +53,44 @@ fn cache_post_params(post_config: PoStConfig) {
...
-                VDFPoStCircuit<Bls12>,
+                RationalPoSt<PedersenHasher>,
+                RationalPoStCircuit<Bls12, PedersenHasher>,
             >>::blank_circuit(&post_public_params, &ENGINE_PARAMS);
-        let _ = VDFPostCompound::get_groth_params(post_circuit, &post_public_params);
+        let _ = <RationalPoStCompound<PedersenHasher>>::get_groth_params(
+            post_circuit,
+            &post_public_params,
+        )
+        .expect("failed to get groth params");
```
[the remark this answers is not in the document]
dignifiedquire: good question

filecoin-proofs/src/types/sector_class.rs
dignifiedquire: isn't sector class used for PoRep too, and that still uses proof partitions

filecoin-proofs/src/api.rs
```diff
@@ -67,19 +65,93 @@ pub struct SealOutput {
...
+                    s,
+                    PaddedBytesAmount(pub_params.vanilla_params.sector_size as u64),
+                )
+                .unwrap()
+            } else {
+                panic!("faults are not yet supported")
```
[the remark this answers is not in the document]
dignifiedquire: I implemented a thing, please review

storage-proofs/src/circuit/multi_proof.rs
dignifiedquire: > Why do you defile the holy `MultiProof` so?

@porcuquine Is this a real questions?

dignifiedquire: > Are you sure you want to panic here?

This can't panic, this is how write works, the write trait only panics if the underlying storage is full, which does not happen for a `Vec`. 

> Instead of adding this method, could you instead satisfy the From trait?

The way the std lib and most libraries do conversions to `Vec` is using a `to_vec` method, `Into` and `From` are for consuming the thing, where as `to_vec` only takes a reference to `self`.

storage-proofs/src/circuit/rational_post.rs
dignifiedquire: we copy them from one test to the other, it's just some seed so we have reproducibility between tests,

storage-proofs/src/rational_post.rs
dignifiedquire: see above

storage-proofs/src/rational_post.rs
dignifiedquire: yes

filecoin-proofs/src/api.rs
dignifiedquire: For now yes, we will improve on this with disk backed merkle trees as you know.

filecoin-proofs/src/api.rs
```diff
@@ -26,35 +44,15 @@ use storage_proofs::piece_inclusion_proof::{
...
-    PaddedBytesAmount, PoRepConfig, PoRepProofPartitions, PoStConfig, PoStProofPartitions,
-    SectorSize, UnpaddedByteIndex, UnpaddedBytesAmount,
-};
-
 /// FrSafe is an array of the largest whole number of bytes guaranteed not to overflow the field.
 pub type FrSafe = [u8; 31];
 
 pub type Commitment = Fr32Ary;
-pub type ChallengeSeed = Fr32Ary;
+pub type ChallengeSeed = [u8; 32];
```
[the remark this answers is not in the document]
dignifiedquire: because the challenge seed as per spec is 32 bytes and no relationship with `Fr`s anymore

filecoin-proofs/src/bin/paramcache.rs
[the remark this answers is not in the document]
dignifiedquire: I remember. If there is an issue in any of these we need to abort, just printing an error makes no sense, given that this failing to execute any of these will result in a total failure of the program.

filecoin-proofs/src/api.rs
```diff
@@ -88,17 +160,66 @@ pub fn generate_post(
...
+        sector_count,
         challenge_seed,
-        proofs,
         faults,
-    })
+    );
+
+    let commitments: Vec<_> = challenges
+        .iter()
+        .map(|c| commitments_all[c.sector as usize])
```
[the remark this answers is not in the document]
dignifiedquire: fixed

filecoin-proofs/src/api.rs
```diff
@@ -88,17 +160,66 @@ pub fn generate_post(
...
+    };
+
+    let verifying_key = get_post_verifying_key(post_config)?;
+
+    let proof = MultiProof::new_from_reader(None, &proof[..], &verifying_key)?;
+
+    let is_valid =
+        RationalPoStCompound::verify(&public_params, &public_inputs, &proof, &NoRequirements)?;
+
+    // Since callers may rely on previous mocked success, just pretend verification succeeded, for now.
```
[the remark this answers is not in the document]
dignifiedquire: fixed

filecoin-proofs/src/api.rs
```diff
@@ -88,17 +160,66 @@ pub fn generate_post(
...
+
+    let commitments: Vec<_> = challenges
+        .iter()
+        .map(|c| commitments_all[c.sector as usize])
+        .collect();
+
+    let public_params: compound_proof::PublicParams<
+        _,
+        rational_post::RationalPoSt<PedersenHasher>,
+    > = RationalPoStCompound::setup(&setup_params).expect("setup failed");
```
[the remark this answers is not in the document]
dignifiedquire: fixed

filecoin-proofs/src/api.rs
```diff
@@ -67,19 +65,93 @@ pub struct SealOutput {
...
+
 /// Generates a proof-of-spacetime, returning and detected storage faults.
 /// Accepts as input a challenge seed, configuration struct, and a vector of
 /// sealed sector file-path plus CommR tuples.
-///
 pub fn generate_post(
     post_config: PoStConfig,
-    challenge_seed: ChallengeSeed,
+    challenge_seed: &ChallengeSeed,
     input_parts: Vec<(Option<String>, Commitment)>,
```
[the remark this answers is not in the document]
dignifiedquire: fixed

filecoin-proofs/src/api.rs
```diff
@@ -67,19 +65,93 @@ pub struct SealOutput {
...
+    let borrowed_trees: Vec<&Tree> = trees.iter().map(|t| t).collect();
+
+    let priv_inputs = rational_post::PrivateInputs::<PedersenHasher> {
+        trees: &borrowed_trees[..],
+    };
+
+    let groth_params = get_post_params(post_config)?;
+
+    let proof = RationalPoStCompound::prove(&pub_params, &pub_inputs, &priv_inputs, &groth_params)
+        .expect("failed while proving");
```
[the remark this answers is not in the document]
dignifiedquire: fixed

storage-proofs/src/rational_post.rs
dignifiedquire: I think I would like to introduce a general `pub type SectorId = u64` into the code base, and as part of that introduce this kind of lists. But I would like to defer this to another PR.

storage-proofs/src/rational_post.rs
dignifiedquire: It is a sector id, and it has the same name as in the spec

filecoin-proofs/src/api.rs
```diff
@@ -67,19 +65,93 @@ pub struct SealOutput {
...
+        sector_size,
+        sector_count,
         challenge_seed,
-        input_parts,
+        &faults,
+    );
+
+    let commitments: Vec<_> = challenges
+        .iter()
+        .map(|c| commitments_all[c.sector as usize])
```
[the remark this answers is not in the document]
laser: We don't have any panic-recovery tooling yet, so this will unwind the stack and crash the Filecoin node. Until we build those tools, please avoid using panics.

filecoin-proofs/src/types/sector_class.rs
laser: Here is the `SectorClass` type:

```
pub struct SectorClass(
    pub SectorSize,
    pub PoRepProofPartitions,
    pub PoStProofPartitions,
);
```

If `PostConfig` no longer takes a proof partitions-value, then you should drop `PoStProofPartitions` from `SectorClass`.

storage-proofs/src/circuit/rational_post.rs
laser: > we copy them from one test to the other

Will you factor these magic values into a constant that can be shared? Looking through the codebase, I found 49 instances of `[0x3dbe6259, 0x8d313d76, 0x3237db17, 0xe5bc0654]`.

storage-proofs/src/rational_post.rs
laser: See comment above. Requesting that you factor all of the `[0x3dbe6259, 0x8d313d76, 0x3237db17, 0xe5bc0654]` into a constant.

filecoin-proofs/src/api.rs
laser: @dignifiedquire - Still waiting on an answer to this question :)

storage-proofs/src/rational_post.rs
laser: > It is a sector id

OK. Would you change the struct field from `sector` to `sector_id`?

storage-proofs/src/circuit/rational_post.rs
dignifiedquire: not in this pr

filecoin-proofs/src/api.rs
dignifiedquire: sector id

storage-proofs/src/rational_post.rs
dignifiedquire: no, I would rather stay in sync with the spec

storage-proofs/src/rational_post.rs
laser: > pub type SectorId = u64

Type aliases don't provide value beyond documentation. If you are going to introduce a `SectorId` type (or change the one we have), please [create a newtype](https://doc.rust-lang.org/1.0.0/style/features/types/newtype.html), as per the docs. 

Motivation: In the above case, any `u64` can be used where a `SectorId` is required (as they're the same type).

storage-proofs/src/rational_post.rs
dignifiedquire: not in this pr, this would be a very independent change

dignifiedquire: not in this pr, this would be a very independent change

storage-proofs/src/rational_post.rs
laser: > But I would like to defer this to another PR.

If you're going to do that, would you please create an issue in rust-fil-proofs and link to that issue in a TODO-comment?

filecoin-proofs/src/api.rs
```diff
@@ -67,38 +67,230 @@ pub struct SealOutput {
     pub piece_inclusion_proofs: Vec<PieceInclusionProof<PedersenHasher>>,
 }
 
-/// Generates a proof-of-spacetime, returning and detected storage faults.
+/// The minimal information required about a replica, in order to be able to generate
+/// a PoSt over it.
+#[derive(Clone, Debug, PartialEq, Eq, Hash)]
+pub struct ReplicaInfo {
```
laser: Nice!

filecoin-proofs/src/api.rs
```diff
@@ -67,38 +67,230 @@ pub struct SealOutput {
...
+            err
+        )
+    })
+}
+
+/// Generates a proof-of-spacetime.
 /// Accepts as input a challenge seed, configuration struct, and a vector of
-/// sealed sector file-path plus CommR tuples.
-///
+/// sealed sector file-path plus CommR tuples, as well as a list of faults.
```
laser: > as well as a list of faults.

This comment tells lies!

dignifiedquire: It does not, this function takes a list of faults as inputs. `faults: &[u64]`

storage-proofs/src/circuit/rational_post.rs
dignifiedquire: https://github.com/filecoin-project/rust-fil-proofs/issues/838

filecoin-proofs/src/api.rs
```diff
@@ -67,38 +67,230 @@ pub struct SealOutput {
...
-        input_parts,
-    })
+        &faults,
+    );
+
+    // Match the replicas to the challenges, as these are the only ones required.
+    let challenged_replicas: Vec<_> = challenges
+        .iter()
+        .map(|c| {
+            if let Some(replica) = replicas.get(c.sector as usize) {
```
laser: BLOCKING: `replicas.get(c.sector as usize)` is going to index into `replicas` by sector id. `replicas` is a `Vec` (i.e. not sector id-indexed data structure).

storage-proofs/src/rational_post.rs
dignifiedquire: https://github.com/filecoin-project/rust-fil-proofs/issues/839

filecoin-proofs/src/api.rs
dignifiedquire: How do I get the sector id from the the replica then?

laser: You'll need to extend `ReplicaInfo` to include this information, e.g.

```
#[derive(Clone, Debug, PartialEq, Eq, Hash)]
pub struct ReplicaInfo {
    /// Path to the replica.
    access: String,
    /// The replica commitment.
    commitment: Commitment,

    sector_id: SectorId
}
```

filecoin-proofs/src/api.rs
laser: I tell lies! The comment tells truths.

storage-proofs/src/rational_post.rs
laser: Thank you sir

storage-proofs/src/rational_post.rs
laser: > no, I would rather stay in sync with the spec

Are you referring [to the field names used in the Challenge Derivation](https://github.com/filecoin-project/specs/blob/master/proof-of-spacetime.md#challenge-derivation) _pseudocode_?

storage-proofs/src/circuit/multi_proof.rs
porcuquine: @dignifiedquire It was a real question, which I found the answer to myself eventually. The form of the question was just to amuse you as you fight through getting this finished.

storage-proofs/src/rational_post.rs
porcuquine: I'm a bit confused by this. If it is indeed a sector id, we should probably use that term in this part of the spec — since it appears elsewhere in the spec. (But this is a spec question, and spec/implementation don't *have* to agree.)

I wonder, though, if it is actually a sector id at all? Doesn't challenge derivation produce an index into an array of sectors? As far as I know, the set of sectors being challenged won't usually have sequential sector ids, starting from 0 — but I do think challenge derivation assumes that the `sector` value produced will index into such a set of values. Since you've decided not to use `sector_id` anyway, this is mostly just a clarification question to make sure we're all on the same page about what is happening.

It might also mean this line of the spec needs changing:
>         attemptedSectors := {SectorID:bool}

laser: > As far as I know, the set of sectors being challenged won't usually have sequential sector ids, starting from 0 — but I do think challenge derivation assumes that the sector value produced will index into such a set of values.

Reviewing the `GeneratePoSt` pseudocode:

```
func GeneratePoSt(
    sectorSize BytesAmount, 
    sectors SectorSet, 
    seed Seed, 
    faults FaultSet
) PoStProof {
```

An important thing to note here is that `SectorSet` is a sector id-indexed data structure (effectively a map from sector id to bool). It is not something you can index in to by _offset_.

In the guts of `GeneratePoSt`, after generating the inclusion proof, we add sector ids (from `sectors`) into `sectorsSorted`. We can index into a `SectorSet` by sector id only - which implies that `challenge.Sector` is a sector id:

```
sectorsSorted[i] = sectors[challenge.Sector]
```

Reviewing the challenge _generation_ pseudocode, it becomes clear that the `Sector` field's value is intended to be an _offset_ in the `SectorSet`:

```
    return Challenge {
        Sector: sector_challenge % sectorCount,
        Leaf: leaf_challenge % (sectorSize / NODE_SIZE),
    }
```

We either need to change, in the spec, the definition of `SectorSet` such that it becomes a sorted list of sector ids (into which we can index by offset) OR we need to rethink challenge generation.

filecoin-proofs/src/api.rs
laser: We should pause this thread until we're all on the same page regarding whether or not `GeneratePoSt` should accept a `SectorSet` (a sector id-indexed map) or an ordered list of sector ids.

storage-proofs/src/rational_post.rs
porcuquine: I don't think challenge generation itself should be modified. There's no sane world in which it produces sector ids. I don't mind which option for fussing with the plumbing we use, though.

storage-proofs/src/rational_post.rs
```diff
@@ -0,0 +1,497 @@
...
+
+        Ok(true)
+    }
+}
+
+/// A challenge specifying a sector and leaf.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct Challenge {
+    // The index of the sector into the sector list.
+    pub sector: SectorId,
```
porcuquine: It it the index, or is it the Id?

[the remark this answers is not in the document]
dignifiedquire: ID, forgot to update the comment

storage-proofs/src/rational_post.rs
```diff
@@ -0,0 +1,497 @@
...
+            pub_inputs.commitments.len(),
+            "mismatched challenges and commitments"
+        );
+        let challenges = pub_inputs.challenges;
+
+        let proofs = challenges
+            .iter()
+            .zip(pub_inputs.commitments.iter())
+            .map(|(challenge, commitment)| {
+                let challenged_sector = u64::from(challenge.sector);
```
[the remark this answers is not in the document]
dignifiedquire: fixed

storage-proofs/src/rational_post.rs
```diff
@@ -0,0 +1,497 @@
...
+    pub challenges: &'a [Challenge],
+    pub faults: &'a SectorSet,
+    /// The root hashes of the underlying merkle trees.
+    pub commitments: &'a [T],
+}
+
+#[derive(Debug, Clone)]
+pub struct PrivateInputs<'a, H: 'a + Hasher> {
+    pub trees: &'a [&'a MerkleTree<H::Domain, H::Function>],
+}
```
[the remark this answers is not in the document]
dignifiedquire: fixed

storage-proofs/src/rational_post.rs
```diff
@@ -0,0 +1,497 @@
...
+        let challenges = pub_inputs.challenges;
+
+        let proofs = challenges
+            .iter()
+            .zip(pub_inputs.commitments.iter())
+            .map(|(challenge, commitment)| {
+                let challenged_sector = u64::from(challenge.sector);
+                let challenged_leaf = challenge.leaf;
+
+                let tree = priv_inputs.trees[challenged_sector as usize];
```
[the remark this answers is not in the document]
dignifiedquire: fixed

filecoin-proofs/src/api.rs
```diff
@@ -129,8 +135,9 @@ fn as_safe_commitment(
 pub fn generate_post(
     post_config: PoStConfig,
     challenge_seed: &ChallengeSeed,
-    replicas: Vec<ReplicaInfo>,
-    faults: &[SectorId],
+    replicas: &[ReplicaInfo],
+    sectors: &SectorSet,
+    faults: &SectorSet,
```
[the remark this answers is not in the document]
dignifiedquire: fixed

filecoin-proofs/src/api.rs
```diff
@@ -163,7 +170,7 @@ pub fn generate_post(
     let challenged_replicas: Vec<_> = challenges
         .iter()
         .map(|c| {
-            if let Some(replica) = replicas.get(u64::from(c.sector) as usize) {
+            if let Some(replica) = replicas.iter().find(|r| r.sector_id == c.sector) {
```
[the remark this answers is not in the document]
dignifiedquire: fixed

storage-proofs/src/rational_post.rs
```diff
@@ -0,0 +1,497 @@
...
+        let graph2 = BucketGraph::<H>::new(32, 5, 0, new_seed());
+        let tree1 = graph1.merkle_tree(data1.as_slice()).unwrap();
+        let tree2 = graph2.merkle_tree(data2.as_slice()).unwrap();
+
+        let seed = (0..32).map(|_| rng.gen()).collect::<Vec<u8>>();
+        let mut faults = SectorSet::new();
+        faults.insert(1.into());
+        let mut sectors = SectorSet::new();
+        sectors.insert(0.into());
+        sectors.insert(1.into());
```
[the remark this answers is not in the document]
dignifiedquire: fixed

filecoin-proofs/src/api/post.rs
```diff
@@ -0,0 +1,307 @@
...
+    let unique_trees_res: Vec<_> = unique_challenged_replicas
+        .into_par_iter()
+        .map(|(id, replica)| replica.merkle_tree(sector_size).map(|tree| (id, tree)))
+        .collect();
+
+    // resolve results
+    let unique_trees: BTreeMap<SectorId, Tree> =
+        unique_trees_res.into_iter().collect::<Result<_, _>>()?;
+
+    let borrowed_trees: BTreeMap<SectorId, &Tree> = challenged_replicas
```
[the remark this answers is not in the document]
dignifiedquire: but I want to unwrap :( this is provably not possible, just the compiler doesn't know about it

filecoin-proofs/src/api/post.rs
```diff
@@ -0,0 +1,313 @@
...
+    unique_challenged_replicas.dedup();
+
+    let unique_trees_res: Vec<_> = unique_challenged_replicas
+        .into_par_iter()
+        .map(|(id, replica)| replica.merkle_tree(sector_size).map(|tree| (id, tree)))
+        .collect();
+
+    // resolve results
+    let unique_trees: BTreeMap<SectorId, Tree> =
+        unique_trees_res.into_iter().collect::<Result<_, _>>()?;
```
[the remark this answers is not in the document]
dignifiedquire: not possible, as `BTreeMap` does not support collection into from a parallel iterator :(

storage-proofs/src/rational_post.rs
```diff
@@ -0,0 +1,514 @@
...
+    challenge_count: usize,
+    sector_size: u64,
+    sectors: &SectorSet,
+    seed: &[u8],
+    faults: &SectorSet,
+) -> Vec<Challenge> {
+    (0..challenge_count)
+        .map(|n| {
+            let mut attempt = 0;
+            loop {
```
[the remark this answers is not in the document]
dignifiedquire: pushed a fix, thank you for the test case

storage-proofs/src/sector.rs
```diff
@@ -1,7 +1,13 @@
-use byteorder::ByteOrder;
+use std::collections::BTreeSet;
 use std::fmt;
 
-#[derive(Debug, Clone, Copy, PartialOrd, Ord, PartialEq, Eq)]
+use byteorder::ByteOrder;
+
+/// An ordered set of `SectorId`s.
+pub type SectorSet = BTreeSet<SectorId>;
```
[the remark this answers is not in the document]
dignifiedquire: changed to `OrderedSectorSet`

PR #765, opened 2019-07-30
feat(move zigzag/benchy over to rust-proofs)

laser:
## Why does this PR exist?

We'd like to couple the ZigZag benchmarker (benchy) to a specific Git SHA of the ZigZag implementation. Moving it out of rust-filbase allows us to easily do that.

## What's in this PR?

This PR moves benchy from rust-filbase into rust-proofs.

### DIFF
--- a/fil-proofs-tooling/Cargo.toml
+++ b/fil-proofs-tooling/Cargo.toml
@@ -23,3 +23,12 @@ prettytable-rs = "0.8"
 regex = "1.1.6"
 prometheus = { version = "0.6.1", features = ["push"] }
 commandspec = "0.12.2"
+
+bellperson = "0.2.0"
+chrono = "0.4.6"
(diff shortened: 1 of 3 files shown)


fil-proofs-tooling/src/bin/benchy.rs
```diff
@@ -0,0 +1,636 @@
...
+        .get_matches();
+
+    match matches.subcommand() {
+        ("zigzag", Some(m)) => zigzag_cmd(m),
+        _ => panic!("carnation"),
+    }
+    .expect("command failed")
+}
+
+lazy_static! {
```
dignifiedquire: please put the zigzag details into its own file, so I can add others easily

laser: done!

PR #777, opened 2019-08-02
Hybrid Merkle Trees and Hybrid Domains

DrPeterVanNostrand:
### Description

Hybrid Merkle Trees are Merkle Trees that use two hashers, one hasher (the "beta" hasher) is used for the first n layers and a second hasher (the "alpha" hasher) is used for the remaining layers until the tree's root is reached.

In order to store two different `Hasher` domains in a single `merkle_light::merkle::Store`, we implement a wrapper enum called `HybridDomain`.

### What's in This PR

- Implements `HybridDomain` in [`storage_proofs/src/hasher/hybrid.rs`](https://github.com/filecoin-project/rust-fil-proofs/blob/hybrid-merkle-hybrid-domain/storage-proofs/src/hasher/hybrid.rs).
- Implements Hybrid Merkle Trees in [`storage_proofs/src/hybrid_merkle.rs`](https://github.com/filecoin-project/rust-fil-proofs/blob/hybrid-merkle-hybrid-domain/storage-proofs/src/hybrid_merkle.rs)
- Adds module for Hybrid Merkle backed PoR [`storage_proofs/src/hybrid_merklepor.rs`](https://github.com/filecoin-project/rust-fil-proofs/blob/hybrid-merkle-hybrid-domain/storage-proofs/src/hybrid_merklepor.rs)
- Changes PIPs to use Hyrbid Merkle Trees 
- Refactors storage-proofs and filecoin-proofs to use Hybrid Merkle Trees and Hybrid Domains
- Adds benchmarking for Hybrid Merkle Trees [`storage-proofs/benches/hybrid_merkle.rs`](https://github.com/filecoin-project/rust-fil-proofs/blob/hybrid-merkle-hybrid-domain/storage-proofs/benches/hybrid_merkle.rs)

### DIFF
--- a/filecoin-proofs/src/api.rs
+++ b/filecoin-proofs/src/api.rs
@@ -13,9 +13,10 @@ use storage_proofs::circuit::multi_proof::MultiProof;
 use storage_proofs::circuit::vdf_post::VDFPostCompound;
 use storage_proofs::circuit::zigzag::ZigZagCompound;
 use storage_proofs::compound_proof::{self, CompoundProof};
-use storage_proofs::drgraph::{DefaultTreeHasher, Graph};
+use storage_proofs::drgraph::{DefaultAlphaHasher, DefaultBetaHasher, Graph};
 use storage_proofs::error::Error;
 use storage_proofs::fr32::{bytes_into_fr, fr_into_bytes, Fr32Ary};
+use storage_proofs::hasher::hybrid::HybridDomain;
 use storage_proofs::hasher::pedersen::{PedersenDomain, PedersenHasher};
 use storage_proofs::hasher::{Domain, Hasher};
 use storage_proofs::layered_drgporep::{self, ChallengeRequirements};
@@ -41,7 +42,7 @@ use crate::constants::{
 use crate::error;
 use crate::file_cleanup::FileCleanup;
 use crate::fr32::{write_padded, write_unpadded};
-use crate::parameters::{post_setup_params, public_params, setup_params};
+use crate::parameters::{post_setup_params, public_params, setup_params, BETA_HEIGHTS};
 use crate::pieces::{get_aligned_source, get_piece_alignment, PieceAlignment};
 use crate::post_adapter::*;
 use crate::singletons::ENGINE_PARAMS;
@@ -64,7 +65,7 @@ pub struct SealOutput {
     pub comm_d: Commitment,
     pub proof: Vec<u8>,
     pub comm_ps: Vec<Commitment>,
-    pub piece_inclusion_proofs: Vec<PieceInclusionProof<PedersenHasher>>,
+    pub piece_inclusion_proofs: Vec<PieceInclusionProof<DefaultAlphaHasher, DefaultBetaHasher>>,
 }
 
 /// Generates a proof-of-spacetime, returning and detected storage faults.
@@ -160,7 +161,10 @@ fn generate_piece_specs_from_source(
         source.read_exact(&mut buf)?;
 
         let mut source = Cursor::new(&buf);
-        let comm_p = generate_piece_commitment_bytes_from_source::<PedersenHasher>(&mut source)?;
+        let comm_p = generate_piece_commitment_bytes_from_source::<
+            DefaultAlphaHasher,
+            DefaultBetaHasher,
+        >(&mut source)?;
 
         piece_specs.push(PieceSpec {
             comm_p,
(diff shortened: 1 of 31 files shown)


filecoin-proofs/src/api.rs
```diff
@@ -388,7 +388,7 @@ pub fn verify_seal(
     > = ZigZagCompound::setup(&compound_setup_params)?;
 
     // Convert comm_d, comm_r, and comm_r_star from `Fr` into the correct `HybridDomain` variants.
-    // Note that comm_r_last is always the same variant as the last encoding layer's comm_r.
+    // Note that comm_r_start is always the same variant as the last encoding layer's comm_r.
```
porcuquine: Is this supposed to be `comm_r_star`?

DrPeterVanNostrand: It's a typo, `comm_r_star` is correct

DrPeterVanNostrand: fixed

DrPeterVanNostrand:
@porcuquine Yea, I can revert those changes in PiP and filecoin-proofs

DrPeterVanNostrand:
@porcuquine Do you mean that the data layer should use a `beta_height == tree height` (i.e all beta hashers) or `beta_height == 0` (all alpha hashers)?

porcuquine:
In practice, it will eventually be all beta hashers (e.g. blake2s, not pedersen) for the data layer. For now it mostly matters that it not be a hybrid tree (to keep PIP simple), and that once we are using hybrid trees it can use the beta hasher. If it's simplest to just establish it as all beta now, that seems fine — even good. We just need to make sure that if this is the case we don't introduce inconsistency in how `CommP` is generated for piece inclusion proofs.

storage-proofs/src/hasher/hybrid.rs
```diff
@@ -0,0 +1,256 @@
...
+    }
+}
+
+impl<AD, BD> Element for HybridDomain<AD, BD>
+where
+    AD: Domain,
+    BD: Domain,
+{
+    fn byte_len() -> usize {
+        32
```
schomatis: Is the 32 a constant independent of the particular alpha/beta domains?

DrPeterVanNostrand: Both `Blake2sDomain` and `PedersenDomain` are 32 bytes and are the only types that implement `Domain`, so we make the assumption that a `HybridDomain` will always be 32 bytes.

schomatis:
> Both `Blake2sDomain` and `PedersenDomain` are 32 bytes and are the only types that implement `Domain`, so we make the assumption that a `HybridDomain` will always be 32 bytes.

This seems like a strong assumption to me, if we indeed feel comfortable with it (which would be great) I would argue in favor of dropping the `Hasher` type parameter from the `HybridMerkleTree` (and similarly `Element` from `MerkleTree`, basically converging both implementations).

If the MT knows and relies on a fixed node size (32 in this case) we can move towards a full-byte API, which is in fact what we feed and take out from the MT in the current implementation. What changes, depending on the tree level we're at, is the `Algorithm`/`Hasher` used (to which the corresponding `Domain` can be bound to), and that (seems to me) can be dynamically dispatched because it's resolved _once_ during the entire MT build or proof generation.

What I'm describing is probably an oversimplification of `Domain`'s responsibilities, I'm ignoring the `Fr` conversion included in the trait, but if those requirements can be offloaded to the `Algorithm` I think we could greatly simplify the MT design (both the hybrid and the original one).

Need feedback from others to check if this makes sense. /cc @porcuquine @dignifiedquire

porcuquine:
> > Both `Blake2sDomain` and `PedersenDomain` are 32 bytes and are the only types that implement `Domain`, so we make the assumption that a `HybridDomain` will always be 32 bytes.
> 
> This seems like a strong assumption to me, if we indeed feel comfortable with it (which would be great) I would argue in favor of dropping the `Hasher` type parameter from the `HybridMerkleTree` (and similarly `Element` from `MerkleTree`, basically converging both implementations).

Whether or not we might want to formalize the assumption that all `Hasher` domains are 32 bytes, I don't think we should do it as part of this PR — which is already maximally extended.

I agree that this is too strong an assumption to make here. However, it's perfectly reasonable to assume (and enforce) that a `HybridMerkleTree` can only ever be created when its `Hasher`s have compatible (same size) domains.

Instead of hard-coding a return value of 32, the most accurate thing would be to call the `byte_len` of both alpha and beta `Hasher` and assert that they are the same before returning that value. If we want to avoid that overhead, we could just arbitrarily return, say, the alpha value and enforce the sameness when instantiating a `HybridMerkleTree`. In practice, we won't ever try to create 'mixed' trees (different byte-len for alpha and beta hashers), but if we are going to depend on the assumption that they *cannot* differ, then we need to check it somewhere.

storage-proofs/src/hasher/hybrid.rs
dignifiedquire: I don't think this is a good assumption. We should probably put a check in places that both the domains have the same length, and then delegate to that.

dignifiedquire: otherwise we should hardcode this much more strictly than just in this one place

schomatis: I realized (by chance) that we're already doing this:

https://github.com/filecoin-project/rust-fil-proofs/blob/b40d12ab3aa6ef38bc7b92e558837136a89bb221/storage-proofs/src/util.rs#L7

This assumption is hard-coded and implied in the entire MT build process, if this is indeed the case then I'm going to leverage it as another argument against the `Element`/`Domain` abstraction tied to the MT.

schomatis:
@dignifiedquire @DrPeterVanNostrand What's the strategy to follow this? It seems like a big PR to leave it open for too long (conflicts will keep piling up).

dignifiedquire:
Branch is now protected to ensure its existence. But closing this PR as we do not have intentions of merging for now.

Issue #761, opened 2019-07-27
Get realistic benchmarks into uber-calc (by Aug 2)


laser:
### Description

Notes:

- @laser will take over this project from @dignifiedquire 
- @dignifiedquire will provide code pointers and some prose explaining what his current solution does
- @porcuquine to describe the delta between what dig's current solution does and what he needs
- @laser will write acceptance criteria from ^^ and, if needed, more stories

### Acceptance criteria

- [ ] get ZigZag and micro benchmark output as JSON
- [ ] ZigZag benchmark output should include max resident set size used
- [ ] benchmark output should include Git SHA
- [ ] benchmark output should include information about the hardware (CPU model, total memory, etc.)
- [ ] single command for developer to run benchmarks in the cloud (on a Packet box)
- [ ] run benchmarks on Packet box with each push to rust-proofs `master`
- [ ] push benchmark JSON to benchmark server on each run

### Risks + pitfalls

- TBD

### Where to begin

- TBD

dignifiedquire:
## Overview

### Data collection

The data points below can currently be collected using two different scripts
- `filbase --benchy` collects information on ZigZag (rust-filbase)
- `cargo run --bin micro -p fil-proofs-tooling --release` collects the micro benchmarks (rust-fil-proofs) 

### Data storage

The current solution can store the information into prometheus. This might or might not be a good idea. An alternative that is possible much simpler and more flexible in the long run is to store the data into a sql or nosql database. 

### Reporting

- Visual: Initial prototype used Graphana, but this might not be that great, given certain limitations I have run into. Needs some exploration
- Machine readable: The hard requirement right now is to have a JSON file with the data points, that can be consumed by @porcuquine 

### Configuration

- rust-filbase has some flags on the commandline and we have a configuration file we can use to handle storage-proofs specific configs. In the long run that should likely hold all configurations, but needs more thought over the exact things that need configuration.

### Execution

- Manual runs, 
  - triggered via ssh
  - triggered via comment on a PR
- Automated runs for 
  - every commit to master of rust-fil-proofs


## Data Points needed

### ZigZag
### Parameters
- sector-size
  - Size of one sector. Unit: bytes
- base-degree
  - In-degree of the base depth-robust graph (DRG).
- expansion-degree
  - Maximum in-degree of the bipartite expander graph component of a ZigZag graph.
- sloth-iter
  - Number of iterations of sloth verifiable delay encoding (VDE) to perform.
- partitions
[...]

PR #780, opened 2019-08-05
feat(run benchy remotely)

laser:
Makes incremental progress towards #761 

## What's in this PR?

This changeset includes a script which will run `benchy.sh` on a remote server via SSH and write its output to the client's stdout.

### DIFF
--- a/fil-proofs-tooling/scripts/benchy-remote.sh
+++ b/fil-proofs-tooling/scripts/benchy-remote.sh
@@ -7,7 +7,7 @@ cd \$(mktemp -d)
 git clone -q https://github.com/filecoin-project/rust-fil-proofs.git
 cd rust-fil-proofs
 git checkout -q master
-RUSTFLAGS="-C target-cpu=native" ./fil-proofs-tooling/scripts/benchy.sh ${@:2}
+RUSTFLAGS="-Awarnings -C target-cpu=native" ./fil-proofs-tooling/scripts/benchy.sh ${@:2}
 EOF
 )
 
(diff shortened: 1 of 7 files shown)


fil-proofs-tooling/scripts/benchy-remote.sh
```diff
@@ -0,0 +1,15 @@
+#!/usr/bin/env bash
+
+set -e
+
+CMDS=$(cat <<EOF
+cd \$(mktemp -d)
+git clone -q https://github.com/filecoin-project/rust-fil-proofs.git
+cd rust-fil-proofs
+git checkout -q master
+/root/.cargo/bin/cargo build --release --all > /dev/null 2>&1
```
dignifiedquire: we should always build with `RUSTFLAGS="-C target-cpu=native"` for benchmarks

fil-proofs-tooling/scripts/benchy-remote.sh
```diff
@@ -0,0 +1,15 @@
+#!/usr/bin/env bash
+
+set -e
+
+CMDS=$(cat <<EOF
+cd \$(mktemp -d)
+git clone -q https://github.com/filecoin-project/rust-fil-proofs.git
+cd rust-fil-proofs
+git checkout -q master
```
dignifiedquire: what does `-q` do?

laser: It puts git in quiet mode, i.e. suppresses confirmation output

fil-proofs-tooling/scripts/benchy-remote.sh
laser: Done!

PR #791, opened 2019-08-07
refactor(storage-proofs): make base and expansion parents compile time constants

dignifiedquire:
This is the first step to pulling in optimizations from r2. 

Important changes:
- All graphs are now using the degrees 5 + 8, including all tests and exposed sector sizes.
- degrees are no inputs anymore for the circuits, need to validate this is okay
- This will require new params
- Adds compile-time feature `allow-small-degree` which allows graphs to be constructed with base and expansion degrees smaller than 5 and 8 respectively. This feature is enabled in the CI for the ignored `storage-proofs` tests.

### DIFF
--- a/storage-proofs/src/circuit/zigzag.rs
+++ b/storage-proofs/src/circuit/zigzag.rs
@@ -10,13 +10,14 @@ use crate::circuit::drgporep::{ComponentPrivateInputs, DrgPoRepCompound};
 use crate::circuit::variables::Root;
 use crate::compound_proof::{CircuitComponent, CompoundProof};
 use crate::drgporep::{self, DrgPoRep};
-use crate::drgraph::Graph;
+use crate::drgraph::{Graph, BASE_DEGREE};
 use crate::hasher::{HashFunction, Hasher};
 use crate::layered_drgporep::{self, Layers as LayersTrait};
 use crate::parameter_cache::{CacheableParameters, ParameterSetMetadata};
 use crate::porep;
 use crate::proof::ProofScheme;
 use crate::zigzag_drgporep::ZigZagDrgPoRep;
+use crate::zigzag_graph::{ZigZag, EXP_DEGREE};
 
 type Layers<'a, H, G> = Vec<
     Option<(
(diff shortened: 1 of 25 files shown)


dignifiedquire:
> +1 to that, not very impactful to my work but i usually reduce the `-m` and `--expansion` parameters to speed up certain memory benchmarks and also to test the impact of things like the parents cache

This will not be possible if we want to get the performance improvements in from r2 (except for changing the compile time constants). What @porcuquine is suggesting, simply enforces recording the values in the circuits

dignifiedquire:
At least that is how I understand the comment. Anything else is not really possible, if we want to get the improved hashing and use arrays.

dignifiedquire:
@porcuquine I disagree with that, this makes dev life harder, without providing real benefits.

dignifiedquire:
@porcuquine we need to be able to write things like `[u64; PARENTS_SIZE]`, in order to make this change worthwile. So while we can keep some paramters around, they wouldn't be actually used, and be just for show.

To make sure we can change things we should have a second set of constants for tests and debugging, which can be switched to by using a feature flag.

dignifiedquire:
alright, I talked with @porcuquine about this and we have come to the following agreement: 

- revert changes such that circuits do not change and the params stay the same
- add asserts that the params in the circuits, that they match the constants
- change the vanilla params such that they only can be created with the values of the constants, and assert that they match as part of proofing and verifying

schomatis:
Can the default values of the `zigzag` example also draw from theses constants?

https://github.com/filecoin-project/rust-fil-proofs/blob/d07e03a4c0ca317d38a0852725d68c62af101bae/filecoin-proofs/examples/zigzag.rs#L407-L419

DrPeterVanNostrand:
@schomatis yup, np

storage-proofs/src/circuit/zigzag.rs
```diff
@@ -84,6 +85,11 @@ impl<'a, H: Hasher> Circuit<Bls12> for ZigZagCircuit<'a, Bls12, H> {
         let graph = &self.public_params.graph;
         let layer_challenges = &self.public_params.layer_challenges;
 
+        if !cfg!(feature = "allow-small-degree") {
+            assert_eq!(graph.base_graph().degree(), BASE_DEGREE);
```
dignifiedquire: we should validate this in the base graph as well

DrPeterVanNostrand: @dignifiedquire ok, I added degree validation to `ZigZagBucketGraph`'s and `BucketGraph`'s constructors and changed all the places in the tests where `BucketGraph` was being instantiated with a degree not equal to `BASE_DEGREE`.

PR #793, opened 2019-08-07
Feat/787 serialized benchmark runs

laser:
## Why does this PR exist?

We need to ensure that benchmarks are run sequentially; we can't have more than one running at a time or they'll steal resources from each other.

## What's in this PR?

This PR adds a new script `retry.sh`, which retries a command after sleeping (with exponential backoff) if it exits with a given exit code.

This PR adds a new script `with-lock.sh`, which runs a command only after acquiring an exclusive lock. This script uses `mkdir`, which is an atomic operation, to emulate a mutex.

Together, these two scripts can be used to acquire an exclusive (system-wide) lock before running benchmarks.

### DIFF
--- a/fil-proofs-tooling/scripts/retry.sh
+++ b/fil-proofs-tooling/scripts/retry.sh
@@ -16,8 +16,8 @@ set -o pipefail
 # exit_code: The exit code to retry on.
 # attempts: The number of attempts to make.
 # sleep_millis: Multiplier for sleep between attempts. Examples:
-#     If multiplier is 1000, sleep intervals are 1, 2, 4, 8, 16, etc. seconds.
-#     If multiplier is 5000, sleep intervals are 5, 10, 20, 40, 80, etc. seconds.
+#     If multiplier is 1000, sleep intervals are 1, 4, 9, 16, etc. seconds.
+#     If multiplier is 5000, sleep intervals are 5, 20, 45, 80, 125, etc. seconds.
 
 exit_code=$1
 attempts=$2
--- a/fil-proofs-tooling/scripts/retry.sh
+++ b/fil-proofs-tooling/scripts/retry.sh
@@ -1,8 +1,5 @@
 #!/usr/bin/env bash
 
-# This software is freely available under the Apache 2.0 license.
-# Full license available here: https://www.apache.org/licenses/LICENSE-2.0.txt
-
 # Inspired by https://gist.github.com/reacocard/28611bfaa2395072119464521d48729a
 
 set -o errexit
--- a/fil-proofs-tooling/scripts/retry.sh
+++ b/fil-proofs-tooling/scripts/retry.sh
@@ -1,0 +1,43 @@
+#!/usr/bin/env bash
+
+# This software is freely available under the Apache 2.0 license.
+# Full license available here: https://www.apache.org/licenses/LICENSE-2.0.txt
+
+# Inspired by https://gist.github.com/reacocard/28611bfaa2395072119464521d48729a
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
+# Retry a command on a particular exit code, up to a max number of attempts,
+# with exponential backoff.
+# Invocation:
+#   err_retry exit_code attempts sleep_multiplier <command>
+# exit_code: The exit code to retry on.
+# attempts: The number of attempts to make.
+# sleep_millis: Multiplier for sleep between attempts. Examples:
+#     If multiplier is 1000, sleep intervals are 1, 2, 4, 8, 16, etc. seconds.
+#     If multiplier is 5000, sleep intervals are 5, 10, 20, 40, 80, etc. seconds.
+
+exit_code=$1
+attempts=$2
+sleep_millis=$3
+shift 3
+for attempt in `seq 1 $attempts`; do
+    # This weird construction lets us capture return codes under -o errexit
+    "$@" && rc=$? || rc=$?
+
+    if [[ ! $rc -eq $exit_code ]]; then
+        exit $rc
+    fi
+
+    if [[ $attempt -eq $attempts ]]; then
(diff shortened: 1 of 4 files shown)


fil-proofs-tooling/scripts/retry.sh
```diff
@@ -0,0 +1,44 @@
+#!/usr/bin/env bash
+
+# This software is freely available under the Apache 2.0 license.
+# Full license available here: https://www.apache.org/licenses/LICENSE-2.0.txt
```
dignifiedquire: no need for license header, this repo is already covered under mit/apache2 dual license

fil-proofs-tooling/scripts/retry.sh
```diff
@@ -0,0 +1,44 @@
...
+#     If multiplier is 5000, sleep intervals are 5, 10, 20, 40, 80, etc. seconds.
+
+exit_code=$1
+attempts=$2
+sleep_millis=$3
+shift 3
+
+for attempt in `seq 1 $attempts`; do
+    # This weird construction lets us capture return codes under -o errexit
+    "$@" && rc=$? || rc=$?
```
porcuquine: @laser, I see that this comes from the source gist. Do you understand what case is caught by the first part of 'this weird construction' that wouldn't be if the line above were just `rc=$?`?

fil-proofs-tooling/scripts/retry.sh
```diff
@@ -0,0 +1,44 @@
...
+
+    if [[ ! $rc -eq $exit_code ]]; then
+        exit $rc
+    fi
+
+    if [[ $attempt -eq $attempts ]]; then
+        exit $rc
+    fi
+
+    sleep_ms="$(($attempt * $attempt * $sleep_millis))"
```
porcuquine: This might be a fine backoff strategy, but it's not exponential as documented above, as far as I can tell. This seems to back off  by the square of the attempt. I would either change the implementation or the documentation. For example, the more accurate documentation would describe the first set of sleep intervals as `1, 4, 9, 16, 25`.

I also have to say that the precision/accuracy mismatch between the documentation and implementation do not inspire confidence in the source of this code snippet. It's likely not a major issue for this glue code, but as a matter of principle I'm concerned about introducing fuzziness into the `rust-fil-proofs` code base — so let's at least make it consistent.

porcuquine: If you actually want the sleep time to double on each iteration, you can repeatedly assign to a scale variable, doubling it each time, and multiply that by `$sleep_millis`.

fil-proofs-tooling/scripts/with-lock.sh
```diff
@@ -0,0 +1,21 @@
...
+failure_code=$1
+lockdir=$2
+shift 2
+
+if mkdir "$lockdir" > /dev/null 2>&1
+then
+    echo >&2 "successfully acquired lock (${lockdir})"
+
+    # Unlock (by removing dir) when the script finishes
+    trap 'echo >&2 "relinquishing lock (${lockdir})"; rm -rf "$lockdir"' EXIT
```
porcuquine: I guess this is safe, as long as we know that `/tmp` will be cleared between runs on CircleCI (or wherever this script will run). If there's any chance of an orphaned lock dir, we should manually delete it before running the `retry.sh` commands or use a mechanism which cannot lead to that problem.

porcuquine:
> ## Why does this PR exist?
> We need to ensure that benchmarks are run sequentially; we can't have more than one running at a time or they'll steal resources from each other.

Will the underlying problem still manifest if multiple jobs are running at once? This may not be an issue if the benchmarks are restricted to only run on master, but even then simultaneous jobs can be manually started? We can probably live with that risk if we're clear that it exists. If that's the case, we will need something more robust if/when we want to be able to run benchmarks against non-master commits — which is likely, eventually.

fil-proofs-tooling/scripts/retry.sh
laser: > This might be a fine backoff strategy, but it's not exponential as documented above, as far as I can tell.

That's fair. The source of the discrepancy between the documentation and implementation (specifically the `If multiplier is 1000, sleep intervals are` section) is me. I changed the original implementation but did not update the copied documentation.

fil-proofs-tooling/scripts/with-lock.sh
laser: > I guess this is safe, as long as we know that /tmp will be cleared between runs on CircleCI

This will definitely not be cleared between runs on CircleCI. The lock directory is on the Packet box, and by design it is not cleared between CircleCI runs (because it's used to coordinate across processes _on the Packet box).

fil-proofs-tooling/scripts/retry.sh
laser: The `"$@"` executes the command, and putting everything on the same line means that we evaluate the `rc=$?` is evaluated _even if the command produced a non-zero exit code_. Given that I'm using `set -o errexit`, running the command on its own line would cause the script to exit if the command failed.

porcuquine: Got it, thanks.

fil-proofs-tooling/scripts/retry.sh
porcuquine: If you say so. When I look back at the original it *appears* to have the same problem. I mention this only because maybe we're looking at something different if you think you introduced the issue.

fil-proofs-tooling/scripts/with-lock.sh
porcuquine: Ah… great.

Are you confident it *will* be cleared on restart of the Packet box if it goes down in the middle of a run? (If not, this seems like it will be a pain if/when it inevitably happens.)

Also, sorry for missing the point originally.

PR #794, opened 2019-08-08
feat(storage-proofs): test Feistel permutation for collisions

schomatis:


storage-proofs/src/zigzag_graph.rs
```diff
@@ -542,4 +542,26 @@ mod tests {
...
+        let mut shuffled: HashSet<u64> = HashSet::with_capacity((n * d) as usize);
+
+        let feistel_keys = &[1, 2, 3, 4];
+        let feistel_precomputed = feistel::precompute((n * d) as feistel::Index);
+
+        for i in 0..n {
+            for k in 0..d {
+                let permuted =
+                    feistel::permute(n * d, i * d + k, feistel_keys, feistel_precomputed);
+                assert!(shuffled.insert(permuted));
```
dignifiedquire: can you add an explanation why this assert has to hold please?

storage-proofs/src/zigzag_graph.rs
```diff
@@ -542,4 +542,26 @@ mod tests {
...
+            }
+        }
+        assert_eq!(shuffled.len(), (n * d) as usize);
```
dignifiedquire: same here

storage-proofs/src/zigzag_graph.rs
schomatis: indeed i should because that's exactly what i was trying to figure out with these tests

schomatis: done

PR #796, opened 2019-08-08
Disk-only MT

schomatis:
Use new `DiskStore` from https://github.com/filecoin-project/merkle_light/pull/27 instead of the file-`mmap` version, remove the `try_offload` logic since we work _always_ on offloaded (disk-only) stores now (when the `disk-trees` feature is set).

```
# in storage-proofs/
cargo build                                                                   \
    disk-trees                                                               &&
export TREES_DIR=rust-fil-proofs-trees-dir                                   &&
FIL_PROOFS_REPLICATED_TREES_DIR="$TREES_DIR"                                  \
RUST_LOG=trace                                                                \
nohup /usr/bin/time -v ../target/release/examples/zigzag                      \
    --size 1048576                                                            \
    --m 1  --expansion 0 --layers 5                                           \
     > zigzag-1gib-disk-only.out &
```
```
MASTER
	User time (seconds): 333.93
	System time (seconds): 24.50
	Percent of CPU this job got: 154%
	Elapsed (wall clock) time (h:mm:ss or m:ss): 3:52.39
	Maximum resident set size (kbytes): 4201012

DISK-ONLY PR
	User time (seconds): 306.70
	System time (seconds): 28.66                           <--- (similar disk time)
	Percent of CPU this job got: 147%
	Elapsed (wall clock) time (h:mm:ss or m:ss): 3:46.74
	Maximum resident set size (kbytes): 2138624            <--- (half memory used)
```

### DIFF
--- a/storage-proofs/Cargo.toml
+++ b/storage-proofs/Cargo.toml
@@ -15,7 +15,7 @@ bench = false
 bitvec = "0.5"
 rand = "0.4"
 libc = "0.2"
-merkletree = "0.7.1"
+merkletree = { git = "https://github.com/filecoin-project/merkle_light", branch = "feat/store/add-disk-only-version"}
 failure = "0.1"
 byteorder = "1"
 config = "0.9.3"
--- a/storage-proofs/Cargo.toml
+++ b/storage-proofs/Cargo.toml
@@ -15,7 +15,7 @@ bench = false
 bitvec = "0.5"
 rand = "0.4"
 libc = "0.2"
-merkletree = { git = "https://github.com/filecoin-project/merkle_light", branch = "feat/store/add-disk-only-version"}
+merkletree = "=0.8"
 failure = "0.1"
 byteorder = "1"
 config = "0.9.3"
--- a/storage-proofs/src/drgraph.rs
+++ b/storage-proofs/src/drgraph.rs
@@ -12,7 +12,7 @@ use crate::parameter_cache::ParameterSetMetadata;
 use crate::util::{data_at_node, NODE_SIZE};
 
 #[cfg(feature = "disk-trees")]
-use crate::merkle::DiskMmapStore;
+use crate::merkle::DiskStore;
 #[cfg(feature = "disk-trees")]
 use merkletree::merkle::next_pow2;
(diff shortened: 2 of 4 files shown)


schomatis:
@dignifiedquire What would be the most elegant way to add another option for the user, once he selected that he's willing to work with disk MTs, to select whether is an `mmap` or a disk-only version (without adding another bazillion compile time switches).

dignifiedquire:
This depends on if we want this to be a compile time or a runtime switch. 

runtime: 
- keep disk-trees feature
- add a config to switch between the two modes to our settings.rs

compile time
- remove disk-trees feature
- add `memmap-trees` and `file-trees` features to use those

dignifiedquire:
side note: the names of those features doesn't have to be perfect as long as we document them well enough. 
Some things to make your life easier

there is https://docs.rs/cfg-if/0.1.9/cfg_if/ and https://doc.rust-lang.org/std/macro.cfg.html to make your life easier when working with features

Also I personally would lean to compile time for now.

dignifiedquire:
@schomatis to clarify this is faster than the mmaped disk store according to the above measurements, correct?

schomatis:
It would _seem_ so, but let's be conservative and assume it doesn't, what I wanted to measure here is that disk performance doesn't significantly degrade transitioning from the file-`mmap` to the new disk-only store.

schomatis:
What should have been tested for completeness is that, generating one MT at a time and working directly on disks we now take only _one_ sector size of memory:

```
cargo build                                                                   \
    disk-trees                                                               &&
export TREES_DIR=rust-fil-proofs-trees-dir                                   &&
FIL_PROOFS_REPLICATED_TREES_DIR="$TREES_DIR"                                  \
FIL_PROOFS_GENERATE_MERKLE_TREES_IN_PARALLEL=0                                \
RUST_LOG=trace                                                                \
nohup /usr/bin/time -v ../target/release/examples/zigzag                      \
    --size 1048576                                                            \
    --m 1  --expansion 0 --layers 5                                           \
     > zigzag-1gib-disk-only.out &
```

```
	Maximum resident set size (kbytes): 1056524     <--- ONE sector size
```

PR #807, opened 2019-08-12
VDF PoSt: POST_SECTORS_COUNT to 512, from 2

### LINKED ISSUE #806: change "VDF PoSt" construction such that it uses POST_SECTORS_COUNT=512 instead of 2
### Description

The existing ("VDF PoSt") construction can prove on a fixed number of sectors at a time. This is controlled by the `POST_SECTORS_COUNT` value, and is currently set to 2. When the FFI `generate_post` consumer inputs more than 2 replica commitments, these are put into groups of `POST_SECTORS_COUNT` and one PoSt is generated per group, sequentially. This means that a single `generate_post` call can take a very long time when provided a large number of replica commitments, and number of proof-bytes returned can be very high, too.

Rational PoSt eliminates the need for `POST_SECTORS_COUNT`, but we haven't yet switched to Rational PoSt. In the mean time, we should bump `POST_SECTORS_COUNT` from 2 to 512.

laser:
##  Why does this PR exist?

Fixes #806.

## What's in this PR?

- change `POST_SECTORS_COUNT` from 2 to 512
- remove all the tests which assumed 2

## Notes on test deletion

The tests which I'd previously written all assumed a `POST_SECTORS_COUNT` of 2. Instead of updating them (at the cost of some time) to assume 512 (or to make no assumptions at all), I deleted them. Rational PoSt will obviate the need for `POST_SECTORS_COUNT` and the adapter layer.


filecoin-proofs/src/post_adapter.rs
```diff
@@ -66,7 +66,7 @@ pub fn generate_post_spread_input(
 
     for chunk in chunks {
         let mut input_parts: [(Option<String>, Commitment); POST_SECTORS_COUNT] =
-            Default::default();
+            array![(None, [0u8;32]); POST_SECTORS_COUNT];
```
dignifiedquire: blocking: none of these should not be stack allocated arrays, please use a Vec

dignifiedquire: 512 is simply too large for many stacks, the inner ones can stay as arrays

laser: Note for future programmers: We did a little [messing around](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=985091298e3700c643a0768cfe23c560) with `mem::size_of_val`, and it looks like a value of type `[(Option<String>, Commitment); 512]` takes up 28672B of stack. So we should be g2g.

steven004:
Is this planned to be in 0.4 user-devnet?  currently staging-devnet-0.4.1 is still using an old one which does not have this change.

PR #809, opened 2019-08-12
embed parameters.json in paramfetch binary

sidke:
# what's in this pr

- puts the raw `parameters.json` file as a string into the paramfetch binary

# why?

- in https://github.com/filecoin-project/rust-fil-sector-builder/issues/12, `cargo install` would be the simplest way to include `rfp` binaries in with a `rfsb` release, but `parameters.json` is not part of the files downloaded. by including it by default in `paramfetch`, we don't need to worry about packaging anything other than binaries

### DIFF
--- a/filecoin-proofs/src/bin/paramfetch.rs
+++ b/filecoin-proofs/src/bin/paramfetch.rs
@@ -22,6 +22,8 @@ use storage_proofs::parameter_cache::{
 const ERROR_PARAMETER_FILE: &str = "failed to find file in cache";
 const ERROR_PARAMETER_ID: &str = "failed to find key in manifest";
 
+const DEFAULT_PARAMETERS: &str = include_str!("../../../parameters.json");
+
 struct FetchProgress<R> {
     inner: R,
     progress_bar: ProgressBar<Stdout>,
@@ -117,14 +119,12 @@ fn fetch(matches: &ArgMatches) -> Result<()> {
     let retry = matches.is_present("retry");
     let gateway = matches.value_of("gateway").unwrap_or("https://ipfs.io");
 
-    if !json_path.exists() {
-        return Err(err_msg(format!(
-            "json file '{}' does not exist",
-            &json_path.to_str().unwrap_or("")
-        )));
-    }
+    let manifest = if json_path.exists() {
+        read_parameter_map_from_disk(&json_path)?
+    } else {
+        read_parameter_map_from_str(&DEFAULT_PARAMETERS)?
+    };
 
-    let manifest = read_parameter_map_from_disk(&json_path)?;
     let mut filenames = get_filenames_from_parameter_map(&manifest)?;
 
     println!("{} files in manifest...", filenames.len());
--- a/filecoin-proofs/src/bin/paramfetch.rs
+++ b/filecoin-proofs/src/bin/paramfetch.rs
@@ -3,7 +3,7 @@ use std::fs::{create_dir_all, rename, File};
 use std::io;
 use std::io::copy;
 use std::io::prelude::*;
-use std::io::Stdout;
+use std::io::{BufReader, Stdout};
 use std::path::PathBuf;
 use std::process::exit;
 
@@ -115,15 +115,27 @@ Set http_proxy/https_proxy environment variables to specify proxy for ipfs gatew
 }
 
 fn fetch(matches: &ArgMatches) -> Result<()> {
-    let json_path = PathBuf::from(matches.value_of("json").unwrap_or("./parameters.json"));
+    let manifest = if matches.is_present("json") {
+        let json_path = PathBuf::from(matches.value_of("json").unwrap());
+        println!("using json file: {:?}", json_path);
+
+        if !json_path.exists() {
+            return Err(err_msg(format!(
+                "json file '{}' does not exist",
+                &json_path.to_str().unwrap_or("")
+            )));
+        }
+
+        let file = File::open(json_path)?;
+        let reader = BufReader::new(file);
+        serde_json::from_reader(reader)?
+    } else {
+        println!("using built-in manifest");
+        serde_json::from_str(&DEFAULT_PARAMETERS)?
+    };
+
     let retry = matches.is_present("retry");
     let gateway = matches.value_of("gateway").unwrap_or("https://ipfs.io");
-
-    let manifest = if json_path.exists() {
-        read_parameter_map_from_disk(&json_path)?
-    } else {
-        read_parameter_map_from_str(&DEFAULT_PARAMETERS)?
-    };
 
     let mut filenames = get_filenames_from_parameter_map(&manifest)?;
 
--- a/filecoin-proofs/src/bin/paramfetch.rs
+++ b/filecoin-proofs/src/bin/paramfetch.rs
@@ -64,7 +64,7 @@ Set http_proxy/https_proxy environment variables to specify proxy for ipfs gatew
                 .takes_value(true)
                 .short("j")
                 .long("json")
-                .help("Use specific json file"),
+                .help("Use specific JSON file"),
         )
         .arg(
             Arg::with_name("gateway")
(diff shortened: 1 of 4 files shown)


filecoin-proofs/src/bin/paramfetch.rs
```diff
@@ -117,14 +119,12 @@ fn fetch(matches: &ArgMatches) -> Result<()> {
     let retry = matches.is_present("retry");
     let gateway = matches.value_of("gateway").unwrap_or("https://ipfs.io");
 
-    if !json_path.exists() {
-        return Err(err_msg(format!(
-            "json file '{}' does not exist",
-            &json_path.to_str().unwrap_or("")
-        )));
-    }
+    let manifest = if json_path.exists() {
```
laser: Please do the following:

- If the user provides a value for `json` and it is not a path for a file that exists on the disk, produce an error
- If the user does not provide a value for `json`, use the default-JSON baked into the binary
- After this change merges, modify the `paramfetch` section [of the go-filecoin wiki](https://github.com/filecoin-project/go-filecoin/wiki/Getting-Started#installing-from-binary) to demonstrate pointing `paramfetch` at the JSON manifest using the `--json` arg

I've [already written some tests for `paramfetch`](https://github.com/filecoin-project/rust-fil-proofs/blob/79b92f77b96209e387630d77bfa124b71c8610aa/filecoin-proofs/tests/paramfetch/prompts_to_fetch.rs#L30). I'd really like it if you'd augment the exist test suite with:

- a test which verifies that `paramfetch` fails with a non-zero exit code if given a `json` value which is not a path to a file on disk
- a test which verifies that `paramfetch` fails if given a `json` value which does not contain JSON which it can deserialize to a `ParameterMap`
-  a test which verifies that `paramfetch`, when not provided a `json` argument, [checks](https://github.com/filecoin-project/rust-fil-proofs/blob/79b92f77b96209e387630d77bfa124b71c8610aa/filecoin-proofs/tests/paramfetch/prompts_to_fetch.rs#L75) for the existence of the files in the manifest

filecoin-proofs/src/param.rs
```diff
@@ -25,6 +25,13 @@ pub struct ParameterData {
     pub sector_size: u64,
 }
 
+// Deserializes bytes from the provided path into a ParameterMap
+pub fn read_parameter_map_from_str(string: &str) -> Result<ParameterMap> {
```
laser: If this function is not used outside of `paramfetch.rs`, I see no need to put it in `param.rs`. Do you?

filecoin-proofs/tests/paramfetch/prompts_to_fetch.rs
```diff
@@ -146,3 +155,41 @@ fn fetches_vk_even_if_sector_size_does_not_match() -> Result<(), FailureError> {
...
+    let manifest_pbuf = tmp_manifest(None)?;
+
+    let mut file = File::create(&manifest_pbuf)?;
+    file.write_all(b"invalid json")?;
+
+    let mut session = ParamFetchSessionBuilder::new(Some(manifest_pbuf))
+        .with_session_timeout_ms(1000)
+        .build();
+
+    session.exp_string("fatal error: expected value at line 1 column 1")?;
```
laser: I would like to see a string like: `fatal error: file with path '/invalid/path' did not contain valid manifest JSON`

filecoin-proofs/tests/paramfetch/prompts_to_fetch.rs
```diff
@@ -146,3 +155,41 @@ fn fetches_vk_even_if_sector_size_does_not_match() -> Result<(), FailureError> {
...
+#[test]
+fn no_json_path_uses_default_manifest() -> Result<(), FailureError> {
+    let mut session = ParamFetchSessionBuilder::new(None)
+        .with_session_timeout_ms(1000)
+        .build();
+
+    session.exp_string("using built-in manifest")?;
+
+    Ok(())
+}
```
laser: I would like to see one more test which verifies that paramfetch, when not provided a json argument, checks for the existence of the files in the manifest. To be more specific, I think you should:

- modify the test to deserialize the contents of the `parameters.json` file into some `ParameterMap`
- run `paramfetch` without `--json` and verify you see output _for each asset in the map_

The `using built-in manifest` check is fine, but I want a 'lil more.

filecoin-proofs/src/bin/paramfetch.rs
```diff
@@ -113,18 +115,38 @@ Set http_proxy/https_proxy environment variables to specify proxy for ipfs gatew
...
+                "json file '{}' does not exist",
+                &json_path.to_str().unwrap_or("")
+            )));
+        }
+
+        let file = File::open(&json_path)?;
+        let reader = BufReader::new(file);
+        let result = serde_json::from_reader(reader);
+
+        if result.is_err() {
```
laser: Please use `.map_err` here instead of `result.is_err()` and `unwrap()`.

filecoin-proofs/src/bin/paramfetch.rs
```diff
@@ -113,18 +115,38 @@ Set http_proxy/https_proxy environment variables to specify proxy for ipfs gatew
...
+                &json_path.to_str().unwrap_or(""),
+                result.err().unwrap(),
+            )));
+        }
+
+        result.unwrap()
+    } else {
+        println!("using built-in manifest");
+        serde_json::from_str(&DEFAULT_PARAMETERS)?
+    };
```
laser: See note above. If you use `map_err`, then you can use `?` after the `else` block, e.g.

```rust
let manifest = if matches.is_present("json") {
    serde_json::from_reader(reader).map_err(|err| {
        Err(err_msg(format!( // etc
    })
} else {
    println!("using built-in manifest");
    serde_json::from_str(&DEFAULT_PARAMETERS)
}?;

```

filecoin-proofs/tests/paramfetch/prompts_to_fetch.rs
```diff
@@ -146,3 +155,50 @@ fn fetches_vk_even_if_sector_size_does_not_match() -> Result<(), FailureError> {
...
     Ok(())
 }
+
+#[test]
+fn invalid_json_path_produces_error() -> Result<(), FailureError> {
+    let mut session = ParamFetchSessionBuilder::new(Some(PathBuf::from("/invalid/path")))
+        .with_session_timeout_ms(1000)
+        .build();
+
+    session.exp_string("fatal error: json file '/invalid/path' does not exist")?;
```
laser: JSON is an acronym. Please update to read: "fatal error: JSON file '/invalid/path' does not exist.

filecoin-proofs/src/bin/paramfetch.rs
```diff
@@ -113,18 +115,38 @@ Set http_proxy/https_proxy environment variables to specify proxy for ipfs gatew
...
+            )));
+        }
+
+        let file = File::open(&json_path)?;
+        let reader = BufReader::new(file);
+        let result = serde_json::from_reader(reader);
+
+        if result.is_err() {
+            return Err(err_msg(format!(
+                "json file '{}' did not parse correctly: {}",
```
laser: Replace with: `"JSON file '{}' did not parse correctly: {}"`

filecoin-proofs/tests/paramfetch/prompts_to_fetch.rs
```diff
@@ -146,3 +155,50 @@ fn fetches_vk_even_if_sector_size_does_not_match() -> Result<(), FailureError> {
...
+    let reader = BufReader::new(file);
+    let manifest: ParameterMap = serde_json::from_reader(reader)?;
+
+    let mut session = ParamFetchSessionBuilder::new(None)
+        .with_session_timeout_ms(1000)
+        .build();
+
+    session.exp_string("using built-in manifest")?;
+
+    for parameter in manifest.keys() {
```
laser: If the order of the keys in `manifest` is nondeterministic, will this test occasionally fail? I don't recall how `session.exp_string` works.

sidke: afaict the order of the keys is deterministic. `BTreeMap` is a sorted map

PR #812, opened 2019-08-13
feat(filecoin-proofs): Use ipget to download params

dignifiedquire:
This downloads and uses `igpet` to download params, instead of going through the gateway, which has been pretty flaky for us. 

We loose progress bars, as `ipget` doesn't support this atm.

### DIFF
--- a/filecoin-proofs/Cargo.toml
+++ b/filecoin-proofs/Cargo.toml
@@ -37,6 +37,8 @@ pretty_env_logger = "0.3.0"
 env_proxy = "0.3"
 array-macro = "1.0.4"
 os_type = "2.2.0"
+flate2 = { version = "1.0.9", features = ["rust_backend"]}
+tar = "0.4.26"
 
 [dependencies.reqwest]
 version = "0.9"
--- a/filecoin-proofs/Cargo.toml
+++ b/filecoin-proofs/Cargo.toml
@@ -36,6 +36,7 @@ log = "0.4.7"
 pretty_env_logger = "0.3.0"
 env_proxy = "0.3"
 array-macro = "1.0.4"
+os_type = "2.2.0"
 
 [dependencies.reqwest]
 version = "0.9"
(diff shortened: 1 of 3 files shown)


anorth:
Should close https://github.com/protocol/infra/issues/461

laser:
@dignifiedquire - I would really like to see some tests for this. 

We have a variety of tests for `paramfetch` already written (see `prompts_to_fetch.rs`); you should be able to augment those. 

I suggest modifying `paramfetch` to accept an optional `--ipget-bin` argument (like [`paramcache`](https://github.com/filecoin-project/rust-fil-proofs/blob/65973f60774df0dadc56b67c8ca008f886497393/filecoin-proofs/src/bin/parampublish.rs#L54) does with `--ipfs-bin`). At test time, you can provide a fake [like we do for `parampublish`](https://github.com/filecoin-project/rust-fil-proofs/blob/79b92f77b96209e387630d77bfa124b71c8610aa/filecoin-proofs/src/bin/fakeipfsadd.rs#L5).

gmasgras:
I doubt that using `ipget` will be any faster. 
When pinning a CID, it gets persisted to the ipfs-cluster which should always be peered with the gateways so there is a much smaller discovery cost than running ipget locally which involves bootstrapping + discovery.

laser:
@gmasgras

> I doubt that using ipget will be any faster.

As it currently stands, no developer is able to download all published (V11) Groth parameters via the IPFS Gateway. On my machine and on @acruikshank's, we get about 1/3rd of the way through each ~1.5GiB file and then our transfer fails with either `error: error reading a body from connection: end of file before message length reached` or `error: timed out`. 

You should be able to reproduce this issue yourself by [downloading the 0.6.1 release](https://github.com/filecoin-project/rust-fil-proofs/releases/tag/23b7b9fa9028ca29), extracting `paramfetch` and `parameters.json` somewhere, and then running `paramfetch --all --json=./parameters.json --verbose`.

I have been able to download all V11 Groth parameters using this ipget-based solution, but it does take ~2 hours.

laser:
Note: This changeset removes the Groth parameter and verifying key-downloading progress bar, because `ipget` doesn't give any such output. This changeset also removes the `--gateway` argument; `ipget` doesn't support this, either.

PR #817, opened 2019-08-16
Paramfetch flex

### LINKED ISSUE #818: paramfetch should pull Groth parameters from IPFS using ipget instead of from IPFS Gateway using HTTP
### Description

Thus far, no developer has been able to download the full set of V11 Groth parameters and keys from the IPFS Gateway using HTTP. Downloads progress to ~30% completion and then bandwidth dwindles to 0B/sec. 

Instead of using the IPFS Gateway to download Groth parameters, `paramfetch` should use `ipget` to interact directly with the IPFS network.

dignifiedquire:
Fixes #818 

- allow for specific ipget version
- allow for passthrough of arguments of ipget

This is in anticipation of using the new ipget https://github.com/ipfs/ipget/pull/59

### DIFF
--- a/filecoin-proofs/src/bin/paramfetch.rs
+++ b/filecoin-proofs/src/bin/paramfetch.rs
@@ -103,6 +103,12 @@ Defaults to '{}'
                 .help("Use specific ipget binary instead of looking for (or installing) one in /var/tmp/ipget/ipget"),
         )
         .arg(
+            Arg::with_name("ipget-args")
+                .takes_value(true)
+                .long("ipget-args")
+                .help("Specify additional arguments for ipget")
+        )
+        .arg(
             Arg::with_name("ipget-version")
                 .long("ipget-version")
                 .takes_value(true)
@@ -199,6 +205,7 @@ fn fetch(matches: &ArgMatches) -> Result<()> {
     let is_verbose = matches.is_present("verbose");
     let ipget_bin_path = matches.value_of("ipget-bin");
     let ipget_version = matches.value_of("ipget-version").unwrap();
+    let ipget_args = matches.value_of("ipget-args");
 
     // Make sure we have ipget available
     if ipget_bin_path.is_none() {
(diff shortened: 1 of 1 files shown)


filecoin-proofs/src/bin/paramfetch.rs
```diff
@@ -102,6 +102,19 @@ Defaults to '{}'
                 .long("ipget-bin")
                 .help("Use specific ipget binary instead of looking for (or installing) one in /var/tmp/ipget/ipget"),
         )
+        .arg(
+            Arg::with_name("ipget-args")
```
laser: What motivates the existence of this ipget-args argument? What arguments does ipget provide that we want to expose to paramfetch users?

laser: Blocking: This argument cannot be combined with `ipget-bin`. Please do use `conflicts_with`.

filecoin-proofs/src/bin/paramfetch.rs
```diff
@@ -102,6 +102,19 @@ Defaults to '{}'
...
+                .takes_value(true)
+                .long("ipget-args")
+                .help("Specify additional arguments for ipget")
+        )
+        .arg(
+            Arg::with_name("ipget-version")
```
laser: Blocking: This argument cannot be combined with `ipget-bin`. Please do use `conflicts_with`.

filecoin-proofs/src/bin/paramfetch.rs
```diff
@@ -102,6 +102,19 @@ Defaults to '{}'
                 .long("ipget-bin")
```
laser: Blocking: This argument cannot be combined with `ipget-args` or `ipget-version`. Please do use `conflicts_with`.

filecoin-proofs/src/bin/paramfetch.rs
dignifiedquire: From the ipget pr

> Also, add a -peers flag for directly connecting to a set of peers on start.

This will be very useful to speed up connection and downloads

filecoin-proofs/src/bin/paramfetch.rs
dignifiedquire: we should be able to combine this with `ipget-args`

PR #834, opened 2019-08-22
fix(filecoin-proofs): post: avoid generating trees for padded inputs

dignifiedquire:
Needs more testing, but the idea is to avoid the fixed memory requirements, but only actually build trees for the sectors that are real, and rereference the last one.


dignifiedquire:
I verified that I can run the `simple` example with 256Mib sectors using this, which I could not before.

dignifiedquire:
not needed anymore, included in rational post

Issue #831, opened 2019-08-22
Make disk the default storage for MTs


schomatis:
The last implementation stage was merged in https://github.com/filecoin-project/rust-fil-proofs/pull/796. Have the `disk-trees` feature as a default.

Blockers:

* Manage generated files (at the moment we don't clean anything and it quickly fills up the disk), see https://github.com/filecoin-project/rust-fil-proofs#memory.

* Benchmark speed degradation. Recently I reported that disk-MTs may have a negligible impact of less than 5% replication time but that assessment may be incorrect and only apply to the 2x sector size case, when we still generate MTs in parallel with encoding; in the 1x case, with `FIL_PROOFS_GENERATE_MERKLE_TREES_IN_PARALLEL=0` this may have a bigger impact, we need to benchmark this, see https://github.com/filecoin-project/rust-fil-proofs/issues/826 for initial thoughts on how to improve this.

porcuquine:
As we measure things. We should target the case of parallel generation. This needs to be the normal mode of replication, with replication time bounded by serial encoding. The goal is to minimize memory consumption *without pausing replication*. It sounds like what you are saying is that under those conditions we achieve 2x sector size memory consumption. Can you please clarify whether that is *for Merkle tree generation* or if it also includes the replica itself being encoded in memory?

schomatis:
The easiest solution around managing files I can think of is generating temporal files that get deleted once the MT is dropped (assuming we don't allow reusing old MTs, at least for this initial iteration).

We used to need specific non-temporal files for the file-`mmap` variant (to offload the mapping temporarily and load it back again when needed) but now that has been completely replaced by the disk-only store.

porcuquine:
I think it's okay to throw away the trees for now. In the future and in general, I think miners would prefer to keep disk-cached trees if they have the space — since generating the trees is expensive.

schomatis:
> Can you please clarify whether that is _for Merkle tree generation_ or if it also includes the replica itself being encoded in memory?

I always benchmark memory for the entire replication process. So I think that's number 2 (I may be misinterpreting the question though).

schomatis:
Ok, so if 2x is an acceptable RAM-to-sector-size ratio I can take care of the file management and we can move ahead with this. For parallel MT generation and encoding, replication time (will confirm with more tests) was pretty much under control, see https://github.com/filecoin-project/rust-fil-proofs/pull/796#issue-305565394 as an example.

[...]

PR #836, opened 2019-08-22
feat(storage-proofs): use disk-trees feature as default

schomatis:
Alternatively we could invert the feature and have a `ram-trees`/`mem-trees` feature.

* [x] Adapt `disk-tree` test accordingly.

* [x] Update readme, we don't need to set this feature. Also, the RSS will be now 1x sector size.

Closes https://github.com/filecoin-project/rust-fil-proofs/issues/831.

-------------------------------

Benchmark:

```
RUSTFLAGS="-C target-cpu=native" cargo build --bin benchy --release

RUST_LOG=info /usr/bin/time -v ./target/release/benchy zigzag --size 262144
```
```
MASTER (disk-trees *not* set):
	User time (seconds): 10724.92
	System time (seconds): 12.24
	Percent of CPU this job got: 643%
	Elapsed (wall clock) time (h:mm:ss or m:ss): 27:48.98
	Maximum resident set size (kbytes): 6693416

PR (disk-trees set as default):
	User time (seconds): 10712.94
	System time (seconds): 20.55
	Percent of CPU this job got: 648%
	Elapsed (wall clock) time (h:mm:ss or m:ss): 27:34.49
	Maximum resident set size (kbytes): 926748
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -159,24 +159,23 @@ To check that it's working you can inspect the replication log to find `using pa
 
 (You can also verify if the cache is working by inspecting the time each layer takes to encode, `encoding, layer:` in the log, where the first two layers, forward and reverse, will take more time than the rest to populate the cache while the remaining 8 should see a considerable time drop.)
 
-### Memory
-
-We try to generate the MTs in parallel to speed up the process but that takes 2 sector sizes per each layer (e.g., a 1 GiB sector may require, in the worst case scenario, up to 20 GiB of memory to hold the MTs alone). To reduce that (at the cost of speed) we have the (experimental) `disk-trees` feature to offload the MTs to disk when we don't use them. For example, to run the `zigzag` example with this feature you'd need to indicate so to `cargo`,
+In the most extreme case, to reduce time at the cost of *a lot* of memory consumption you can turn off the feature that stores MTs on disk (`disk-trees`) to generate them all on RAM and avoid disk I/O (if the HW doesn't have enough RAM to handle the MTs, roughly 20x the sector size, this won't have the desired effect as the OS will start backing them on disk anyway). For example, to run the `zigzag` example with this feature turned off you'd need to indicate so to `cargo`,
 
 ```
-# From inside the `storage-proofs` directory, where this feature
-# needs to be activated:
-cargo build                                                                   \
+# NEEDS TO BE RUN INSIDE `storage-proofs\` directory
+# for the `--no-default-features` to take effect.
+cargo run                                                                     \
   -p filecoin-proofs                                                          \
   --release                                                                   \
   --example zigzag                                                            \
-  --features                                                                  \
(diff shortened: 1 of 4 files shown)


dignifiedquire:
do you have a before benchmark, with mem trees?

storage-proofs/Cargo.toml
```diff
@@ -48,7 +48,7 @@ log = "0.4.7"
 pretty_env_logger = "0.3.0"
 
 [features]
-default = []
+default = ["disk-trees"]
 simd = []
 asm = ["sha2/sha2-asm"]
 disk-trees = []
```
dignifiedquire: `fil-proofs-tooling` and `filecoin-proofs` should get this feature too, which just proxies this, so there is control at that level

dignifiedquire: actually fine to not do this for now, if we switch this on by default

schomatis:
With https://github.com/filecoin-project/rust-fil-proofs/pull/837 merged this is now the final PR in the disk-MTs queue, will be resolved this week.

README.md
```diff
@@ -159,24 +159,23 @@ To check that it's working you can inspect the replication log to find `using pa
...
 
-### Memory
-
-We try to generate the MTs in parallel to speed up the process but that takes 2 sector sizes per each layer (e.g., a 1 GiB sector may require, in the worst case scenario, up to 20 GiB of memory to hold the MTs alone). To reduce that (at the cost of speed) we have the (experimental) `disk-trees` feature to offload the MTs to disk when we don't use them. For example, to run the `zigzag` example with this feature you'd need to indicate so to `cargo`,
+In the most extreme case, to reduce time at the cost of *a lot* of memory consumption you can turn off the feature that stores MTs on disk (`disk-trees`) to generate them all on RAM and avoid disk I/O (if the HW doesn't have enough RAM to handle the MTs, roughly 20x the sector size, this won't have the desired effect as the OS will start backing them on disk anyway). For example, to run the `zigzag` example with this feature turned off you'd need to indicate so to `cargo`,
```
dignifiedquire: I don't think this is really true, I am not seeing any speed differences between mmemap and disk versions

schomatis: Me neither, but this might also depend on the HW, "the most extreme case" aimed at conveying something like that, could be rephrased.

schomatis: Now that you mention it, actually the mmap-MT should be replaced with the vector-MT (which would be selected only if the disk-trees is explicitly turned off).

schomatis:
@dignifiedquire As there's no easy way to disable a default feature I think it would be better to have an optional `mem-trees` feature instead, that would switch from the (now) default disk MT to a vector based one when set.

dignifiedquire:
@schomatis I agree switching to mem-tree feature and removing disk-tree feature makes more sense

Issue #826, opened 2019-08-20
Smartly share the data being used for encoding and MT generation


schomatis:
In the default case of `transform_and_replicate_layers` where we generate MTs in parallel with layer encoding we currently duplicate the _entire_ sector to avoid collisions between the MT that is reading the data to generate the tree and the VDE that is modifying that same data to generate the next layer.

To avoid that duplication and reduce memory consumption we provide an option to pipe those two operations and avoid data races, but that comes at the extreme cost of the MT being "another" VDE with comparable delay times which has a significant speed degradation.

Ideally those two process should be able to operate together on the same sector provided they do so in a more granular fashion, e.g., if we subdivide the sector in blocks of say, 64KiB, the only restriction would be that the VDE should mutate a block with a lower index that the one the MT is reading (implied is that the other blocks have already been incorporated into the tree).

This would (I think) almost remove performance degradation at both, speed and memory. Memory because we can comfortably choose a block size orders of magnitude smaller than any production-ready sector size, and speed because the MT should be faster than VDE and not be gating its progress.

This proposal has some implementation considerations (like that VDE sometimes traverses the layer in a forward manner and sometimes in reverse, which the MT currently doesn't) which have a clear solution, my question would be if this violates some of the assumptions of the ZigZag construction though.

porcuquine:
Just a note that in practice merkle-tree generation is only faster than encoding when sufficient parallelism. This doesn't contradict your analysis, but it adds a nuance to 'the MT should be faster than VDE '.

Bear in mind also that we are moving to a construction which will not generate one merkle tree per layer. This will lessen that consideration but also means we cannot generate trees as we go. Rather, we will need to accumulate leaf values into an aggregate hash then generate the whole tree at the end. I'm not trying to fully specify this here because design will be forthcoming. Just an early thought to keep assumptions aligned.

schomatis:
> Just a note that in practice merkle-tree generation is only faster than encoding when sufficient parallelism. This doesn't contradict your analysis, but it adds a nuance to 'the MT should be faster than VDE '.

Good point, I'm always assuming that scenario, in this case I think that if the user doesn't have enough parallelism then it has not much to win from parallel MT/VDE processing in the first place.

> we will need to accumulate leaf values into an aggregate hash then generate the whole tree at the end.

[...]

PR #837, opened 2019-08-23
Remove extra `data` allocation used for MT generation

schomatis:
Towards making disk-MTs default (https://github.com/filecoin-project/rust-fil-proofs/issues/831), a suboptimal solution to https://github.com/filecoin-project/rust-fil-proofs/issues/826. (Using new features from MT repo in https://github.com/filecoin-project/merkle_light/pull/31).

* [x] Update `merkletree` to (future) `0.10` release.

* [x] Remove FIL_PROOFS_GENERATE_MERKLE_TREES_IN_PARALLEL, that case would no longer be needed, it's too extreme compared to this new architecture.

* [x] Call `populate_leaves` directly instead of copying its code (should remove `Algorithm` then and the ambiguities introduced there).

* [x] Remove `VecStore` if not used, we use `MmapStore` by default.

-------------------------------------------

For speed we both build MT and encode in parallel, the price to pay in an entire sector size duplication to decouple the two process so the modification of the encoding of the new layer doesn't affect the MT being build off the current one. For memory we do both processes sequentially, so we don't need the extra allocation to decouple them, greatly slowing down replication as a whole.

In the middle of these two extremes there seems to be an acceptable scenario that removes the duplication by adding a negligible (compared to encoding) subpart of the process in sequence. What actually needs to be preserved from encoding modification for the MT is just the leaves store creation, once the MT has its bottom store populated all its other work depends on that, independently of what the encoding does. So encoding only needs to wait for the MT to populate its leaves store (from the original `data`) before proceeding (and both processes continue in parallel from there). This sequential part I think (benchmarking to be sure) is negligible because populating the leaves store *does not involve hashing*, just (in the most expensive case of the `DiskStore`) writing the sector to disk, which is not minor, but still compared to the actual encoding process seems like a small price to pay to bring the RAM requirements from 2x to 1x sector size.

This is just an intermediary solution, a much better alternative could be implemented so the encoding phase doesn't have to wait for the MT to copy the _entire_ store before beginning (by partitioning it so both process work on `data` simultaneously, see https://github.com/filecoin-project/rust-fil-proofs/issues/826).

-------------------------------------------

Benchmark:

```
# in storage-proofs/
cargo build                                                                   \
    disk-trees                                                               &&
RUST_LOG=trace                                                                \
nohup /usr/bin/time -v ../target/release/examples/zigzag                      \
    --size 1048576                                                            \
    --m 1  --expansion 0 --layers 5                                           \
     > zigzag-1gib-disk-only.out &
```

```
MASTER:

	User time (seconds): 301.13
	System time (seconds): 31.36
	Percent of CPU this job got: 149%
	Elapsed (wall clock) time (h:mm:ss or m:ss): 3:42.88

	Maximum resident set size (kbytes): 2135512

PR:
	User time (seconds): 297.50
	System time (seconds): 25.88
	Percent of CPU this job got: 135%
	Elapsed (wall clock) time (h:mm:ss or m:ss): 3:58.06

	Maximum resident set size (kbytes): 1057256             <-- 1x sector size
```

Why does this PR has more wall clock time than `master` even though user/system times are less than it? Because it got less percentage of CPU since we are de-parallelizing part of it, however, this is noticeable here only because we are basically doing no hashing (`--m 1  --expansion 0`), a normal replication of a sector of 1 GiB takes 2 hours and this is done in 3 minutes, we should benchmark the full example to test this.

### DIFF
--- a/storage-proofs/src/merkle.rs
+++ b/storage-proofs/src/merkle.rs
@@ -5,8 +5,8 @@ use std::marker::PhantomData;
 // Reexport here, so we don't depend on merkletree directly in other places.
 pub use merkletree::hash::Algorithm;
 use merkletree::merkle;
-#[cfg(not(feature = "disk-trees"))]
-use merkletree::merkle::MmapStore;
+// #[cfg(not(feature = "disk-trees"))]
+// use merkletree::merkle::MmapStore;
 use merkletree::merkle::VecStore;
 use merkletree::proof;
 use paired::bls12_381::Fr;
@@ -16,8 +16,8 @@ use crate::hasher::{Domain, Hasher};
 // `mmap`ed `MerkleTree` (replacing the previously `Vec`-backed
 // `MerkleTree`, now encapsulated in `merkle::VecStore` and exposed
 // as `VecMerkleTree`).
-pub use merkletree::merkle::Store;
 pub use merkletree::merkle::next_pow2;
+pub use merkletree::merkle::Store;
 pub use merkletree::merkle::BUILD_LEAVES_BLOCK_SIZE;
 pub type DiskStore<E> = merkletree::merkle::DiskStore<E>;
 pub type MmapStore<E> = merkletree::merkle::MmapStore<E>;
--- a/storage-proofs/src/merkle.rs
+++ b/storage-proofs/src/merkle.rs
@@ -7,21 +7,16 @@ pub use merkletree::hash::Algorithm;
 use merkletree::merkle;
 // #[cfg(not(feature = "disk-trees"))]
 // use merkletree::merkle::MmapStore;
-use merkletree::merkle::VecStore;
 use merkletree::proof;
 use paired::bls12_381::Fr;
 
 use crate::hasher::{Domain, Hasher};
 
-// `mmap`ed `MerkleTree` (replacing the previously `Vec`-backed
-// `MerkleTree`, now encapsulated in `merkle::VecStore` and exposed
-// as `VecMerkleTree`).
 pub use merkletree::merkle::next_pow2;
 pub use merkletree::merkle::populate_leaves;
 pub use merkletree::merkle::Store;
 pub type DiskStore<E> = merkletree::merkle::DiskStore<E>;
 pub type MmapStore<E> = merkletree::merkle::MmapStore<E>;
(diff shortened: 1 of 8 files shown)


storage-proofs/src/merkle.rs
```diff
@@ -3,10 +3,10 @@
 use std::marker::PhantomData;
 
 // Reexport here, so we don't depend on merkletree directly in other places.
-use merkletree::hash::Algorithm;
+pub use merkletree::hash::Algorithm;
 use merkletree::merkle;
-#[cfg(not(feature = "disk-trees"))]
-use merkletree::merkle::MmapStore;
+// #[cfg(not(feature = "disk-trees"))]
+// use merkletree::merkle::MmapStore;
```
dignifiedquire: this should probably be fixed right

schomatis: dont review yet, poc

dignifiedquire:
@schomatis you posted empty results

schomatis:
placeholders so i don't forget, benchmarking takes a loooong time

dignifiedquire:
uhmmm on my machine I got this (on master)

```
--size 262144
Elapsed (wall clock) time (h:mm:ss or m:ss): 7:49.6
Maximum resident set size (kbytes): 7017944
```

which is like 29x sector size, I suspect this is memap behaviour and the fact that this machine has a lot of memory available

dignifiedquire:
this branch on my machine:

```
Elapsed (wall clock) time (h:mm:ss or m:ss): 7:53.97
Maximum resident set size (kbytes): 6692916
```

dignifiedquire:
Doing some new runs, `master` with `disk-trees` enabled:

```
Sector Size: 4Gib
Wall clock time: 2:09:22
Maximum resident set size (kbytes): 9116856 // 2.225x SectorSize
```

dignifiedquire:
On this branch (running benchy zigzag)


```
Sector Size: 4Gib
Wall clock time: 2:10:42
Maximum resident set size (kbytes): 4861648 // 1.2x SectorSize
```

dignifiedquire:
I was able to replicate `16GiB` with `16.6GiB` max memory usage on this PR. 
@schomatis lets get this ready and merged, it is really helpful

PR #840, opened 2019-08-23
feat(add SectorId type and transforms)

laser:
## Why does this PR exist?

`u64` doesn't tell us much about what values of this type represent (in our domain).

## What's in this PR?

This changeset uses [the newtype pattern](https://doc.rust-lang.org/1.0.0/style/features/types/newtype.html) to wrap `u64` up into a `SectorId` type.

### DIFF
--- a/filecoin-proofs/src/api.rs
+++ b/filecoin-proofs/src/api.rs
@@ -47,6 +47,7 @@ use storage_proofs::piece_inclusion_proof::{
 use storage_proofs::porep::{replica_id, PoRep, Tau};
 use storage_proofs::proof::NoRequirements;
 use storage_proofs::rational_post;
+use storage_proofs::sector_id::SectorId;
 use storage_proofs::zigzag_drgporep::ZigZagDrgPoRep;
 use tempfile::tempfile;
 
@@ -129,7 +130,7 @@ pub fn generate_post(
     post_config: PoStConfig,
     challenge_seed: &ChallengeSeed,
     replicas: Vec<ReplicaInfo>,
-    faults: &[u64],
+    faults: &[SectorId],
 ) -> error::Result<Vec<u8>> {
     let sector_count = replicas.len() as u64;
     let sector_size = u64::from(PaddedBytesAmount::from(post_config));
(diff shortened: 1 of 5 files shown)


filecoin-proofs/src/api.rs
```diff
@@ -155,19 +156,19 @@ pub fn generate_post(
...
-        &faults,
+        faults,
     );
 
     // Match the replicas to the challenges, as these are the only ones required.
     let challenged_replicas: Vec<_> = challenges
         .iter()
         .map(|c| {
-            if let Some(replica) = replicas.get(c.sector as usize) {
+            if let Some(replica) = replicas.get(u64::from(c.sector_id) as usize) {
```
dignifiedquire: blocking: if we do this, `Challenge` should contain a `SectorId`

filecoin-proofs/src/api.rs
```diff
@@ -155,19 +156,19 @@ pub fn generate_post(
...
                 Ok(replica)
             } else {
                 Err(format_err!(
-                    "Invalid challenge generated: {}, only {} sectors are being proven",
-                    c.sector as usize,
+                    "Invalid challenge generated: {:?}, only {} sectors are being proven",
+                    c.sector_id,
```
dignifiedquire: please implement `Display` for `SectorId`

storage-proofs/src/sector_id.rs
```diff
@@ -0,0 +1,30 @@
...
+    }
+}
+
+impl From<SectorId> for u64 {
+    fn from(n: SectorId) -> Self {
+        n.0
+    }
+}
+
+impl std::cmp::Ord for SectorId {
```
dignifiedquire: you should be able to derive `PartialOrd` and `Ord`

storage-proofs/src/sector_id.rs
```diff
@@ -0,0 +1,30 @@
+use byteorder::ByteOrder;
+
+#[derive(Debug, Clone, Copy, PartialOrd, PartialEq, Eq)]
+pub struct SectorId(u64);
```
dignifiedquire: this should be `#[repr(C)]` otherwise things will be painful in the ffi boundary

dignifiedquire: sorry, spoke too soon `repr(transparent)` isl likely what we want

dignifiedquire: https://github.com/rust-lang/rfcs/blob/master/text/1758-repr-transparent.md

laser: I would not expose `SectorId` through the FFI (and rust-fil-proofs doesn't know about FFIs).

Instead, the crates which do FFI things would [create their own FFI-compatible representation of `SectorId`](https://github.com/filecoin-project/rust-fil-sector-builder/blob/f5ac51b7739041d63a9345f0f0d13ab77762d91a/sector-builder-ffi/src/api.rs#L16).

filecoin-proofs/src/api.rs
laser: Word. I [made that change](https://github.com/filecoin-project/rust-fil-proofs/pull/840/files#diff-7ff052526c811cb90665a7388bfdcc1fR200) already.

filecoin-proofs/src/api.rs
laser: Your wish is my command!

storage-proofs/src/sector_id.rs
laser: Right you are! Updated.

filecoin-proofs/src/api.rs
laser: Just changed the field name back to `sector`, from `sector_id` as per your request.

PR #843, opened 2019-08-27
fix(storage-proofs): ensure unique topological ordering of base graph

### LINKED ISSUE #744: Base DAG should have a unique topological ordering
Add edges `(i, i+1)` for all `i < N-1` for the [`parents`](https://github.com/filecoin-project/rust-fil-proofs/blob/master/storage-proofs/src/drgraph.rs#L209) of the base `Graph`. Discount that from the base degree, so we'll now have only 4 parents from bucket sampling.

Not very clear in Algorithm 1 of the [original paper](https://acmccs.github.io/papers/p1001-alwenA.pdf) but we do _know_ it has to be so since its code is based on [Argon2i](https://password-hashing.net/submissions/specs/Argon-v3.pdf).

(edit: once resolved we'll also need to update the spec which at this time is just a copy of the code here)

-------------------------------------------

PR: https://github.com/filecoin-project/rust-fil-proofs/pull/843.

schomatis:
BREAKING CHANGE: adding one more parent to the base graph.

Option 1 of https://github.com/filecoin-project/rust-fil-proofs/issues/744#issuecomment-519643859.

Fixes #744.

### DIFF
--- a/storage-proofs/src/circuit/drgporep.rs
+++ b/storage-proofs/src/circuit/drgporep.rs
@@ -668,8 +668,8 @@ mod tests {
         }
 
         assert!(cs.is_satisfied(), "constraints not satisfied");
-        assert_eq!(cs.num_inputs(), 16, "wrong number of inputs");
-        assert_eq!(cs.num_constraints(), 103813, "wrong number of constraints");
+        assert_eq!(cs.num_inputs(), 18, "wrong number of inputs");
+        assert_eq!(cs.num_constraints(), 130957, "wrong number of constraints");
 
         assert_eq!(cs.get_input(0, "ONE"), Fr::one());
 
@@ -711,8 +711,8 @@ mod tests {
         )
         .expect("failed to synthesize circuit");
 
-        assert_eq!(cs.num_inputs(), 16, "wrong number of inputs");
-        assert_eq!(cs.num_constraints(), 305791, "wrong number of constraints");
+        assert_eq!(cs.num_inputs(), 18, "wrong number of inputs");
+        assert_eq!(cs.num_constraints(), 361789, "wrong number of constraints");
     }
 
     #[test]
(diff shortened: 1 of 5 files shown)


dignifiedquire:
@dignifiedquire needs to review this

dignifiedquire:
I don't think I agree with changing the functions to return a `-1`, especially now that the constants are merged, this makes it much harder to predict how things should work.

schomatis:
Ok, blocking for the moment then. Could you input your opinion on #744 please?

schomatis:
Thanks for fixing the PR @DrPeterVanNostrand, LGTM.

storage-proofs/src/circuit/drgporep.rs
```diff
@@ -566,7 +566,7 @@ mod tests {
         let sp = drgporep::SetupParams {
             drg: drgporep::DrgParams {
                 nodes,
-                degree,
+                degree: m,
```
dignifiedquire: Either this is `m` or `degree`, but not `degree: m` please.

storage-proofs/src/circuit/drgporep.rs
```diff
@@ -545,7 +545,7 @@ mod tests {
         let rng = &mut XorShiftRng::from_seed([0x3dbe6259, 0x8d313d76, 0x3237db17, 0xe5bc0654]);
 
         let nodes = 12;
-        let degree = BASE_DEGREE;
+        let m = BASE_DEGREE;
```
dignifiedquire: it should hold that `graph.degree() == BASE_DEGREE`, so the constants should be adjusted

storage-proofs/src/circuit/drgporep.rs
schomatis: Fair enough, I think Jake changed that because you didn't want the `degree - 1`.

dignifiedquire:
merged manually into https://github.com/filecoin-project/rust-fil-proofs/pull/864

PR #853, opened 2019-08-29
Update readme to include Pedersen Hash window-size optimization

DrPeterVanNostrand:
Add sections to readme for Pedersen Hash speed and memory optimizations.

### DIFF
--- a/README.md
+++ b/README.md
@@ -176,9 +176,30 @@ cargo run                                                                     \
     --size 1048576
 ```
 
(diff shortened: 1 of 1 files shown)


DrPeterVanNostrand:
From porcu's calculations, changing the Pedersen Hash window from 16 to 8 bits results in a 50 MiB decrease in memory usage at a cost of a 60% slowdown to Pedersen hashing. I'll add a blurb about changing the the window-size setting along with the expected memory-speed effects to the memory section in the readme.

schomatis:
@DrPeterVanNostrand Thanks for the update. I still feel we're missing some concrete numbers, we're talking about a "50 MiB reduction" and "9% increase in memory" but I'm not sure if it's clear to the user the final number, how much does this cost in terms for memory, to make the decision to turn it off or not. (I'm lacking some context here so maybe that number is already implicit in your description and I'm missing it, I'd just like to have it as obvious as possible to avoid any confusions.)

README.md
```diff
@@ -174,10 +174,35 @@ cargo run                                                                     \
...
+The following benchmarks were observed when running replication on 1MiB (1048576 bytes) of data on a new `m5a.2xlarge` EC2 instance with 32GB of RAM for Pedersen Hash window-sizes of 16 (the current default) and 8 bits:
+
+```
+window-size: 16
+User time (seconds): 87.82
+Maximum resident set size (kbytes): 1712320
+
+window-size: 8
+User time (seconds): 128.85
+Maximum resident set size (kbytes): 1061564
```
schomatis: This seems more than a 50 MiB reduction.

Issue #857, opened 2019-09-01
Increase expansion parents from 8 to 16


schomatis:
To conform with new analysis (see https://github.com/filecoin-project/rust-fil-proofs/issues/827). Although straightforward we should check any potential impacts (e.g., parents cache size, replication time, change in the circuits).

schomatis:
Ideally this should come after the changes in https://github.com/filecoin-project/rust-fil-proofs/pull/791 I think (not blocking, though).

porcuquine:
In keeping with notes on #827, I don't think we should change the value of the parameter. I think we should instead be generating more parents given the same parameter.

dignifiedquire:
>  I think we should instead be generating more parents given the same parameter.

In this case we would need a second parameter, or something else, that gives us the upper bound of parents, as that is critical to have, especially in light of #791

schomatis:
Will be fixed in https://github.com/filecoin-project/rust-fil-proofs/pull/864.

Issue #827, opened 2019-08-20
Fix expander parents generation


nicola:
### Description
See the work done in https://github.com/filecoin-project/research/issues/144, the newest spec for the chung construction lives in https://www.overleaf.com/3456316943pfxsqxkdghsy%E2%80%A9

schomatis:
Studying new analysis in [Stacked-DRGs & ZigZag Security parameters](https://www.overleaf.com/read/kcdhnxwptxbc), section 2.1, the conclusion for implementation fixes until now is:

1. Increase `expansion_degree` from 8 to 16.
2. Generate those extra 8 parents _not_ from the same permutation but form its _inverse_.

@irenegia @porcuquine Could either of you confirm point 2 please? Reading `Algorithm 1` I'm having trouble following the notation and it's not clear to me that this is the case (although I've been told as such, but would rather have a clear confirmation here please before moving forward).

porcuquine:
@schomatis Your general reading is correct, but let me make a few notes and help prioritize.

1. I think it might be better to leave `expansion_degree` as 8 and instead change the interpretation of that parameter. This is because making this change is a *correction*. The *average* expansion degree will now be 8. Since `d = 8` is the formal concept we're trying to replicate, we should probably just define `expansion_degree = d` and propagate the change that way for now.

There will probably be changes to how parents are handled in the future.

2. Yes, we need to generate as many parents from the inverse as from the permutation itself. In practice, this means we will use both the inverse and the permutation on every layer.

It's not clear to me that the PDF is updated to reflect this, btw. cc: @irenegia

porcuquine:
@schomatis 	Elsewhere (will try to comment there also), I think you mentioned wanting to defer this until after caching is moved to disk. I think it's important that we make the logic changes as soon as possible so that they don't block or complicate the many other changes which will be coming soon.

If we really want to avoid caches growing in the short term, you could change the parameters to be 'half-strength' (i.e. lower `expansion_degree` to 4) as a temporary measure.

But please prioritize this over restructuring the cache. It's very high priority that this be taken care of before we start in on new ZigZag work.

schomatis:
[...]

PR #859, opened 2019-09-02
fix(storage-proofs): increase expansion parents from 8 to 16

schomatis:
* What benchmarks should be run? (At least one with the cache turned on.)

* Is there another place where this value is hardcoded?

Towards #827.
Closes https://github.com/filecoin-project/rust-fil-proofs/issues/857.


schomatis:
Will be fixed in https://github.com/filecoin-project/rust-fil-proofs/pull/864.

PR #860, opened 2019-09-02
fix(storage-proofs): double expansion parents using permutation inverse

schomatis:
Closes https://github.com/filecoin-project/rust-fil-proofs/issues/827.
Closes https://github.com/filecoin-project/rust-fil-proofs/pull/859.
Closes https://github.com/filecoin-project/rust-fil-proofs/issues/857.

### DIFF
--- a/storage-proofs/src/circuit/zigzag.rs
+++ b/storage-proofs/src/circuit/zigzag.rs
@@ -460,8 +460,8 @@ mod tests {
 
         // End copied section.
 
-        let expected_inputs = 36;
-        let expected_constraints = 432312;
+        let expected_inputs = 52;
+        let expected_constraints = 673496;
         {
             // Verify that MetricCS returns the same metrics as TestConstraintSystem.
             let mut cs = MetricCS::<Bls12>::new();
--- a/storage-proofs/src/zigzag_graph.rs
+++ b/storage-proofs/src/zigzag_graph.rs
@@ -190,7 +190,7 @@ pub trait ZigZag: ::std::fmt::Debug + Clone + PartialEq + Eq {
     fn base_graph(&self) -> Self::BaseGraph;
     fn expansion_degree(&self) -> usize;
     fn reversed(&self) -> bool;
-    fn expanded_parents<F, T>(&self, node: usize, cb: F) -> T
+    fn expanded_parents<F, T>(&self, node: usize, invert_permutation_order: bool, cb: F) -> T
     where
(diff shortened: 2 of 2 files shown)


storage-proofs/src/zigzag_drgporep.rs
```diff
@@ -199,16 +199,14 @@ mod tests {
             // prove_verify_32_3_1(32, 3, 1);
             // prove_verify_32_3_2(32, 3, 2);
 
-           prove_verify_fixed_32_5_1(5, 1);
            prove_verify_fixed_32_5_2(5, 2);
-           prove_verify_fixed_32_5_3(5, 3);
+           prove_verify_fixed_32_5_4(5, 4);
```
schomatis: Changing these just to use an even expansion degree (don't know how the tests work or if this is fine to do).

dignifiedquire: that is fine

dignifiedquire:
@schomatis needs conflict resolution with `master`

Are you planning to write a test or not? (not sure how to interpret your :+1: on my comment)

schomatis:
I'm waiting on confirmation from @porcuquine first (https://github.com/filecoin-project/rust-fil-proofs/issues/827#issuecomment-527146852) if this is the right way to handle the parents increase. (:+1:: test sounds like a good idea, I need to think a bit how to implement it, but after confirmation)

porcuquine:
I commented on #827. I think it probably makes most sense to double the expansion degree rather than require it to be even. If we do this, the value of the expansion degree will not increase, but we will (correctly) generate more actual expansion parents for each node. See #827 for a bit more detail about how `degree` would be calculated in this model.

schomatis:
@dignifiedquire @porcuquine Did my best to interpret the corrections in https://github.com/filecoin-project/rust-fil-proofs/issues/827#issuecomment-528351096, but if this is still not the implementation required I'm going to suggest writing the specification _first_, and then I can implement whatever lands there.

schomatis:
Will be included in https://github.com/filecoin-project/rust-fil-proofs/pull/864.

dignifiedquire:
@schomatis can you please change this, such that the logic is all in expanded_parents/generate_expanded_parents, for the zigzag commitments work I need the parents() call to be only concatenating the base and expander parents

schomatis:
Closing in favor of https://github.com/filecoin-project/rust-fil-proofs/pull/888 which will be directly merged into #864.

Issue #847, opened 2019-08-27
Cache and stream drgprah


dignifiedquire:
General Idea:

How to cache the drgraph

- precalculate all parents:
  - regular: `[[u64; 13], NODES]`
  - reversed: `[[u64; 13], NODES]`
- store on disk: offsets into the data, eg `parent * NODE_SIZE`
  - regular: `[[u64; 13]; NODES]`
  - reversed: `[[u64; 13]; NODES]`

To read, memmap either `regular.cache` or `reversed.cache`, and read in chunks of `13` the required offsets.
This will be a continous read, and memmap + OS should be able to make this efficient for us, optimal for available memory.

- Estimated size: 3.25GiB for each cache file at a sector size of `1Gib`
- This is of course strongly dependent on the amount of base and expander parents. (assumed to be 5 + 8 = 13 above).

dignifiedquire:
When using sector sizes of up to 128GiB we can get away with storing `u32`s for the actual parents.

PR #861, opened 2019-09-03
refactor(storage-proofs): streamline expander parents cache

dignifiedquire:
Switches the cache structure to something towards what is described in #847. This still keeps everything in memory though. 
This also removes the cloning of the cache, when cloning the graph.

For 128MiB sectors the following memory usage changes can be seen 

**with `disk-trees`**

| state  | max res memory   |
|--------|------------------|
| master |  1_714_648 bytes |
| 02718b6 | 1_285_292 bytes |
| beeee94  | 1_285_232 bytes |

**without `disk-trees`**

| state  | max res memory   |
|--------|------------------|
| master  | 4_532_592 bytes |
| 02718b6 | 4_167_780 bytes |
| beeee94  | 4_068_068 bytes |

### DIFF
--- a/storage-proofs/src/zigzag_graph.rs
+++ b/storage-proofs/src/zigzag_graph.rs
@@ -1,3 +1,4 @@
+use std::collections::HashMap;
 use std::marker::PhantomData;
 use std::sync::{Arc, RwLock};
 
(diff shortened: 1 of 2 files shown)


storage-proofs/src/zigzag_graph.rs
```diff
@@ -11,18 +11,78 @@ use crate::settings;
 
 pub const DEFAULT_EXPANSION_DEGREE: usize = 8;
 
-// Cache of node's parents.
-pub type ParentCache = HashMap<usize, Vec<usize>>;
+lazy_static! {
+    // This parents cache is currently used for the *expanded parents only*, generated
+    // by the expensive Feistel operations in the ZigZag, it doesn't contain the
+    // "base" (in the `Graph` terminology) parents, which are cheaper to compute.
+    static ref PARENT_CACHE: Arc<RwLock<HashMap<String, ParentCache>>> = Arc::new(RwLock::new(HashMap::new()));
```
DrPeterVanNostrand: It may be beneficial to document how the hashmap is keyed, it's not obvious to me that the `String` key is the `ZigZagGraph`'s id.

storage-proofs/src/zigzag_graph.rs
```diff
@@ -359,31 +390,56 @@ where
...
-            return parents;
-        }
-
-        let parents: Vec<usize> = (0..self.expansion_degree)
-            .filter_map(|i| {
-                let other = self.correspondent(node, i);
-                if self.reversed {
-                    if other > node {
-                        Some(other)
+    fn expanded_parents<F, T>(&self, node: usize, mut cb: F) -> T
```
DrPeterVanNostrand: This method (and `Graph.parents()`) are pretty hard for me to read. I'm not saying change anything in this PR, but in the future we may want to document them a little bit more.

storage-proofs/src/zigzag_graph.rs
```diff
@@ -11,18 +11,78 @@ use crate::settings;
...
+        assert!(node < self.cache_entries);
+        cb(self.reverse[node as usize].as_ref())
+    }
+
+    pub fn write_forward(&mut self, node: u32, parents: Vec<u32>) {
+        assert!(node < self.cache_entries);
+
+        let old_value = std::mem::replace(&mut self.forward[node as usize], Some(parents));
+
+        debug_assert_eq!(old_value, None);
```
schomatis: nit: if this is the only reason we're adding an `Option` to the cache type I would just do a read before (or some other check to see if this entry is empty) in the debug case since we'd be calling this write function only once per node.

dignifiedquire: I think this structure is pretty good for now, we should eventually fill the cache in one go, and then we can remove the `Option` here.

storage-proofs/src/zigzag_graph.rs
dignifiedquire: improved the documentation

DrPeterVanNostrand: nice, thanks

PR #865, opened 2019-09-08
feat(fil-proofs-tooling): add rational-post subcommand to benchy

### LINKED ISSUE #846: add `rational-post` subcommand to `benchy` + capture benchmark on `master` builds
### BLOCKERS

Note that this story is blocked on Rational PoSt being merged into rust-fil-proofs master.

### Description

We should be able to use `benchy` to generate and verify a PoSt. To start, we should add a `benchy rational-post` command which accepts a single argument, `size` (the sector size, in KiB, like `benchy zigzag` accepts).

We should measure wall clock and CPU time for `generate_post` and `verify_post` as well as maximum resident set size (RAM).

DrPeterVanNostrand:
Adds `rational-post` subcommand to `benchy` for benchmarking `generate_post()` and `verify_post()`.

Closes #846

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -215,9 +215,15 @@ jobs:
             cat zigzag-benchmarks.json
           no_output_timeout: 60m
       - run:
+          name: Run Rational PoST benchmarks using a 1GiB sector
+          command: |
+            ./fil-proofs-tooling/scripts/benchy-remote.sh "${CIRCLE_BRANCH}" "${BENCHMARK_SERVER_SSH_USERNAME}@${BENCHMARK_SERVER_IP_ADDR}" rational-post --size=$((1024*1024)) > rational-post-benchmarks.json
+            cat rational-post-benchmarks.json
+          no_output_timeout: 60m
+      - run:
           name: Aggregate benchmarks into single JSON document
           command: |
-            ./fil-proofs-tooling/scripts/aggregate-benchmarks.sh zigzag-benchmarks.json micro-benchmarks.json hash-constraints.json > aggregated-benchmarks.json
+            ./fil-proofs-tooling/scripts/aggregate-benchmarks.sh zigzag-benchmarks.json micro-benchmarks.json hash-constraints.json rational-post-benchmarks.json > aggregated-benchmarks.json
             cat aggregated-benchmarks.json
       - store_artifacts:
           path: zigzag-benchmarks.json
--- a/fil-proofs-tooling/scripts/aggregate-benchmarks.sh
+++ b/fil-proofs-tooling/scripts/aggregate-benchmarks.sh
@@ -5,11 +5,13 @@ set -e
 zigzag_path=$1
 micro_path=$2
 hash_constraints_path=$3
+rational_post_path=$4
 
 jq --sort-keys -s '{ benchmarks: { "zigzag-benchmarks": { outputs: { "max-resident-set-size-kb": .[0] } } } } * .[1]' \
   <(jq '.["max-resident-set-size-kb"]' $zigzag_path) \
-  <(jq -s '.[0] * { benchmarks: { "hash-constraints": .[1], "zigzag-benchmarks": .[2], "micro-benchmarks": .[3] } }' \
+  <(jq -s '.[0] * { benchmarks: { "hash-constraints": .[1], "zigzag-benchmarks": .[2], "micro-benchmarks": .[3], "rational-post-benchmarks": .[4] } }' \
     <(jq 'del (.benchmarks)' $micro_path) \
     <(jq '.benchmarks' $hash_constraints_path) \
     <(jq '.benchmarks' $zigzag_path) \
-    <(jq '.benchmarks' $micro_path))
+    <(jq '.benchmarks' $micro_path) \
+    <(jq '.benchmarks' $rational_post_path))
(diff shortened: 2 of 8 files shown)


.circleci/config.yml
```diff
@@ -214,10 +214,16 @@ jobs:
             ./fil-proofs-tooling/scripts/benchy-remote.sh "${BENCHMARK_SERVER_SSH_USERNAME}@${BENCHMARK_SERVER_IP_ADDR}" zigzag --size=$((1024*1024)) > zigzag-benchmarks.json
             cat zigzag-benchmarks.json
           no_output_timeout: 60m
+      - run:
+          name: Run Rational PoST benchmarks using a 1MiB sector
+          command: |
+            ./fil-proofs-tooling/scripts/benchy-remote.sh "${BENCHMARK_SERVER_SSH_USERNAME}@${BENCHMARK_SERVER_IP_ADDR}" rational-post --size=1024 > rational-post-benchmarks.json
```
laser: QUESTION: If you changed `--size=1024` to `--size=$((1024*1024))`, does this thing still run successfully?

fil-proofs-tooling/scripts/aggregate-benchmarks.sh
```diff
@@ -5,11 +5,13 @@ set -e
 zigzag_path=$1
 micro_path=$2
 hash_constraints_path=$3
+rational_post_path=$4
 
 jq --sort-keys -s '{ benchmarks: { "zigzag-benchmarks": { outputs: { "max-resident-set-size-kb": .[0] } } } } * .[1]' \
   <(jq '.["max-resident-set-size-kb"]' $zigzag_path) \
-  <(jq -s '.[0] * { benchmarks: { "hash-constraints": .[1], "zigzag-benchmarks": .[2], "micro-benchmarks": .[3] } }' \
+  <(jq -s '.[0] * { benchmarks: { "hash-constraints": .[1], "zigzag-benchmarks": .[2], "micro-benchmarks": .[3], "rational-post-benchmarks": .[4] } }' \
```
laser: You are now a jq expert, congratulations!

fil-proofs-tooling/src/bin/benchy/rational_post.rs
```diff
@@ -0,0 +1,167 @@
...
+
+impl Report {
+    /// Print all results to stdout
+    pub fn print(&self) {
+        let wrapped = Metadata::wrap(&self).expect("failed to retrieve metadata");
+        serde_json::to_writer(stdout(), &wrapped).expect("cannot write report JSON to stdout");
+    }
+}
+
+struct FuncMeasurement<T> {
```
laser: BLOCKING: I see that `FuncMeasurement` is also defined in `benchy::zigzag`. Would you mind refactoring such that the definition can be shared between the two modules?

fil-proofs-tooling/src/bin/benchy/rational_post.rs
```diff
@@ -0,0 +1,167 @@
...
+pub fn run(sector_size: usize) -> Result<(), failure::Error> {
+    info!("Benchy Rational PoST: sector-size={}", sector_size);
+
+    let sector_size_unpadded_bytes_ammount =
+        UnpaddedBytesAmount::from(PaddedBytesAmount(sector_size as u64));
+
+    // Create a directory and setup some files, we will create one file for each: the staged sector,
+    // sealed sector, and the sealed sector's comm_r. `ss` in the filenames stands for
+    // "sector-size".
+    create_dir_all(BASE_DIR)?;
```
laser: BLOCKING: This implementation assumes that there will be only one Rational PoSt benchmark running on the same system at any point in time - and if of with-lock.sh is used as intended (and has no bugs), that assumption will be true. I don't think there's a good reason to encode that assumption in this function, though.

Fortunately, removing that assumption is as simple as replacing your staged_path (and sealed_path) with something like:

```
use tempfile::NamedTempFile;
let mut staged_file = NamedTempFile::new().expect("could not create named temp file");
let staged_path = staged_file.path();
```

This will help to ensure that each `run` call is atomic.

.circleci/config.yml
laser: QUESTION: What happens if you run this thing with `--size=((1024 * 1024))` instead of `--size=1024`? Does it run successfully?

fil-proofs-tooling/src/bin/benchy/rational_post.rs
```diff
@@ -0,0 +1,135 @@
...
+}
+
+pub fn run(sector_size: usize) -> Result<(), failure::Error> {
+    info!("Benchy Rational PoST: sector-size={}", sector_size);
+
+    let sector_size_unpadded_bytes_ammount =
+        UnpaddedBytesAmount::from(PaddedBytesAmount(sector_size as u64));
+
+    // Create files for the staged and sealed sectors.
+    let mut staged_file =
```
laser: Very nice. Thanks!

fil-proofs-tooling/src/bin/benchy/rational_post.rs
```diff
@@ -0,0 +1,135 @@
...
+    let mut priv_replica_info: BTreeMap<SectorId, PrivateReplicaInfo> = BTreeMap::new();
+
+    pub_replica_info.insert(sector_id, PublicReplicaInfo::new(seal_output.comm_r));
+
+    priv_replica_info.insert(
+        sector_id,
+        PrivateReplicaInfo::new(sealed_path_string, seal_output.comm_r),
+    );
+
+    // Measure PoST generation and verification.
```
laser: NITPICK: It should be `PoSt`, not `PoST`.

PR #888, opened 2019-09-20
encapsulate logic in expaned parents

schomatis:
New `parents()` method:

https://github.com/filecoin-project/rust-fil-proofs/blob/4a6f1944cfa9e77f18f1592a3e5c041fba7ec475/storage-proofs/src/zigzag_graph.rs#L215-L246

### DIFF
--- a/storage-proofs/src/zigzag_graph.rs
+++ b/storage-proofs/src/zigzag_graph.rs
@@ -190,7 +190,7 @@ pub trait ZigZag: ::std::fmt::Debug + Clone + PartialEq + Eq {
     fn base_graph(&self) -> Self::BaseGraph;
     fn expansion_degree(&self) -> usize;
     fn reversed(&self) -> bool;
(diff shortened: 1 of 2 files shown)


storage-proofs/src/zigzag_graph.rs
```diff
@@ -352,22 +345,29 @@ where
...
-            .collect()
+            }
+        }
+
+        // Add padding parents.
+        let pad_parent = if self.reversed() { self.size() - 1 } else { 0 } as u32;
+        expanded_parents.extend(vec![
+            pad_parent;
+            self.expansion_degree * 2 - expanded_parents.len()
+        ]);
```
dignifiedquire: you can use `.resize` to avoid allocating the temp vector

schomatis:
Will be merged into #864.

dignifiedquire:
Merged

Issue #887, opened 2019-09-19
Integrate and test bellman with GPU


dignifiedquire:
Now that https://github.com/filecoin-project/bellman/pull/4 is merged need to

- add a `gpu` feature to filecoin-proofs and storage-proofs
- publish a new version of bellperson
- enable tests on CI to use it once we have access to GPU instances 
- do full cycle tests
- once all things are working enable it by default

deltazxm:
Is it done?It seems to can't use GPU run bellman

dignifiedquire:
it is implemented but disabled by default at the moment, it needs activation using the feature flag `gpu` on `filecoin-proofs`

deltazxm:
I have already opened it,but the benchmark result is  slower than CPU.I can not believe it

dignifiedquire:
integration is done, bug fixing and testing is ongoing

PR #891, opened 2019-09-23
feat(deps): update and add gpu feature

dignifiedquire:
First part of #887


dignifiedquire:
Update: not working as expected, getting

```
GPU Multiexp is NOT supported!
GPU FFT is NOT supported!
```

at the moment

dignifiedquire:
While this doesn to fully enable gpu operations, it does fix a dependency issue on master, so merging it to fix that

PR #895, opened 2019-09-25
feat: switch to stacked drg

### LINKED ISSUE #904: Implement tapering for column commitments
Requires #895

dignifiedquire:
- move from zigzag to stacked drg
- implement column commitments for stacked drg

- [x] Needs new parameters generated before merging

Closes #904


storage-proofs/src/stacked/proof_scheme.rs
```diff
@@ -0,0 +1,128 @@
...
+
+        Ok(PublicParams::new(graph, sp.layer_challenges.clone()))
+    }
+
+    fn prove<'b>(
+        pub_params: &'b Self::PublicParams,
+        pub_inputs: &'b Self::PublicInputs,
+        priv_inputs: &'b Self::PrivateInputs,
+    ) -> Result<Self::Proof> {
+        let proofs = Self::prove_all_partitions(pub_params, pub_inputs, priv_inputs, 1)?;
```
porcuquine: Why always one partition? I'm probably not following the logic here — which may or may not be a sign that this is either wrong or too tricky for its own good.

storage-proofs/src/vde.rs
```diff
@@ -1,4 +1,5 @@
 use blake2s_simd::Params as Blake2s;
```
porcuquine: Trying to comment on the file name: Can we please finally rename this from `vde`, even if we don't kill all remaining sloths (Slothfather forgive me.)

storage-proofs/src/stacked/graph.rs
```diff
@@ -0,0 +1,579 @@
...
+        expanded_parents.sort();
+
+        expanded_parents
+    }
+
+    pub fn new_stacked(
+        nodes: usize,
+        base_degree: usize,
+        expansion_degree: usize,
+        layer: usize,
```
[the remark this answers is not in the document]
dignifiedquire: It is in tests. It is used to track if we have exp parents or not.

storage-proofs/src/stacked/hash.rs
```diff
@@ -0,0 +1,33 @@
...
+}
+
+/// Hash all elements in the given buffer
+pub fn hash1(data: impl AsRef<[u8]>) -> PedersenDomain {
+    let data = data.as_ref();
+
+    if data.len() > 32 {
+        pedersen_md_no_padding(&data).into()
+    } else {
+        pedersen(data).into()
```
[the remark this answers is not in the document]
dignifiedquire: Lets cross that bridge when actually implementing supersectors

storage-proofs/src/stacked/proof_scheme.rs
porcuquine: Okay, I am fresh now. I will look more closely.

porcuquine: I figured this out. It was a legacy hack to make `prove` match the trait and should only have been used with a single partition. I think it may have been jammed in on short notice for API reasons at one point. I'll submit a small PR making the issue more explicit, and you can propagate to this one.

storage-proofs/src/circuit/stacked/hash.rs
```diff
@@ -0,0 +1,147 @@
...
+    use crate::util::bytes_into_boolean_vec;
+
+    #[test]
+    fn test_hash2_circuit() {
+        let mut rng = XorShiftRng::from_seed([0x5dbe6259, 0x8d313d76, 0x3237db17, 0xe5bc0654]);
+        let window_size = settings::SETTINGS
+            .lock()
+            .unwrap()
+            .pedersen_hash_exp_window_size;
+        let params = &JubjubBls12::new_with_window_size(window_size);
```
DrPeterVanNostrand: We actually already create the Pedersen Hash params here: https://github.com/filecoin-project/rust-fil-proofs/blob/feat/stacked/storage-proofs/src/crypto/pedersen.rs#L11

I know that there are many places in the code where we are recreating this value even though we have it statically allocated. Just a heads up because the Pedersen Hash params take 1 or 2 seconds to generate.

dignifiedquire: yes, created https://github.com/filecoin-project/rust-fil-proofs/issues/906 to do this for the whole codebase

storage-proofs/src/vde.rs
dignifiedquire: haha, more changes :P

filecoin-proofs/src/api/mod.rs
```diff
@@ -156,6 +159,7 @@ pub fn seal<T: AsRef<Path>>(
     out_path: T,
     prover_id_in: &FrSafe,
     sector_id: SectorId,
+    ticket: Ticket,
```
laser: @dignifiedquire - Will `seal` accept any 32-element byte array? It'll take some time to write the software to get the ticket from the chain to the sector builder, so I'd like to start off by passing 32 NUL bytes to both `seal` and `verify_seal`.

dignifiedquire: yes, it just needs to be consistent

storage-proofs/src/circuit/stacked/proof.rs
```diff
@@ -0,0 +1,603 @@
...
+            comm_d,
+            comm_r_last,
+            comm_c,
+            ..
+        } = self;
+
+        let graph = &public_params.graph;
+        let params = &self.params;
+        // In most cases (the exception being during testing) we want to ensure that the base and
+        // expansion degrees are the optimal values.
```
porcuquine: By 'optimal', do you mean 'correct'?

storage-proofs/src/stacked/porep.rs
```diff
@@ -0,0 +1,55 @@
...
+        data: &'b [u8],
+    ) -> Result<Vec<u8>> {
+        let mut data = data.to_vec();
+
+        Self::extract_and_invert_transform_layers(
+            &pp.graph,
+            &pp.layer_challenges,
+            replica_id,
+            &mut data,
+        )?;
```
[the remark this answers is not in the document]
dignifiedquire: I don't quite understand this comment, that is effictively what we are doing.

storage-proofs/src/stacked/params.rs
```diff
@@ -0,0 +1,441 @@
...
+    #[serde(bound(
+        serialize = "MerkleProof<H>: Serialize",
+        deserialize = "MerkleProof<H>: Deserialize<'de>"
+    ))]
+    pub comm_d_proofs: MerkleProof<H>,
+    #[serde(bound(
+        serialize = "MerkleProof<H>: Serialize, ColumnProof<H>: Serialize",
+        deserialize = "MerkleProof<H>: Deserialize<'de>, ColumnProof<H>: Deserialize<'de>"
+    ))]
+    pub comm_r_last_proofs: (MerkleProof<H>, Vec<MerkleProof<H>>),
```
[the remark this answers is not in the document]
dignifiedquire: yes, you are right, removing it

storage-proofs/src/drgraph.rs
```diff
@@ -175,8 +215,11 @@ impl<H: Hasher> Graph<H> for BucketGraph<H> {
...
                 }
 
-                // Use the degree of the curren graph (`m`), as parents.len() might be bigger
-                // than that (that's the case for ZigZag Graph)
+                // Add the immediate predecessor as a parent to ensure unique topological ordering.
+                parents[m_prime] = node - 1;
+
+                // Use the degree of the current graph (`m`) as `parents.len()` might be bigger than
+                // that (that's the case for Stacked Graph).
                 parents[0..m].sort_unstable();
```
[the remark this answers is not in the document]
dignifiedquire: are you sure, I thought inside the individual parents we would still sort

storage-proofs/src/circuit/stacked/proof.rs
dignifiedquire: I did not write this

DrPeterVanNostrand: I wrote this comment. The constant base and expansion degrees were an optimization brought in from r2 (I thought) and were not values that were ensuring correctness.

storage-proofs/src/stacked/porep.rs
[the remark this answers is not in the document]
porcuquine: As far as I can tell, we don't have a simple function to just generate the key, which would be good to eventually have. (I know we can do it by using all zeroes as the data, but we don't then have a direct way combining that key with raw/replicated data to encode/decode it. Doesn't have to happen now — I'm just noting that we will want to have and expose that ability.

storage-proofs/src/drgraph.rs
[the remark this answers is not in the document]
porcuquine: We should take the immediate predecessor first, then the other DRG parents in the order generated plus any padding. Then the expander parents in order generated plus any padding. This is deterministic, so no need for wasteful sorting.

storage-proofs/src/circuit/stacked/proof.rs
porcuquine: Got it. Using global constants is itself the optimization. It's not that the specific values chosen are somehow more optimal. Rather, those will just be the ones required by the protocol and therefore 'correct' (anything else would be an incorrect implementation of the protocol). The check here is to make sure what we have hard-coded as an optimization agrees with what has been declared in the public parameters. In theory, we implement a range of possible proofs, varying by public parameters. In practice we are limited to performing only those which agree with our hard-coded constants.

storage-proofs/src/stacked/porep.rs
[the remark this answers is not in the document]
dignifiedquire: sounds good

Issue #886, opened 2019-09-18
Seal and Unseal should provide Progress reports


dignifiedquire:
Should return a `crossbeam::channel::Receiver<ProgressEvent>`, to which `ProgressEvent`s are sent.

dignifiedquire:
Maybe a callback based api would make sense too. 

```rust
let seal_output = seal(..., |progress: ProgressEvent| {
  println!("progress: {:?}", progress);
})?;

PR #896, opened 2019-09-26
[WIP] feat: implement progress callback for seal and replicate

dignifiedquire:
Uses a simple callback that is called whenever a new event happens. This currently only covers Seal and Replicate, but should be a good start to figure out the api from start to finish.

Ref #886

### DIFF
--- a/filecoin-proofs/src/api/mod.rs
+++ b/filecoin-proofs/src/api/mod.rs
@@ -147,26 +147,49 @@ fn generate_piece_specs_from_source(
     Ok(piece_specs)
 }
 
+/// The different progress events that happen during a seal operation.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum SealEvent {
+    Start,
+    Copy,
+    Pad,
+    ProofSetup,
+    Replicate(layered_drgporep::ReplicateEvent),
+    CreatePieceInclusionProofs,
+    FetchGrothParams,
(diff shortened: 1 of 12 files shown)


filecoin-proofs/src/api/mod.rs
```diff
@@ -147,26 +147,49 @@ fn generate_piece_specs_from_source(
...
 
+/// The different progress events that happen during a seal operation.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum SealEvent {
+    Start,
+    Copy,
+    Pad,
+    ProofSetup,
+    Replicate(layered_drgporep::ReplicateEvent),
+    CreatePieceInclusionProofs,
```
porcuquine: What's a `PieceInclusionProof`?

filecoin-proofs/src/api/mod.rs
```diff
@@ -147,26 +147,49 @@ fn generate_piece_specs_from_source(
...
+    FetchGrothParams,
+    GenerateSealProof,
+    VerifyPieceInclusionProofs,
```
porcuquine: Same question.

filecoin-proofs/src/api/mod.rs
dignifiedquire: Is this  a real question

dignifiedquire:
out of date, closing for now

PR #899, opened 2019-09-26
feat(storage-proofs): add updatable Pedersen hasher

### LINKED ISSUE #874: Implement updateable Pedersen Hasher
We need an updatable hasher for pedersen, that works like the one we have for Blake2s

```rust
let hasher = Pedersen::new();
hasher.update(data1);
hasher.update(data2);
// result should be the same as hashing (data1 || data2)  at once using the 
// existing MD construction
let result = hasher.finalize();
```

DrPeterVanNostrand:
This PR gives `PedersenFunction` the ability to be updatable/incremental. Here "updatable" means that the entirety of the Pedersen Hash preimage does not need to be available at the same time, instead the preimage can be continuously fed into the hash function in chunks of any size.

This PR is dependent on [this PR](https://github.com/filecoin-project/sapling-crypto/pull/7) in `sapling_crypto` going in.

Closes #874

### DIFF
--- a/storage-proofs/src/hasher/pedersen.rs
+++ b/storage-proofs/src/hasher/pedersen.rs
@@ -1,9 +1,13 @@
+use std::fmt::{self, Debug, Formatter};
 use std::hash::Hasher as StdHasher;
 
 use bellperson::{ConstraintSystem, SynthesisError};
+use bitvec::{BitVec, LittleEndian};
 use ff::{PrimeField, PrimeFieldRepr};
 use fil_sapling_crypto::circuit::{boolean, num, pedersen_hash as pedersen_hash_circuit};
-use fil_sapling_crypto::jubjub::JubjubEngine;
+use fil_sapling_crypto::jubjub::{
+    edwards, pedersen_hash_params_for_segments, JubjubEngine, PrimeOrder,
+};
 use fil_sapling_crypto::pedersen_hash::{pedersen_hash, Personalization};
 use merkletree::hash::{Algorithm as LightAlgorithm, Hashable};
 use merkletree::merkle::Element;
@@ -14,6 +18,11 @@ use crate::circuit::pedersen::pedersen_md_no_padding;
 use crate::crypto::{kdf, pedersen, sloth};
 use crate::error::{Error, Result};
 use crate::hasher::{Domain, HashFunction, Hasher};
+use crate::settings;
+
+const N_CHUNKS_PER_SEGMENT: usize = 63;
+const N_BITS_PER_CHUNK: usize = 3;
+const N_SEGMENT_BITS: usize = N_CHUNKS_PER_SEGMENT * N_BITS_PER_CHUNK;
 
 #[derive(Default, Copy, Clone, Debug, PartialEq, Eq)]
 pub struct PedersenHasher {}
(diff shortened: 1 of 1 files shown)


storage-proofs/src/hasher/pedersen.rs
```diff
@@ -220,14 +273,18 @@ impl HashFunction<PedersenDomain> for PedersenFunction {
...
     fn hash(&mut self) -> PedersenDomain {
-        self.0.into()
+        self.curr_hash.into_xy().0.into()
     }
 
+    // Resets any internal state associated with this hasher.
     #[inline]
     fn reset(&mut self) {
-        self.0 = Fr::from_repr(FrRepr::from(0)).expect("failed 0");
+        *self = PedersenFunction::default();
```
dignifiedquire: lets do this manually to avoid locking on the settings as reset is called a lot in the merkle treed hahsing

storage-proofs/src/hasher/pedersen.rs
```diff
@@ -249,6 +306,96 @@ impl LightAlgorithm<PedersenDomain> for PedersenFunction {
...
+
+        if n_bits_total < N_SEGMENT_BITS {
+            self.unhashed_bits.extend(input_bits);
+            return;
+        }
+
+        let n_bits_to_store = n_bits_total % N_SEGMENT_BITS;
+        let n_bits_to_hash = n_bits_total - n_bits_to_store;
+
+        let bits_to_hash = self
```
dignifiedquire: I had a manual implementation of this chaining before, as this is pretty slow in bitvec if I remember correctly :/

storage-proofs/src/hasher/pedersen.rs
```diff
@@ -432,4 +581,42 @@ mod tests {
 
         assert_eq!(val, val_back);
     }
+
+    // Asserts that hashing 118 bytes using `sapling`'s Pedersen Hash yields the same digest as
+    // hashing with `PedersenHasher`. 118 bytes is the largest buffer that `sapling`'s Pedersen Hash
+    // can handle before `panic`ing (`sapling` only generates 5 segments worth of generators, so it
+    // will run out of generators if hashing more than 118 bytes).
+    #[test]
```
porcuquine: It would be worth spelling out a little more clearly how the 118 is derived — so readers not intimately familiar with this pedersen hashing can follow/verify this logic.

Each segment contains 63 3-bit chunks, so 5 segments support 3 * 5 * 63 = 945 bits. So 118 bytes = 118 * 8 = 944 bits, making 118 the larger number of whole bytes that can be handled.

Although this is implicit, writing it out will be helpful for those coming up to speed in the future.

porcuquine:
Let's also make sure @arielgabizon reviews before merging this.

storage-proofs/src/hasher/pedersen.rs
DrPeterVanNostrand: Oh right, good point!

storage-proofs/src/hasher/pedersen.rs
dignifiedquire: see `NodeBits` below

storage-proofs/src/hasher/pedersen.rs
```diff
@@ -48,12 +57,48 @@ impl Hasher for PedersenHasher {
...
+    // The Pedersen Hash segment index of the next unhashed segment.
+    next_segment: u32,
+    // Stores any unhashed bits between calls to `self.update()`. `unhashed_bits`'s bit-length never
+    // exceeds `N_SEGMENT_BITS` (i.e. never contains more than one segment's worth of unhashed
+    // data).
+    unhashed_bits: BitVec<LittleEndian, u8>,
+    // Stores the current Pederen Hash. A Pedersen Hash is a summation of curve points; everytime
+    // `self.update()` is called with anouther segment's worth of data, an element of the prime
+    // order JubJub subgroup (an Edward's point) is added to this value.
+    curr_hash: edwards::Point<Bls12, PrimeOrder>,
```
dignifiedquire: We need to think about this. This is a very large element (4 Fr elements = 4 * 4 * 64bit = 128 byte). 

We will need to hold one of these in memory for each node in a sector when doing column commitments, so instead of having `NODES_COUNT * 32` we would now have `NODES_COUNT + 128` bytes in memory.

storage-proofs/src/hasher/pedersen.rs
```diff
@@ -48,12 +57,48 @@ impl Hasher for PedersenHasher {
...
-#[derive(Copy, Clone, Debug, PartialEq, Eq)]
-pub struct PedersenFunction(Fr);
+#[derive(Clone, PartialEq)]
+pub struct PedersenFunction {
+    // The Pedersen Hash segment index of the next unhashed segment.
+    next_segment: u32,
+    // Stores any unhashed bits between calls to `self.update()`. `unhashed_bits`'s bit-length never
+    // exceeds `N_SEGMENT_BITS` (i.e. never contains more than one segment's worth of unhashed
+    // data).
+    unhashed_bits: BitVec<LittleEndian, u8>,
```
dignifiedquire: Can we drop this, if we enforce that always 32 bytes will be "updated"?

storage-proofs/src/hasher/pedersen.rs
DrPeterVanNostrand: In your last line you mean `NODE_COUNT * 128` right?

Ok, I'm not sure how to write an updateable pedersen hash without storing an entire edwards point for each node. In order to update the digest you have to do an edward's group law which uses all 3 coordinates and the twist.

We could store compressed curve points (a single Fr), but the cost of decompression is extremely high.

storage-proofs/src/hasher/pedersen.rs
DrPeterVanNostrand: It could be dropped if updates always occurred in multiples of 189 bits. As currently written, `unhashed_bits` will never be larger than 188 bits long.

PR #900, opened 2019-09-27
scratch directory demo - for caching merkle tree and other miscellaneous assets

laser:
I didn't get as far with things as I would have liked, but I have a basic implementation sketched out (0). The idea is that `seal` would accept a path to a "scratch directory" (1) which we'd use to create a `ScratchDirectory` which satisfies the `ScratchArea` trait (2). We'd need to plumb that value down into the places where we create and load merkle trees. Application code uses `Key` to disambiguate different merkle trees being persisted and/or loaded (3).

The `ScratchDirectory#save_item` method delegates to `std::fs::rename`, which (in Darwin and Linux environments) is an atomic `mv` operation. The `ScratchDirectory#open_item` method returns a `NamedFile`, which is a tuple of file descriptor and absolute path to the file. File descriptors returned from `open_item` remain valid (until closed) even if the file they're associated with is renamed or deleted or replaced.

The `ScratchArea` produces and accepts things which can be turned into File because the merkle_light library works with `File`. @schomatis hasn't yet merged the PR (4) which will allow us to create a DiskStore from a File, but I suspect that the PR will land soon.

0: `https://github.com/filecoin-project/rust-fil-proofs/pull/900`
1: `https://github.com/filecoin-project/rust-fil-proofs/pull/900/files#diff-f938055765f58d7df2dd854ddd47f5f8R163`
2: `https://github.com/filecoin-project/rust-fil-proofs/pull/900/files#diff-31f406d4f0b3d6fc4b61567fbd91890bR9`
3: `https://github.com/filecoin-project/rust-fil-proofs/pull/900/files#diff-31f406d4f0b3d6fc4b61567fbd91890bR20`
4: `https://github.com/filecoin-project/merkle_light/pull/32`

### DIFF
--- a/fil-proofs-tooling/src/bin/benchy/rational_post.rs
+++ b/fil-proofs-tooling/src/bin/benchy/rational_post.rs
@@ -69,6 +69,8 @@ pub fn run(sector_size: usize) -> Result<(), failure::Error> {
         .expect("file name is not a UTF-8 string")
         .to_string();
 
+    let metadata_dir = tempfile::TempDir::new()?;
+
     // Generate the data from which we will create a replica, we will then prove the continued
     // storage of that replica using the PoSt.
     let data: Vec<u8> = (0..sector_size).map(|_| random()).collect();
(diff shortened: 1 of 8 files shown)


laser:
@dignifiedquire - May I close this PR? I think that it's served its purpose.

dignifiedquire:
close all the PRs

PR #901, opened 2019-09-27
Simplify `transform_and_replicate_layers`.

schomatis:
Easier to review [ignoring whitespaces](https://github.com/filecoin-project/rust-fil-proofs/pull/901/files?w=1).

Talking with @cryptonm today I realized this function is way more complex than it needs to be:

* The graph doesn't need to be involved in the MT construction.
* We don't need any elaborate channel communication, each thread should store its generated MT, they can share a centralized source without any serious penalty.

Hopefully this will smooth a bit the learning curve for @cryptonm to work on some of the MT integrations aspects of the code.

### DIFF
--- a/storage-proofs/src/layered_drgporep.rs
+++ b/storage-proofs/src/layered_drgporep.rs
@@ -352,6 +352,8 @@ pub trait Layers {
         Ok(())
     }
 
+    /// Take the user `data` and `encode` and `transform` it sequentially `layers`
+    /// times. For each layer (including the original `data`) create a `MerkleTree`.
     fn transform_and_replicate_layers(
         graph: &Self::Graph,
         layer_challenges: &LayerChallenges,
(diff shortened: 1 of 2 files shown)


storage-proofs/src/layered_drgporep.rs
```diff
@@ -363,110 +365,89 @@ pub trait Layers {
...
-
-        let mut sorted_trees = outer_rx.iter().collect::<Vec<_>>();
-        sorted_trees.sort_unstable_by_key(|k| k.0);
-
-        sorted_trees
+            Ok(())
+        })
+        .map_err(errf)??;
+        // FIXME: Replace `errf` here (used once), Rust needs to know the type of `e`.
+        // .map_err(|e| {
```
schomatis: RLS couldn't figure out the type of `e`, need help here.

storage-proofs/src/layered_drgporep.rs
```diff
@@ -363,110 +365,89 @@ pub trait Layers {
...
+            let err_string = format!("{:?}", e);
+            error!(
+                "MerkleTreeGenerationError: {} - {:?}",
+                &err_string,
+                failure::Backtrace::new()
+            );
+            Error::MerkleTreeGenerationError(err_string)
+        };
+
+        let merkle_trees = Arc::new(Mutex::new(vec![None; layers + 1]));
```
schomatis: Is there a nicer way to have a `Vector` with already all the positions available for an our of order write to it? Should we use references instead?

schomatis:
Already fixed in https://github.com/filecoin-project/rust-fil-proofs/blob/feat/stacked/storage-proofs/src/stacked/proof.rs#L328.

PR #908, opened 2019-10-10
fix(storage-proofs): encoding_proof: remove redundant storage

dignifiedquire:


storage-proofs/src/circuit/stacked/encoding_proof.rs
```diff
@@ -91,18 +69,38 @@ impl EncodingProof {
...
+
+        Ok(())
+    }
+
+    pub fn synthesize_decoded<CS: ConstraintSystem<Bls12>>(
+        self,
+        mut cs: CS,
+        params: &<Bls12 as JubjubEngine>::Params,
+        replica_id: &[Boolean],
+        exp_encoded_node: &num::AllocatedNum<Bls12>,
```
porcuquine: Name is confusing. `exp` sounds like it relates to `expander` in the context of SDR. I think this is now the only kind of 'encoding' that happens, but if you want to be more descriptive, I think we previously agreed that `add_encode` is what we would call this operation. (Assuming I am following.)

storage-proofs/src/circuit/stacked/params.rs
```diff
@@ -82,21 +74,44 @@ impl<H: Hasher> Proof<H> {
...
+                let encoded_node = num::AllocatedNum::alloc(
+                    cs.namespace(|| format!("enc_node_{}", layer)),
+                    || {
+                        raw.map(Into::into)
+                            .ok_or_else(|| SynthesisError::AssignmentMissing)
+                    },
+                )?;
+
+                proof.synthesize_key(
+                    cs.namespace(|| format!("encoding_proof_{}", layer)),
```
porcuquine: Please don't call this `encoding_proof_…` if it's just the KDF. It could be `labeling_proof_…`.

dignifiedquire: naming is hard <3

dignifiedquire: will do a renaming pass at a later point to get this nice

PR #912, opened 2019-10-11
feat(filecoin-proofs): api for deals on chain and interactive po…

### LINKED ISSUE #903: Remove PiPs and implement commP - commD consistency check

dignifiedquire:
- Piece Validation
- Interactive PoRep

Closes #903


filecoin-proofs/src/api/mod.rs
```diff
@@ -1,404 +1,120 @@
...
-    > = StackedCompound::setup(&compound_setup_params)?;
-
-    let public_inputs = stacked::PublicInputs::<<DefaultTreeHasher as Hasher>::Domain> {
-        replica_id,
-        tau: Some(Tau { comm_r, comm_d }),
-        seed: None,
-        k: None,
+    let reduce = |stack: &mut Vec<PieceInfo>| {
+        // println!("\n---- reduce ----\n");
+        while reduce_inner(stack) {}
```
porcuquine: This looks good. I don't think it yet handles subtree-alignment  padding, though.

In order to ensure that pieces are always subtree-aligned, the packing process may need to add a variable number of zero-padded bytes. These bytes don't correspond to actual pieces, so they won't appear in the list of `PieceInfo`s that are directly supplied by the chain when verifying.

The cleanest way to handle this is probably to walk the supplied `PieceInfo`s and insert these 'padding pieces' explicitly. It might make sense to precompute (or at least cache after first computation) the `CommP`s for the various sizes of padding which might be needed — since otherwise the cost has to be paid every time by all verifiers.

Once the shadow padding pieces have been inserted, a variation of the approach implemented here should work fine. While there is probably a way to compute this padding on the fly while reducing, I think it becomes complicated — so we should probably just directly apply the padding algorithm to the supplied `CommP`s. The variation required is that if the top (last) item on the stack is larger than the next, then it may need to be removed and held by pushing onto a second stack, while the remaining items are reduced until a long-enough piece has been produced. Then the new top of stack and the held item can be hashed as before. (The example below should exercise this.)

We should definitely add tests exercising this — ideally with randomized piece sizes inserted into a sector until sealing is triggered.

Here's a single example of pieces which I think the current implementation won't handle: A=1, B=4,C=2, D=8

These should be packed as follows:

[{(A0 00) (BB BB)} {(CC 00) (00 00)}] [{(DD DD) (DD DD)} {(00 00) (00 00)}]

This implies the following modified piece ordering (padding pieces labeled as P):

A(1), P(1), P(2), B(4), C(2), P(2), P(4), D(8), P(8)

Verification would proceed as follows:
- Compare lengths of P(8), D(8)
- Equal so Hash(D||P) to get DP(16)
- Compare lengths of DP(16) , P(4).
- DP is greater, so remove from stack S1 and push to hold on S2.
- Compare lengths of P(4), P(2).
- P(4) is greater, so remove from stack S1 and push to hold on S2.
- Compare lengths of C(2), P(2).
- Equal so Hash(C||P) to get CP(4).
- Compare lengths of CP(4), P(4) [top of S2]
- Equal, so Hash(CP(4)||P(4)) to get CPP(8) [and pop S2]
- Compare lengths of CPP(8), DP(16)
- DP is greater, so keep holding
- Compare lengths of CPP(8), B(4).
- CPP(8) is greater, so push on S2.
- Compare lengths of B(4), P(2).
- B(4) is greater, so push on S2.
- Compare lengths of P(2), P(1).
- P(2) is greater, so push on S2.
- Compare lengths of P(1), A(1).
- Equal, so Hash(A(1)||P(1)) to get AP(2).
- Compare lengths of AP(2), P(2) [top of S2]
- Equal so Hash(AP(2)||P(2)) to get APP(4) [and pop S2]
- Compare lengths of APP(4) and B(4) [top of S2]
- Equal so Hash(APP(4), B(4)) to get APPB(8) [and pop S2]
- Compare lengths of APPB(8) and CPP(8) [top of S2]
- Equal, so Hash (APPB(8)||CPP(8)) to get APPBCCPP(16) [and pop S2]
- Compare length of APPBCCPP(16) and DP(16)
- Equal, so Hash(APPBCCPP(16)||DP(16)) to get APPBCCPPDP(32) [and pop S2].
- Both stacks are empty, so APPBCCPPDP(32) is the result.
- Verify that the result equals CommD.

Note that APPBCCPPDP(32) is the expected result and has been constructed only by hashing equal-sized components.

dignifiedquire: @porcuquine I used the logic of your description to devise an algorithm, going forwad and only using one stack + a current hash.

porcuquine: I think what you did is equivalent. Using a 'stack' as a source of elements was just an artifact of going backward.

filecoin-proofs/src/pieces.rs
```diff
@@ -110,6 +110,64 @@ pub fn get_aligned_source(
...
+    let count = usize::from(size) / MINIMUM_PIECE_SIZE;
+
+    Ok(vec![
+        PieceInfo {
+            commitment: *SINGLE_PADDING_COMMITMENT,
+            size: UnpaddedBytesAmount(MINIMUM_PIECE_SIZE as u64)
+        };
+        count
+    ])
+}
```
porcuquine: This works 'for now', but it can lead to a lot of wasteful hashing to keep recomputing padding pieces which are powers of two multiples (other than one) of the `MINIMUM_PIECE_SIZE`. I think we are going to need to perform those optimizations eventually, because this is a potentially expensive action that all verifiers need to perform — so we need to keep the cost to the minimum.

The simplest way to do this is probably just to precompute and cache the doubling so we have a table with all the powers-of-two. At most this will be 32 bytes * the height of our largest sector, and it would be fine to compute this at startup.

Here is an example of the issue, which hopefully highlights why we want to be able to look up arbitrary power-of-2-sized padding pieces in a table.

Supposedthe supplied pieces have sizes (in order) 11, 256.

We will need the following padding pieces between the two: 16 , 4, 1, 32, 64, 128

If we instead just pad with unit padding and allow the pieces to be generated during the piece-verification reduction, we do a lot of extra hashing. In this example, that's 4 + 2 + 0 + 5 + 6 + 7 = 24 hashes.

filecoin-proofs/src/api/mod.rs
```diff
@@ -1,409 +1,167 @@
...
+    if piece_size > u64::from(sector_size) {
+        return Ok(false);
+    }
 
-            PseudoPieceSpec {
-                position: leaf_position,
-                number_of_leaves,
-                left_bytes: padded_left_bytes,
-                rest_bytes: padded_piece_length + padded_right_bytes,
+    // Holds intermediary pieces
```
porcuquine: 'intermediate'

filecoin-proofs/src/api/mod.rs
```diff
@@ -1,409 +1,167 @@
...
+                    //     "S2 <- H(s({}) | {} ({}))",
+                    //     u64::from(stack[stack.len() - 1].size) / 127,
+                    //     i,
+                    //     u64::from(piece.size) / 127
+                    // );
+                    {
+                        let l = stack.len();
+                        let el = &mut stack[l - 1];
+                        el.size = UnpaddedBytesAmount(u64::from(el.size) + u64::from(piece.size));
+                        let h = piece_hash(&el.commitment, &piece.commitment);
```
porcuquine: I don't think it's guaranteed that the piece being processed will be the same size as the one on the top of the stack. If they aren't the same size, then we cannot hash them like this.

One way to guard against this risk would be to modify `piece-hash` to take `PieceInfo`s as input and refuse to hash mismatched pieces.

I think your algorithm will still do the right thing if you just remove this hashing entirely, and leave it to be performed by the reduction if appropriate.

filecoin-proofs/src/api/mod.rs
porcuquine: Actually, I'm not so sure. Ignore my comments here and wait for me to submit review. Sorry for the noise.

filecoin-proofs/src/api/mod.rs
```diff
@@ -1,409 +1,167 @@
...
+                    }
+                } else {
+                    // println!("S2 <- {}({})", i, u64::from(piece.size) / 127);
+                    stack.push(piece.clone());
+                }
+            }
+            Ordering::Greater => {
+                unreachable!("algorithm failed");
+            }
+        }
```
porcuquine: I think this still isn't right. I've listed one example of how above. I think it's easiest just to describe what I think the right algorithm is. I will also try to write this into the spec (in code) for the sake of clarity and lack of ambiguity. I think the quirks of trying to write this 'efficiently' and in Rust aren't helping us.

I think once we assume padding pieces have been added *correctly*, the algorithm i simply this:

- For each successive piece, push it onto the output stack and fully reduce the stack.
- The stack can be reduced whenever the top two pieces have the same size.
- Reduction involves popping both pieces off the stack and pushing the result of hashing them.
- 'Fully reducing' means examining the stack after each reduction and performing another reduction if necessary.

This is almost exactly the algorithm I implemented in my experimental 'streaming merkle trees' PR: https://github.com/filecoin-project/merkle_light/pull/26/files#diff-be3695d843cbc8410e25ba4e24f978a5R77-R143

The main difference here is that instead of always adding `Roots` with height 0, (corresponding to a piece of size 1 node), we can push roots representing subtrees with different heights onto the stack. It's still the case that a root with greater height than the top of the stack can never be pushed, but this requires that piece padding was added correctly.

If it were ever the case that the next piece to be added had a root height *greater* than that on top of the stack, it would mean that piece padding should have been added.

This observation suggests what might be the cleanest algorithm: instead of adding piece padding in its own step, instead apply the reduction algorithm (as described above). Whenever shifting the next piece would produce an illegal state (one violating the invariant that roots heights on the stack never increase — think 'towers of Hanoi'), instead push a commitment to padding of the same size as the piece whose root is on top of the stack.

This will allow adding exactly the right padding, in one pass, without dependencies on potentially more complicated ways of calculating it, and in a way which makes clear *why* the chosen padding piece size is correct.

I am going to prioritize putting this in the spec as the fastest way to clarify any remaining questions either of us might have.

porcuquine: By the way, I think what you have now is probably close to this (ignoring the padding part). I wrote it out prescriptively since that was easier and less error-prone than trying to identify the (perhaps small) differences.

porcuquine: I wrote this into the spec: https://github.com/filecoin-project/specs/pull/586

My intention was to illustrate what I described above as straightforwardly as possible. It's late and spec lacks tests, so it's possible something isn't quite right. But this should get you most of the way there.

porcuquine: https://github.com/filecoin-project/specs/blob/b14a003a9bc144a77e7068f8cc018bc3ec72e278/src/libraries/filcrypto/filproofs/algorithms.go#L419-L482

porcuquine: https://github.com/filecoin-project/specs/blob/b14a003a9bc144a77e7068f8cc018bc3ec72e278/src/libraries/filcrypto/filproofs/algorithms.go#L419-L482

porcuquine: https://github.com/filecoin-project/specs/blob/de114990b925351bb86b6406c37e84ac1dceffcd/src/libraries/filcrypto/filproofs/algorithms.go#L419-L482

filecoin-proofs/src/api/mod.rs
porcuquine: Looking at this again later, I think misread the brackets above. I think you do have that guarantee.

filecoin-proofs/src/api/mod.rs
dignifiedquire: all finally fixed

filecoin-proofs/src/api/seal.rs
```diff
@@ -0,0 +1,278 @@
...
+        comm_r,
+        comm_d,
+        prover_id,
+        sector_id,
+        ticket,
+        seed,
+        &buf,
+        piece_infos,
+    )
+    .expect("post-seal verification sanity check failed");
```
laser: BLOCKING: This does not actually assert that the proof is valid. You need to check that the value of this expression equals `true`.

filecoin-proofs/src/api/mod.rs
```diff
@@ -449,140 +77,78 @@ pub fn get_unsealed_range<T: Into<PathBuf> + AsRef<Path>>(
...
+{
+    ensure!(
+        piece_size >= UnpaddedBytesAmount(MINIMUM_PIECE_SIZE),
+        "Piece must be at least {} bytes",
+        MINIMUM_PIECE_SIZE
+    );
+    let padded_piece_size: PaddedBytesAmount = piece_size.into();
+    ensure!(
+        u64::from(padded_piece_size).is_power_of_two(),
+        "Aligned and padded piece size must be a power of 2 ({:?})",
```
laser: BLOCKING: This error message doesn't make much sense to me, and tripped me up when I was integrating with the sector builder. 

The caller doesn't control adding alignment bytes (this function adds them).

Perhaps this error message should be:

```
Bit-padded piece size must be a power of 2 ({:?})
```

filecoin-proofs/src/api/seal.rs
porcuquine: ⭐️

filecoin-proofs/src/api/mod.rs
dignifiedquire: fixed

filecoin-proofs/src/api/seal.rs
dignifiedquire: fixed

PR #916, opened 2019-10-15
feat: enable gpu snarks by default

dignifiedquire:
They work well enough that it makes sense to run with them when available

@laser for awareness


fil-proofs-tooling/Cargo.toml
```diff
@@ -42,5 +42,5 @@ log = "0.4.8"
 uom = "0.25.0"
 
 [features]
-default = []
+default = ["gpu"]
```
cryptonemo: By making this the default, does it mean a GPU is required? Or that it'll be used by default if detected?

dignifiedquire: it means it well run the detection and use it if it finds a compatible one

dignifiedquire:
Blocked by https://github.com/cogciprocate/ocl/pull/163

dignifiedquire:
> Blocked by [cogciprocate/ocl#163](https://github.com/cogciprocate/ocl/pull/163)

actually it is not, seems some other issue

PR #919, opened 2019-10-18
feat(filecoin-proofs): generate_post and seal both accept cache dir-path

laser:
## Why does this PR exist?

The `seal` function creates a bunch of merkle trees which `generate_post` needs to make use of (so that it doesn't have to regenerate said merkle trees).

## What's in this PR?

This PR modifies both `seal` and `generate_post` to accept a per-sector cache path. The `seal` function will ultimately write to this directory and `generate_post` will read from it.

### DIFF
--- a/filecoin-proofs/src/api/post.rs
+++ b/filecoin-proofs/src/api/post.rs
@@ -19,6 +19,7 @@ use crate::error;
 use crate::parameters::{post_setup_params, public_params};
 use crate::singletons::ENGINE_PARAMS;
 use crate::types::{PaddedBytesAmount, PoStConfig};
+use std::path::PathBuf;
 
 /// The minimal information required about a replica, in order to be able to generate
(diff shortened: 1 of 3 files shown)


filecoin-proofs/src/api/post.rs
```diff
@@ -32,6 +33,8 @@ pub struct PrivateReplicaInfo {
     aux: PersistentAux,
     /// Is this sector marked as a fault?
     is_fault: bool,
+    /// Contains sector-specific (e.g. merkle trees) assets
+    cache_dir: PathBuf,
```
dignifiedquire: do we need this?

laser: @dignifiedquire - Don't you want to re-use the sector-specific merkle trees generated during `seal` when generating a PoSt?

PR #925, opened 2019-10-25
feat(storage-proofs): add updateable and batched pedersen hash

### LINKED ISSUE #874: Implement updateable Pedersen Hasher

DrPeterVanNostrand:
In this PR:
- adds an updateable Pedersen Hash function: `hasher::batched_pedersen::State`
- adds an batched updateable Pedersen Hash function: `hasher::batched_pedersen::BatchedPedersenFunction`

Todo:
- store the exp-table behind a statically allocated `Arc<Barrier<...>>` so we load the exp-table from disk once for all 8 Pedersen Hash threads
- merge `BatchedPedersenFunction` and `State` into `PedersenFunction`

This PR is dependent upon [this PR](https://github.com/filecoin-project/sapling-crypto/pull/8) in `sapling_crypto` being accepted, which allows for the Pedersen Hash exp-table to be read from and written to a file.

[Previous PR](https://github.com/filecoin-project/rust-fil-proofs/pull/899)

Closes #874


dignifiedquire:
This looks good to me in general, but lets make sure that the baseline hashing does not regress in speed.

dignifiedquire:
gonna close this for now, that we have moved to poseidon

PR #937, opened 2019-11-04
feat(storage-proofs): revive drgporep

dignifiedquire:


storage-proofs/src/circuit/constraint.rs
```diff
@@ -95,4 +142,24 @@ mod tests {
...
+
+            let res = sub(cs.namespace(|| "a-b"), &a, &b).expect("subtraction failed");
+
+            let mut tmp = a.get_value().unwrap().clone();
+            tmp.sub_assign(&b.get_value().unwrap());
+
+            assert_eq!(res.get_value().unwrap(), tmp);
+            assert!(cs.is_satisfied());
+        }
+    }
```
porcuquine: We probably didn't need to remove these general-purpose constraints previously.

PR #939, opened 2019-11-07
refactor: switch hashing from blake2s to sha256 by default

dignifiedquire:


storage-proofs/src/circuit/drgporep.rs
```diff
@@ -712,6 +722,10 @@ mod tests {
     }
 
     fn drgporep_test_compound<H: Hasher>() {
+        // femme::pretty::Logger::new()
```
DrPeterVanNostrand: Stray comment?

porcuquine:
@dignifiedquire Is there any reason not to switch to sha256 for PoSt challenge derivation too?

dignifiedquire:
> @dignifiedquire Is there any reason not to switch to sha256 for PoSt challenge derivation too?

There is not, I just wasn't sure if it made sense given the upcoming changes to it.

PR #940, opened 2019-11-07
feat: add merkle tree and encoding persistence

cryptonemo:


filecoin-proofs/src/api/seal.rs
```diff
@@ -81,8 +82,18 @@ pub fn seal_pre_commit<R: AsRef<Path>, T: AsRef<Path>, S: AsRef<Path>>(
         _,
     >>::setup(&compound_setup_params)?;
 
+    // MT for original data is always named tree-d, and it will be
+    // referenced later in the process as such.
+    let cache_dir = cache_path.as_ref().to_str().unwrap();
+    let config = StoreConfig::new(
+        cache_dir.to_string(),
```
dignifiedquire: this should really take a `T: AsRef<Path>`, allowing us to just pass the argument through

storage-proofs/src/stacked/params.rs
```diff
@@ -324,16 +324,67 @@ pub struct PersistentAux<D> {
...
+    /// The encoded nodes for 1..layers.
+    pub encodings: EncodingsCache<H>,
     pub tree_d: Tree<G>,
     pub tree_r_last: Tree<H>,
     pub tree_c: Tree<H>,
 }
 
-impl<H: Hasher, G: Hasher> TemporaryAux<H, G> {
+impl<H: Hasher, G: Hasher> TemporaryAuxCache<H, G> {
+    pub fn new(t_aux: &TemporaryAux<H, G>) -> Self {
```
dignifiedquire: this should return a `Result<Self>`, given that these operations can fail

storage-proofs/src/stacked/proof.rs
```diff
@@ -408,24 +432,38 @@ impl<'a, H: 'static + Hasher, G: 'static + Hasher> StackedDrg<'a, H, G> {
...
-                .into_par_iter()
-                .map(|i| get_node::<K>(tree_data, i).unwrap()),
-        )
+        if config.is_some() {
+            info!("building tree (size: {}) with id {}", tree_data.len(), config.clone().unwrap().id);
+            MerkleTree::from_par_iter_with_config(
+                (0..leafs)
+                    .into_par_iter()
+                    .map(|i| get_node::<K>(tree_data, i).unwrap()),
+                config.unwrap()
```
dignifiedquire: you can write 

```rust
if let Some(config) = config {
  
} else {
}
```

avoiding the `unwrap`

storage-proofs/src/stacked/proof.rs
```diff
@@ -435,31 +473,51 @@ impl<'a, H: 'static + Hasher, G: 'static + Hasher> StackedDrg<'a, H, G> {
...
-        let (encodings, column_hashes, tree_d) = crossbeam::thread::scope(|s| {
+        // Generate all store configs that we need based on the
+        // cache_path in the specified config.
+        assert!(config.is_some()); // FIXME: make not an Option
+        let mut tree_d_config = config.clone().unwrap();
+        let mut tree_r_last_config = config.clone().unwrap();
+        let mut tree_c_config = config.clone().unwrap();
+
+        tree_c_config.id = format!("tree-c-{:?}", replica_id);
+        tree_r_last_config.id = format!("tree-r-last-{:?}", replica_id);
```
dignifiedquire: I think to make these operation a bit clearer, there should be a another method like this

```rust
let tree_c_config = Config::from_config(&config, format!("tree-c-{:?}", replica_id));
```

instead of the manual clone and id replacement

filecoin-proofs/src/api/seal.rs
cryptonemo: I like the idea here, but might need some pointers on how to do this properly.  Everything broke in merkle_light regarding the configs with my first attempt ... but will definitely keep in mind as a needed clean-up.

dignifiedquire: lets connect on Monday when you are up

dignifiedquire:
- fixing rustfmt: `cargo fmt --all` in the project root
- fixing clippy: run `cargo clippy --all` and follow what the errors tell you, they always have links where you can read about the details

dignifiedquire:
Regarding the tests, that seems suspicious as they just hang at one point, Are all tests passing locally for you?

cryptonemo:
> Regarding the tests, that seems suspicious as they just hang at one point, Are all tests passing locally for you?

Yes, they do all pass locally.

cryptonemo:
> > Regarding the tests, that seems suspicious as they just hang at one point, Are all tests passing locally for you?
> 
> Yes, they do all pass locally.

Nevermind that one, I forgot to put back in the ignore for the seal test ...

storage-proofs/src/stacked/params.rs
```diff
@@ -340,6 +358,30 @@ impl<H: Hasher, G: Hasher> TemporaryAux<H, G> {
...
+    pub fn delete(t_aux: TemporaryAux<H, G>) -> Result<()> {
+        let tree_d_size = t_aux.tree_d_config.size.unwrap();
+        let tree_d_store: DiskStore<G::Domain> =
+            DiskStore::new_from_disk(tree_d_size, &t_aux.tree_d_config)?;
+        let tree_d: Tree<G> =
+            MerkleTree::from_data_store(tree_d_store, get_merkle_tree_leafs(tree_d_size));
+        tree_d.delete(t_aux.tree_d_config)?;
+
+        let tree_c_size = t_aux.tree_c_config.size.unwrap();
+        #[cfg(not(feature = "mem-trees"))]
```
dignifiedquire: double `cfg`, this is already on the function definition

filecoin-proofs/src/api/post.rs
```diff
@@ -88,13 +89,26 @@ impl PrivateReplicaInfo {
...
     /// Generate the merkle tree of this particular replica.
-    pub fn merkle_tree(&self, sector_size: u64) -> Result<Tree, Error> {
-        let mut f_in = File::open(&self.access)?;
-        let mut data = Vec::new();
-        f_in.read_to_end(&mut data)?;
-
-        let bytes = PaddedBytesAmount(sector_size as u64);
-        public_params(bytes, 1).graph.merkle_tree(&data)
+    pub fn merkle_tree(&self, _sector_size: u64) -> Result<Tree, Error> {
+        let tree_size = {
```
dignifiedquire: We can just compute the size of the tree based on the `sector_size` argument, no need to open the file.

storage-proofs/src/stacked/params.rs
```diff
@@ -19,6 +22,23 @@ use crate::util::data_at_node;
 
 pub type Tree<H> = MerkleTree<<H as Hasher>::Domain, <H as Hasher>::Function>;
 
+#[derive(Debug)]
```
dignifiedquire: lets derive `Copy` and `Clone` as well, makes it a bit easier to work with.

storage-proofs/src/stacked/proof.rs
```diff
@@ -386,7 +402,26 @@ impl<'a, H: 'static + Hasher, G: 'static + Hasher> StackedDrg<'a, H, G> {
             }
 
             // Write the result to disk to avoid keeping it in memory all the time.
-            labels.push(DiskStore::new_from_slice(layer_size, &layer_labels)?);
+            let layer_config = StoreConfig::from_config(
+                &config,
+                format!("replica-{:?}-layer-{}", replica_id, layer),
```
dignifiedquire: no need to use `replica_id` here either, should be part of the enum as well, sth like `CacheKey::LabelLayer(usize)`

storage-proofs/src/stacked/proof.rs
```diff
@@ -386,7 +402,26 @@ impl<'a, H: 'static + Hasher, G: 'static + Hasher> StackedDrg<'a, H, G> {
...
+                Some(layer_size),
+            );
+
+            // Construct and persist the layer data.
+            let _: DiskStore<H::Domain> = DiskStore::new_from_slice_with_config(
```
dignifiedquire: this method should just return a `LabelCache` so you don't need to recreate the stores when using it

fil-proofs-tooling/src/bin/benchy/stacked.rs
```diff
@@ -145,8 +156,13 @@ where
...
                 return_value: (pub_inputs, priv_inputs),
             } = measure(|| {
-                let (tau, (p_aux, t_aux)) =
-                    StackedDrg::<H, Sha256Hasher>::replicate(&pp, &replica_id, &mut data, None)?;
+                let (tau, (p_aux, t_aux)) = StackedDrg::<H, Sha256Hasher>::replicate(
+                    &pp,
+                    &replica_id,
+                    &mut data,
+                    None,
+                    Some(config.clone()),
```
dignifiedquire: given that the `StoreConfig` is always the same that gets passed in, how about we just pass the `cache_path` and construct the `StoreConfig` in the methods?

filecoin-proofs/src/lib.rs
```diff
@@ -7,6 +7,8 @@ extern crate failure;
 #[macro_use]
 extern crate log;
 
+extern crate bincode;
```
dignifiedquire: no need for this

filecoin-proofs/src/api/seal.rs
```diff
@@ -149,18 +155,29 @@ pub fn seal_commit<T: AsRef<Path>>(
...
 
+    let p_aux = {
+        let mut p_aux_bytes = vec![];
+        let mut f_p_aux = File::open(cache_path.as_ref().join(CacheKey::PAux.to_string()))?;
+        f_p_aux.read_to_end(&mut p_aux_bytes)?;
+
+        deserialize(&p_aux_bytes)
+    }?;
+
+    let t_aux = {
```
dignifiedquire: `t_aux` needs to be deleted after `commit` was successfull

fil-proofs-tooling/src/bin/benchy/stacked.rs
dignifiedquire: @cryptonemo did you forget about this one, or just not yet done?

dignifiedquire: Note: I think this applies to all methods, `replicate`, `extract_all`, etc. Can just pass the cache_path through and construct the configs in the places needed.

cryptonemo: We need the size from the config in some places, so unless we can always get that back, I'm not sure.  Otherwise, sounds fine.

filecoin-proofs/src/lib.rs
cryptonemo: Hrm, ok, the serialize/deserialize didn't want to work without it.  Maybe it's not being done in the best way?

filecoin-proofs/src/api/seal.rs
cryptonemo: We delete a little further down, no?

fil-proofs-tooling/src/bin/benchy/stacked.rs
dignifiedquire: fair enough, lets just leave it for now

storage-proofs/src/stacked/params.rs
```diff
@@ -361,6 +479,71 @@ impl<H: Hasher> Labels<H> {
...
+            "Layer {} is not available (only {} layers available)",
+            layer,
+            self.layers()
+        );
+
+        let row_index = layer - 1;
+        let config = self.labels[row_index].clone();
+        assert!(config.size.is_some());
+
+        DiskStore::new_from_disk(config.size.unwrap(), &config).unwrap()
```
laser: This expression will panic inside of `seal_commit` if the merkle tree was deleted from the cache directory after `seal_pre_commit` returned.

@dignifiedquire - If we really really want to get the merkle cache in-use by lotus ASAP, then we could merge this PR as-is and create follow-up stories to deal with panics. Or, we could fix it now. Deferring to your judgement.

---

To repro, check out branch `mt-persistence-panic` and run the following:

```
cargo test --color=always --package filecoin-proofs --lib api::tests::test_seal_lifecycle_layer_missing -- --nocapture --exact --ignored
```

storage-proofs/src/stacked/params.rs
```diff
@@ -361,6 +479,71 @@ impl<H: Hasher> Labels<H> {
...
+    }
+
+    /// Build the column for the given node.
+    pub fn column(&self, node: u32) -> Result<Column<H>> {
+        let rows = self
+            .labels
+            .iter()
+            .map(|label| {
+                assert!(label.size.is_some());
+                let store = DiskStore::new_from_disk(label.size.unwrap(), &label).unwrap();
```
laser: SUGGESTION: Given that you're returning a `Result<Column<H>>`, I'd suspect that it's not rocket science to remove the `unwrap` and instead short-circuit the map if `DiskStore::new_from_disk` produces an error.

storage-proofs/src/stacked/proof.rs
```diff
@@ -177,8 +185,9 @@ impl<'a, H: 'static + Hasher, G: 'static + Hasher> StackedDrg<'a, H, G> {
                                 let labeled_node = rpc.c_x.get_node_at_layer(layer);
                                 assert!(
                                     proof.verify(&pub_inputs.replica_id, &labeled_node),
-                                    "Invalid encoding proof generated"
+                                    format!("Invalid encoding proof generated at layer {}", layer)
```
laser: This will cause `seal_commit` to panic if one of the layer files was corrupted (e.g. its contents were replaced with the string `erin totally rules`) after `seal_pre_commit` returned.

@dignifiedquire - Same question as above - how does dealing with these error cases prioritize?

----

To repro the panic, check out branch `mt-persistence-panic` and run:

```
cargo test --color=always --release --package filecoin-proofs --lib api::tests::test_seal_lifecycle_layer_corruption -- --nocapture --exact --ignored
```

filecoin-proofs/src/api/seal.rs
laser: Yep, I see the delete:

```
TemporaryAux::<DefaultTreeHasher, DefaultPieceHasher>::delete(t_aux)?;
```

dignifiedquire:
The panics are not great, but it will be easier to address them afterwards, given the many changes that are coming in #949

filecoin-proofs/src/api/seal.rs
dignifiedquire: I am just blind :)

filecoin-proofs/src/lib.rs
dignifiedquire: strange

PR #942, opened 2019-11-08
refactor: use fil_logger instead of pretty_env_logger

vmx:
The log output changes and now looks like the one [IPFS go-log] is
producing. You can also toggle the output to JSON with setting
the environment variable `GOLOG_LOG_FMT=json`.

[IPFS go-log]: https://github.com/ipfs/go-log

I've created a new crate called `fil_logger`. Please have a look at its source as well (https://github.com/vmx/rust-fil-logger). It's just a wrapper around [flexi_logger](https://crates.io/crates/flexi_logger).

I decided to create it's own crate as we want to have a custom log format. It would be tedious to either always copy those functions, or having a crate with just them. The advantage of having the whole logger in a crate is, that it's now just a simple `fil_logger::init()` call and that we can change the underlying logger easily (which we are now doing for the 3rd time).

If you agree that having the `fil_logger` crate is a good idea, I'll properly move it to the filecoin org and publish it on crates.io (I would need help for that as I currently don't have a clue how adding a new repo and crate is managed in regards to permissions).

`fil_logger` already contains the functionality that is needed in order to make https://github.com/filecoin-project/rust-fil-sector-builder/issues/34 happen.

PS: I made in a draft PR until the `fil_logger` crate is published.


dignifiedquire:
can you paste some example output. I remember not being a fan of go-log output in the past.

vmx:
Please see the commit message for example output: https://github.com/filecoin-project/rust-fil-proofs/commit/cc1acb37d1b2f73e5588567545560f810b6e1b64

vmx:
I'll rebase when https://github.com/filecoin-project/rust-fil-proofs/pull/956/ is merged.

vmx:
Rebased and ready for review!

PR #946, opened 2019-11-15
feat(storage-proofs): add parallel pedersen hash

DrPeterVanNostrand:
### What is the "Parallel Pedersen Hash"?

The Parallel Pedersen Hash allows for the main thread of `rust-fil-proofs` to hold onto a single set of Pedersen params, then launch many Pedersen hash worker threads where each worker thread can hash any number of preimages simultaneously ("batched" Pedersen) and incrementally ("updateable" Pedersen).

This PR is a superset of functionality implemented in PR #925. The functionality added in this PR is the coordination of Pedersen hash param sharing across threads.

In this PR:

- adds an updateable Pedersen Hash function: `storage_proofs::crypto::parallel_pedersen::State`
- adds the type `storage_proofs::crypto::parallel_pedersen::ParallelPedersen` that coordinates any sharing of Pedersen hash params
    - `ParallelPedersen` does not add any overhead to sapling's `pedersen_hash` function for preimage sizes of less than 5 segments (~945 bits).
- `ParallelPedersen` is a drop in replacement for parallel, updateable Pedersen hash functionality currently in `storage_proofs/stacked/proof.rs`


dignifiedquire:
gonna close this for now, that we have moved to poseidon

PR #950, opened 2019-11-19
feat: pin last 3 versions of params

olizilla:
rewrite the pin-params.sh script to use the ipfs files api, to find and pin just the last 3 versions of the params in named directories, like:

```
.
├── v15
├── v16
└── v17
```

License: MIT
Signed-off-by: Oli Evans <oli@tableflip.io>


olizilla:
This script is now _cleverer than it needs to be_. Closing this in favour of a much simplified script that posts the params to cluster and publishes the new root hash as the dnslink. The gateways will handle retaining the last 3 versions.

PR #956, opened 2019-11-24
refactor: switch to anyhow from failure

dignifiedquire:
First step in preparation for better error handling. 

For some background on different error handling libraries see https://blog.yoshuawuyts.com/error-handling-survey/#libraries

Depends on https://github.com/filecoin-project/merkle_light/pull/54


vmx:
If this is the first step, what would be the next? I would really like to see us using https://github.com/dtolnay/thiserror only as a next step (which is kind of what I started to prepare at https://github.com/filecoin-project/rust-fil-sector-builder/pull/61).

dignifiedquire:
The next step is use the new infrastructure tho avoid `assert`s, `panic`s and `unwrap`s everyhwere. The reason only `thiserror` is not realistic, is that for a lot of assertions string errors make much more sense, instead of structured errors.

dignifiedquire:
The other part I want to do is to add `.context` to a lot of places, to increase the information detail given

vmx:
The reason I'd prefer `thiserror` over `anyhow` is that you could implement a conversion into a FFI error for each error without the need to do downcasting. If you want to have a simple string based error you could add a new type called something like `GenericError` which will just take a string.

vmx:
@dignifiedquire you might want to fixup the `cleanup` commit locally, so that the CI is passing (it's failing due to the commit linting).

dignifiedquire:
I didn’t change the tests due to time, focusing on user visible changes

PR #960, opened 2019-11-28
Reduce unwrap()s in code

vmx:
This is the first step to having less panics. I went through all the `unwrap()`s in the code and removed most of then (except the ones in tests and binaries). There is still a few unwraps left. Some are safe (when there are prior assertions about them) or where it is logically safe (like allocating a 32 byte array and writing 32 bytes into it.

There are still a few `unwrap()`s left with a todo. Those are the ones within a parallel iterator that is using `flat_map()`. Handling errors there is not that simple and needs more thought.

Please also note that this PR is only about `unwrap()`s. There is still other cases where panics could be triggered (e.g. by `assert!()`s) Those other cases will be fixed by subsequent PRs.

Please do a thorough review, some changes are non-trivial and perhaps sometimes other/better error messages might make sense.


whyrusleeping:
This PR makes me happy

PR #965, opened 2019-12-02
Less asserts

vmx:
I went through all the code and replaces asserts with `ensure!()`.

This one only concentrates on `asserts!()`, other cases for panics like `expect()` will be dealt with in a subsequent PR.

It would be great if that one could be merged soon, as I'd expect it to be quite painful to rebase in case of major changes as it touches so many files.


storage-proofs/src/circuit/uint64.rs
```diff
@@ -74,8 +74,8 @@ impl UInt64 {
         self.bits.iter().rev().cloned().collect()
     }
 
-    pub fn from_bits_be(bits: &[Boolean]) -> Self {
-        assert_eq!(bits.len(), 64);
+    pub fn from_bits_be(bits: &[Boolean]) -> anyhow::Result<Self> {
```
dignifiedquire: this will be called inside the circuit, should still be using `assert`

storage-proofs/src/circuit/uint64.rs
```diff
@@ -109,8 +109,8 @@ impl UInt64 {
 
     /// Converts a little-endian byte order representation of bits into a
     /// `UInt64`.
-    pub fn from_bits(bits: &[Boolean]) -> Self {
-        assert_eq!(bits.len(), 64);
+    pub fn from_bits(bits: &[Boolean]) -> anyhow::Result<Self> {
+        ensure!(bits.len() == 64, "Bit length must be 64.");
```
dignifiedquire: same as above

storage-proofs/src/stacked/challenges.rs
```diff
@@ -45,10 +46,10 @@ impl LayerChallenges {
         replica_id: &D,
         seed: &[u8; 32],
         k: u8,
-    ) -> Vec<usize> {
-        assert!(leaves > 2, "Too few leaves: {}", leaves);
+    ) -> Result<Vec<usize>> {
+        ensure!(leaves > 2, "Too few leaves: {}", leaves);
 
-        (0..challenges_count)
+        Ok((0..challenges_count)
```
dignifiedquire: that's a very large `Ok()` wrap, prefer to just say `let val= ..; Ok(val)`

vmx:
All tests pass, all past review comments addressed. It's ready for review again.

PR #968, opened 2019-12-04
Preparation for Testnet paramters

dignifiedquire:


filecoin-proofs/src/bin/paramcache.rs
```diff
@@ -52,19 +59,37 @@ fn cache_porep_params(porep_config: PoRepConfig) {
...
         >>::blank_circuit(&public_params);
-        let _ = StackedCompound::get_verifying_key(circuit, &public_params);
+        let _ = if is_predictable {
+            StackedCompound::get_verifying_key(
+                Some(&mut XorShiftRng::from_seed(SEED)),
+                circuit,
+                &public_params,
+            )
+        } else {
+            StackedCompound::get_verifying_key(Some(&mut OsRng), circuit, &public_params)
```
porcuquine: Does this actually work? I think the verifying key and groth params need to be based on the same randomness, so they will need to be generated together. (If this works, I wonder why and maybe am missing something.)

laser: @porcuquine - I've resolved this. Good catch.

PR #972, opened 2019-12-05
docs(filecoin-proofs): document public api parameters

DrPeterVanNostrand:


filecoin-proofs/src/api/mod.rs
```diff
@@ -148,6 +167,13 @@ pub fn generate_piece_commitment<T: std::io::Read>(
 /// wastes ($SIZESECTORSIZE/2)-$MINIMUM_PIECE_SIZE space. This function will be
 /// deprecated in favor of `write_and_preprocess`, and miners will be prevented
 /// from sealing sectors containing more than $TOOMUCH alignment bytes.
+///
+/// # Arguments
+///
+/// * `source` - a readable source of unprocessed piece bytes.
+/// * `target` - a writer where we will write the processed piece bytes.
+/// * `piece_size` - the number of bytes in this processed piece.
```
laser: This should be:

> the number of bytes in this piece.

or

> the number of unpadded user-bytes which can be read from `source` before EOF

filecoin-proofs/src/api/mod.rs
```diff
@@ -263,6 +289,12 @@ fn ensure_piece_size(piece_size: UnpaddedBytesAmount) -> Result<()> {
 /// `target`; it is the caller's responsibility to ensure properly sized
 /// and ordered writes to `target` such that `source`-bytes occupy whole
 /// subtrees of the final merkle tree built over `target`.
+///
+/// # Arguments
+///
+/// * `source` - a readable source of unprocessed piece bytes.
+/// * `target` - a writer where we will write the processed piece bytes.
+/// * `piece_size` - the number of bytes in this processed piece.
```
laser: Same as other comments: `piece_size` represents a quantity of not yet-preprocessed user bytes. So you could/should do the "quantity of bytes which can be read from `source` before EOF" thing.

PR #975, opened 2019-12-06
Perf/election post

dignifiedquire:


filecoin-proofs/src/api/post.rs
```diff
@@ -253,39 +255,50 @@ pub fn generate_post(
...
+                comm_c,
+                comm_q,
+                comm_r_last,
+            };
+
+            Ok((pub_inputs, priv_inputs))
+        })
+        .collect::<Result<_>>()?;
 
+    for (pub_inputs, priv_inputs) in &inputs {
```
vmx: This looks correct. But why not moving this loop into the `map()` above and return the proofs directly?

So instead of `Ok((pub_inputs, priv_inputs))` it would be

```rust
ElectionPoStCompound::prove(&pub_params, &pub_inputs, &priv_inputs, &groth_params)?
Ok(proof.to_vec()?)
```

dignifiedquire: Because they trigger the GPU in most cases, the expensive part. Which is why I try to avoid to parallelize them

storage-proofs/src/election_post.rs
```diff
@@ -227,14 +228,10 @@ pub fn generate_sector_challenges(
...
 ) -> Result<Vec<SectorId>> {
-    let mut challenges = Vec::with_capacity(challenge_count as usize);
-
-    for n in 0..challenge_count as usize {
-        let sector = generate_sector_challenge(randomness, n, sectors)?;
-        challenges.push(sector);
-    }
-
-    Ok(challenges)
+    (0..challenge_count)
```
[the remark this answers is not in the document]
dignifiedquire: yes, this is a bit inconsistent here, but I think we already have a compile check, or an issue for it, that the whole thing only compiles on systems where u64 == usize

PR #977, opened 2019-12-08
fix: handle updated cache path between preseal and seal

dignifiedquire:


storage-proofs/src/stacked/params.rs
```diff
@@ -486,6 +487,16 @@ pub struct TemporaryAux<H: Hasher, G: Hasher> {
 }
 
 impl<H: Hasher, G: Hasher> TemporaryAux<H, G> {
+    pub fn set_cache_path<P: AsRef<Path>>(&mut self, cache_path: P) {
+        for label in self.labels.labels.iter_mut() {
+            label.path = cache_path.as_ref().to_path_buf();
+        }
+        self.tree_d_config.path = cache_path.as_ref().to_path_buf();
```
vmx: Nit (no need to change it): `cache_path.as_ref().to_path_bug()` is called a lot, it could be extracted in a variable.

dignifiedquire: fixing

PR #978, opened 2019-12-08
fix: reduce add piece memory

dignifiedquire:
Blocked by [this upstream merkle PR](https://github.com/filecoin-project/merkle_light/pull/60).

## What's in this PR?

- stop using in memory merkle tree for generating piece commitments
- stop caching the whole data for a piece commitment in memory


filecoin-proofs/src/api/mod.rs
```diff
@@ -117,20 +117,20 @@ pub fn generate_piece_commitment<T: std::io::Read>(
         ));
     }
 
-    let n = UnpaddedBytesAmount(n as u64);
+    let n_u = UnpaddedBytesAmount(n as u64);
```
vmx: nit: this is really error-prone for the future `n` and `n_u`. I would name it e.g. `unpadded`.

dignifiedquire:
> It looks right after a quick look, but in case it isn't urgent i'll have another look tomorrow (i'm too tired now).

thanks, need to fix some bugs in it first, from the looks of CI,

PR #982, opened 2019-12-10
fix(benchmarking): clean temp directory on exit

laser:
## Why does this PR exist?

The existing remote benchmarking code doesn't clean up the temp directory (in which it builds rust-fil-proofs), causing the host to run out of disk after a few days.

## What's in this PR?

This changeset removes the temporary directory after the benchmark completes.


laser:
@vmx - double quotes [are hard](is generally a good idea) within heredoc. Unless you strongly object (it doesn't appear to be the case that you do), I'm going to move forward with the not-quoted changeset.

vmx:
@laser Oh right. Sure, go ahead. Your shell-foo is way better than mine anyway, I'm just a human linter ;)

PR #984, opened 2019-12-10
feat: make SealPreCommitOutput clonable

vmx:
This is usefult for benchmarking if you want to seal several times with
the same input.

I don't know if the original behaviour is a bug or feature.


vmx:
I'ts a feature, hence closing this PR.

Quoting @dignifiedquire:

> commit deletes some of that data so it really should not be reused/can not be reused for multiple calls

PR #985, opened 2019-12-10
* feat(storage-proofs): switch post challenges to 40

dignifiedquire:


filecoin-proofs/src/bin/paramcache.rs
```diff
@@ -195,12 +202,14 @@ pub fn main() {
...
         );
 
-        cache_porep_params(
-            is_predictable,
-            PoRepConfig {
-                sector_size: SectorSize(sector_size),
-                partitions: DEFAULT_POREP_PROOF_PARTITIONS,
-            },
-        );
+        if !only_election_post {
```
laser: @dignifiedquire - Seems a little weird to allow the consumer to specify that they want:

1. Only PoRep parameters
1. Both PoRep and PoSt parameters

and not also...

1. Only PoRep parameters

dignifiedquire: I know, it wwas what I needed today

PR #986, opened 2019-12-10
feat: read groth parameters lazily via mmap (coarsely)

cryptonemo:


cryptonemo:
Requires: https://github.com/filecoin-project/bellman/pull/37

ValarDragon:
out of curiousity, do you know how much of a speed improvement this is?

cryptonemo:
> out of curiousity, do you know how much of a speed improvement this is?

Thanks for asking, but the purpose of this PR is strictly runtime RAM reduction, not necessarily speedup type improvements.  In my testing, runtime speed is about the same.

ValarDragon:
Thanks for the explanation!

dignifiedquire:
@cryptonemo I think you have to update the rust version to make the tests work

filecoin-proofs/src/caches.rs
```diff
@@ -62,14 +62,6 @@ where
     Ok(res)
 }
 
-#[inline]
-pub fn lookup_groth_params<F>(identifier: String, generator: F) -> Result<Arc<Bls12GrothParams>>
-where
-    F: FnOnce() -> Result<Bls12GrothParams>,
-{
-    cache_lookup(&*GROTH_PARAM_MEMORY_CACHE, identifier, generator)
```
dignifiedquire: Why did you remove the use of the cache? shouldn't we still use that?

cryptonemo:
> @cryptonemo I think you have to update the rust version to make the tests work

I made this PR as an alternative. Any preference?

https://github.com/filecoin-project/merkle_light/pull/68

filecoin-proofs/src/caches.rs
cryptonemo: Compiler was telling me it wasn't used.

dignifiedquire: I mean, you should keep using it

cryptonemo: Ok, let me take a look, thanks.

cryptonemo:
> > @cryptonemo I think you have to update the rust version to make the tests work
> 
> I made this PR as an alternative. Any preference?
> 
> [filecoin-project/merkle_light#68](https://github.com/filecoin-project/merkle_light/pull/68)

Could I also get a version bump to make this PR happy?

storage-proofs/src/parameter_cache.rs
```diff
@@ -182,31 +181,14 @@ where
     }
 
     fn get_groth_params<R: RngCore>(
-        rng: Option<&mut R>,
-        circuit: C,
+        _rng: Option<&mut R>,
+        _circuit: C,
```
dignifiedquire: can you remove these argumentes, no need to keep them around

cryptonemo: Hrm, so sort of.  I tried removing both at once and all hell broke loose not knowing about the circuit in other parts, so I reverted for the moment and just removed the rng.

fil-proofs-tooling/Cargo.toml
```diff
@@ -39,7 +39,8 @@ blake2s_simd = "0.5.6"
 fil_logger = "0.1"
 log = "0.4.8"
 uom = "0.26"
-merkletree = "0.14.0"
+#merkletree = "0.15.0"
+merkletree = { git = "https://github.com/filecoin-project/merkle_light" }
```
dignifiedquire: I just release `0.15.1`, so you can switch to that

dignifiedquire:
@cryptonemo looks like this actually broke some tests I am afraid

cryptonemo:
> @cryptonemo looks like this actually broke some tests I am afraid

@dignifiedquire My read of the errors looked like they were v21 param related failures. Not the case?

storage-proofs/src/parameter_cache.rs
```diff
@@ -300,6 +303,7 @@ fn write_cached_verifying_key<E: JubjubEngine>(
     })
 }
 
+#[allow(dead_code)]
 fn write_cached_params<E: JubjubEngine>(
```
dignifiedquire: why is this dead code, isn't this used above Oo

dignifiedquire:
@cryptonemo the test failure looks real

storage-proofs/src/parameter_cache.rs
cryptonemo: Yes, was previously unused and I missed it after adding back in, thanks.

storage-proofs/src/parameter_cache.rs
```diff
@@ -181,43 +179,45 @@ where
...
-                Ok(parameters)
-            } else {
-                bail!("No cached parameters found for {}", id);
-            }
+            use rand::SeedableRng;
+            use rand_xorshift::XorShiftRng;
+            use std::time::Instant;
+
+            info!("Actually generating groth params. (id: {})", &id);
+            let rng = &mut XorShiftRng::from_seed(crate::TEST_SEED);
```
dignifiedquire: Sorry I forgot this before, lets not use a seeded rng, but rather `thread_rng`, these should not be regeneratable.

cryptonemo: Ok, thanks, will update this shortly!

cryptonemo:
Updated and rebased!

PR #993, opened 2019-12-11
feat: add gpu-cpu mover test tool

vmx:
Proofs can be forced to run on the CPU instead of the GPU. This is used to
run some proofs with higher priority.

The `gpu-cpu-test` tool tests if this actually works. It spawns two threads
which run proofs. Those get killed after 5 minutes of running. The overall
test runs longer as some input data needs to be generated. By default one
thread will always be prioritized to run on the GPU. The other one might be
moved to the CPU.

When running:

    cd fil-proofs-tooling
    RUST_LOG=trace cargo run --release --bin gpu-cpu-test

The return values of the tests might look like:

    2019-12-11T18:13:19.518 main INFO gpu_cpu_test > Thread HighPrio info: RunInfo { elapsed: 303.366942495s, iterations: 28 }
    2019-12-11T18:13:25.981 main INFO gpu_cpu_test > Thread LowPrio info: RunInfo { elapsed: 309.829930518s, iterations: 15 }

Clearly the high priority thread got more work done.

When running without the "GPU stealing" feature, where one tread demands to
run on the GPU:

    RUST_LOG=trace cargo run --release --bin gpu-cpu-test -- --gpu-stealing false

The return values indicate that both threads got the same amount of time on the GPU:

    2019-12-11T18:30:16.868 main INFO gpu_cpu_test > Thread HighPrio info: RunInfo { elapsed: 307.388469955s, iterations: 23 }
    2019-12-11T18:30:16.868 main INFO gpu_cpu_test > Thread LowPrio info: RunInfo { elapsed: 300.893010419s, iterations: 22 }

This PR depends on https://github.com/filecoin-project/bellman/pull/35.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -5,3 +5,6 @@ members = [
   "storage-proofs",
   "fil-proofs-tooling",
 ]
+
+[patch.crates-io]
+bellperson = { git = "https://github.com/finalitylabs/bellman", branch="fil-lock" }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -7,4 +7,6 @@ members = [
 ]
 
 [patch.crates-io]
-bellperson = { git = "https://github.com/finalitylabs/bellman", branch="fil-lock" }
+#bellperson = { git = "https://github.com/finalitylabs/bellman", branch="fil-lock" }
+bellperson = { git = "https://github.com/vmx/bellman", branch="gpu-lock-plus-fixes" }
+#bellperson = { git = "https://github.com/vmx/bellman", branch="gpu-lock" }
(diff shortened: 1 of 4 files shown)


keyvank:
Can you make two different processes instead of two different threads? Because in case of threads, GPU kernels would be still shared across them and CL_MEM_OBJECT_ALLOC_FAILURES might not happen.

vmx:
@dignifiedquire What's out use case? Multiple threads or multiple processes? Should I add another test with multiple processes?

dignifiedquire:
@vmx both

vmx:
Closing this one if favour of https://github.com/filecoin-project/rust-fil-proofs/pull/1025.

PR #994, opened 2019-12-12
DO NOT MERGE YET: fix(filecoin-proofs): stricter checks on election post

dignifiedquire:


Kubuxu:
@dignifiedquire If this merges we cannot update to it (due to testnet bug we had) or any filecoin-ffi that contains it.

Kubuxu:
@dignifiedquire If this merges we cannot update to it (due to testnet bug we had) or any filecoin-ffi that contains it.

PR #997, opened 2019-12-13
feat(metrics): consolidate metrics scripts + improve cleanup

laser:
## Why does this PR exist?

The previous mechanism of `rm -rf`'ing the temporary metrics directory is a little error prone in that the `rm` won't be run if the benchmarking script fails. Also, I'd duplicated a bunch of code needlessly across the Groth parameter-checker, micro, and benchy scripts.

Additionally, the old `with-lock.sh` script didn't handle leaked locks which could result from the process which acquired the lock having been `SIGKILL`'ed (or some other shutdown will circumvented the `TRAP`).

## What's in this PR?

- clean up any temporary directories created by metrics runs after 1 day
- consolidate all duplicated metrics scripts into single "run-remote.sh" script
- force-relinquish leaked locks


.circleci/config.yml
```diff
@@ -188,30 +188,30 @@ jobs:
       - run:
           name: Ensure existence of Groth parameters and keys on remote host
           command: |
-            ./fil-proofs-tooling/scripts/paramcache-remote.sh "${CIRCLE_BRANCH}" "${BENCHMARK_SERVER_SSH_USERNAME}@${BENCHMARK_SERVER_IP_ADDR}" "-z=$((1024*1024*1024))"
+            ./fil-proofs-tooling/scripts/run-remote.sh "${CIRCLE_BRANCH}" "${BENCHMARK_SERVER_SSH_USERNAME}@${BENCHMARK_SERVER_IP_ADDR}" cargo run --release --package filecoin-proofs --bin=paramcache -- -z=$((1024*1024*1024))
```
vmx: I can see not changing this with the PR, but in general I think it's a good idea to use long parameter names in scripts. things like `-z` are just so typo-prone.

fil-proofs-tooling/scripts/run-remote.sh
```diff
@@ -0,0 +1,31 @@
+#!/usr/bin/env bash
+
+CMDS=$(cat <<EOF
+
+export RUST_LOG=info
+set -e
+
+# Creates a temporary directory in which we build rust-fil-proofs and capture
+# performance metrics. The name of the directory (today's UTC seconds plus 24
```
vmx: Wouldn't it make more sense to put the metrics with the current date and let the cleanup call remove everything from yesterday and older?

But I'm sure you've thought about that, so feel free to just leave it like that :)

laser:
@dignifiedquire may I have a review plz?

dignifiedquire:
@laser anything missing here?

dignifiedquire:
@laser @vmx any thoughts on what to do with this?

laser:
> @laser @vmx any thoughts on what to do with this?

I'll get it ready to merge today.

fil-proofs-tooling/scripts/run-remote.sh
laser: > Wouldn't it make more sense to put the metrics with the current date and let the cleanup call remove everything from yesterday and older?

The way which I've done things keeps the cleanup mechanism ignorant of any expiration policy (in this case, that the directory should be deleted one day after creation); the thing which creates the directory selects an expiration date.

.circleci/config.yml
laser: Just updated to use the long parameter name, as per your suggestion @vmx

dignifiedquire:
@vmx could you take a look, the metric capture is still failing

PR #998, opened 2019-12-16
feat: use LevelCacheStore for restoring tree_r_last from disk

cryptonemo:
~~Happens to be built on top of the mmap params branch~~, but in any case, it's a start of the LevelCacheStore integration where we can compact `tree_r_last` and generate the proofs using partially built trees.

### DIFF
--- a/filecoin-proofs/src/api/post.rs
+++ b/filecoin-proofs/src/api/post.rs
@@ -6,8 +6,8 @@ use std::path::{Path, PathBuf};
 use anyhow::{anyhow, ensure, Context, Result};
 use bincode::deserialize;
 use log::info;
-use merkletree::merkle::{get_merkle_tree_leafs, MerkleTree};
-use merkletree::store::{DiskStore, Store, StoreConfig, DEFAULT_CACHED_ABOVE_BASE_LAYER};
+use merkletree::merkle::get_merkle_tree_leafs;
+use merkletree::store::{LevelCacheStore, Store, StoreConfig, DEFAULT_CACHED_ABOVE_BASE_LAYER};
 use paired::bls12_381::Bls12;
 use rayon::prelude::*;
 use storage_proofs::circuit::election_post::ElectionPoStCompound;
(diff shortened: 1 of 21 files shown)


cryptonemo:
Requires https://github.com/filecoin-project/merkle_light/pull/55

v1 compaction is used here since the external reader usage for v2 compaction isn't addressed yet.

cryptonemo:
I'll be rebasing out the mmap commit, since it seems priority-wise, it will come later.

filecoin-proofs/src/api/post.rs
```diff
@@ -220,14 +223,32 @@ pub fn generate_candidates(
     let unique_trees_res: Vec<_> = unique_challenged_replicas
         .into_par_iter()
         .map(|(id, replica)| {
+            // Ensure that any associated cached data persisted is discarded and our tree is compacted by this point.
+            let f_aux_path = replica.cache_dir_path().join(CacheKey::TAux.to_string());
+            if Path::new(&f_aux_path).exists() {
```
dignifiedquire: this should probably return an error if the path does not exist

dignifiedquire:
@cryptonemo lets wait for this until we merge #1016 and then switch to using the store for `CommRLastTree`, so we avoid double storing the replica.

dignifiedquire:
@cryptonemo tests are failing

dignifiedquire:
@cryptonemo looks like a real test failure

cryptonemo:
> @cryptonemo looks like a real test failure

@dignifiedquire I finally found it.  It's slightly annoying because it's a `merkle_light` bug that was fixed in the `extended-branches` branch a while ago, but apparently lost on the master branch.  See https://github.com/filecoin-project/merkle_light/pull/69

cryptonemo:
Ooh, my rebase made the history very messy after your merge.  Not sure what happened there.  I'll revert and clean up a bit later though

cryptonemo:
> Ooh, my rebase made the history very messy after your merge. Not sure what happened there. I'll revert and clean up a bit later though

Actually, it just looked messy on github until I reloaded.  I did squash some of the commits though to clean it up.  If I missed anything that your merge didn't cover, let me know.

PR #1010, opened 2019-12-19
feat: implement batch verification of snarks

dignifiedquire:
Depends on https://github.com/filecoin-project/bellman/pull/45

TODOS:

- [ ] switch to regular verification if only a single partition
- [ ] write a test for batch verification


filecoin-proofs/src/api/seal.rs
```diff
@@ -395,3 +395,111 @@ pub fn verify_seal(
...
+            &verifying_key,
+        )?);
+    }
+
+    StackedCompound::batch_verify(
+        &compound_public_params,
+        &public_inputs,
+        &proofs,
+        &ChallengeRequirements {
+            minimum_challenges: POREP_WINDOW_MINIMUM_CHALLENGES.load(Ordering::Relaxed) as usize, // TODO: what do we want here?
```
porcuquine: This is okay, but I note that we are assuming the same number of challenges for windows and wrapper layer. If that assumption changes, this will need to become more sophisticated.

storage-proofs/src/compound_proof.rs
```diff
@@ -122,11 +122,13 @@ where
...
+            multi_proof.circuit_proofs.len() == Self::partition_count(public_params),
+            "Inconsistent inputs"
+        );
+
         let vanilla_public_params = &public_params.vanilla_params;
-        let pvk = groth16::prepare_verifying_key(&multi_proof.verifying_key);
-        if multi_proof.circuit_proofs.len() != Self::partition_count(public_params) {
-            return Ok(false);
-        }
+        let pvk = groth16::prepare_batch_verifying_key(&multi_proof.verifying_key);
```
porcuquine: I guess we could cache this and avoid incurring the cost on every batch verification, too.

filecoin-proofs/src/api/seal.rs
dignifiedquire: I agree, but that was already in the code before

storage-proofs/src/compound_proof.rs
dignifiedquire: We could, but that would require leaking this outside of the proofs code

dignifiedquire:
already merged manually

PR #1020, opened 2020-01-29
Feat/poseidon integration exp

porcuquine:
This integrates `neptune` (our Poseidon library) and implements the `Hasher` trait. Tests pass for me locally. I tried to run `benchy flarp`, but it seems to just hang. However, that's the same behavior I see without these changes, so I think I just don't know to run it.

This should give much better SNARK size (fewer constraints), as can be seen in these `por.rs` tests:
```
    #[test]
    fn test_por_input_circuit_with_bls12_381_pedersen() {
        test_por_input_circuit_with_bls12_381::<PedersenHasher>(4125);
    }

    #[test]
    fn test_por_input_circuit_with_bls12_381_blake2s() {
        test_por_input_circuit_with_bls12_381::<Blake2sHasher>(64566);
    }

    #[test]
    fn test_por_input_circuit_with_bls12_381_poseidon() {
        test_por_input_circuit_with_bls12_381::<PoseidonHasher>(1290);
    }
```

I expect non-circuit hashing performance to be worse for now, and improving it is a non-goal of this integration step.

NOTE: CI is failing to generate groth parameters and killing the job, even after I increased the timeout to 30 minutes. I don't know what's up with that.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -23,6 +23,7 @@ jobs:
       - run:
           name: Calculate dependencies
           command: cargo generate-lockfile
+          no_output_timeout: 30m
       - restore_cache:
           keys:
             - cargo-v14-{{ checksum "rust-toolchain" }}-{{ checksum "Cargo.toml" }}-{{ checksum "Cargo.lock" }}-{{ arch }}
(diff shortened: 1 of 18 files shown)


merced:
In poseidon node function, Fr::from_repr is called for both left and right nodes as input values, while in FF package from_repr function does impose the input values has to be less than Fr modulus, otherwise the function will just return error result, please confirm if merkle tree failure due to this is possible or not from you point of view, but I do think we will see a lot of merkle tree failure because of this.

dignifiedquire:
closing in favor of #1027

PR #1021, opened 2020-01-30
fix: unify hash handling and fix up tests

dignifiedquire:
- poseidon tests fail, can't measure yet, this seems to be an issue with the circuit in neptune
- Currently this changes pedersen hashing to add padding to full bytes in all cases, as that was simpler to make happen. Should decide if we are okay with that or not. 

```
---- circuit::stacked::proof::tests::test_stacked_compound_poseidon stdout ----
thread 'circuit::stacked::proof::tests::test_stacked_compound_poseidon' panicked at 'called `Result::unwrap()` on an `Err` value: AssignmentMissing', /Users/dignifiedquire/.cargo/registry/src/github.com-1ecc6299db9ec823/neptune-0.2.1/src/circuit.rs:399:29
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```


dignifiedquire:
closing in favor of #1027

PR #1023, opened 2020-02-04
Feat/api cleanup

dignifiedquire:


filecoin-proofs/src/api/seal.rs
```diff
@@ -175,6 +178,23 @@ where
     };
     let data: storage_proofs::porep::Data<'_> = (data, PathBuf::from(out_path.as_ref())).into();
 
+    // Load data tree from disk
+    let data_tree = {
+        let config = StoreConfig::new(
+            cache_path.as_ref(),
+            CacheKey::CommDTree.to_string(),
+            DEFAULT_CACHED_ABOVE_BASE_LAYER,
```
cryptonemo: This should now be `StoreConfig::default_cached_above_base_layer(tree_leafs)`

dignifiedquire: I don't think this branch is yet updated to the latest, so this tries to keep in line with the call above in the same file.

PR #1024, opened 2020-02-04
feat: allow runtime switch for disk vs memory replication

dignifiedquire:
new config options

- `FIL_PROOFS_REPLICATE_ON_DISK` when set will run replication using as much disk as possible (default: `false`)
- `FIL_PROOFS_CACHE_GRAPH` when set will cache the graph in memory (default: `true`) this option was previously known as `MAXIMIZE_CACHING`


dignifiedquire:
Note: it looks like CI is much slower with this PR, so need to investigate

dignifiedquire:
Closing as this is very outdated now.

PR #1025, opened 2020-02-05
Run PoSt proofs with high priority

vmx:
This is using the priority locking from https://github.com/filecoin-project/bellman/pull/58.

I've split it into two commits, the second commit isn't strictly needed but useful for testing, so that you can run the PoSt also with low priority.


vmx:
When running my test I get strange result. This is only about running things in 1 process in 2 threads (multiprocess is yet a different story).

Checkout this PR and run the command with the root directory.

Running with one thread having a high priority:
```console
$ BELLMAN_CUSTOM_GPU="Tesla V100-SXM2-16GB:5120" RUST_LOG=debug cargo run --release --bin gpu-cpu-test 2>&1| tee out001.log
…
2020-02-05T21:42:44.419 29733 low INFO storage_proofs::compound_proof > snark_proof:start
…
2020-02-05T21:42:45.813 29733 high INFO storage_proofs::compound_proof > snark_proof:start
…
2020-02-05T21:47:38.241 29733 main INFO gpu_cpu_test > Thread high info: RunInfo { elapsed: 310.107977855s, iterations: 23 }
2020-02-05T21:47:38.241 29733 main INFO gpu_cpu_test > Thread low info: RunInfo { elapsed: 308.139481678s, iterations: 6 }
```
Things look correct, Both start the actual proofing at the same time and after 5 mins they are both done. The high priority thread obviously is able to run more iterations in about the same time as it runs on the GPU, while the low priority one is on the CPU.

---

Running with both threads without special priority:
```console
$ BELLMAN_CUSTOM_GPU="Tesla V100-SXM2-16GB:5120" RUST_LOG=debug cargo run --release --bin gpu-cpu-test -- --gpu-stealing false 2>&1| tee out002.log
…
2020-02-05T21:55:11.385 1761 high INFO storage_proofs::compound_proof > snark_proof:start
…
2020-02-05T21:55:19.581 1761 low INFO storage_proofs::compound_proof > snark_proof:start
…
2020-02-05T22:00:08.023 1761 main INFO gpu_cpu_test > Thread high info: RunInfo { elapsed: 304.69049444s, iterations: 24 }
…
2020-02-05T22:00:13.844 1761 main INFO gpu_cpu_test > Thread low info: RunInfo { elapsed: 310.512290479s, iterations: 24 }
```
Now things get weird. No threads tries to get the lock. Again both start at about the same time and are also both finished after about 5 mins. Though now both of them are able to have more iterations. Although they run *both* on the GPU so they should block the GPU longer I guess?.

---

Full logs are available at https://gist.github.com/vmx/43f4f5fa48f8602ad80cd1a58582fb18

vmx:
I had a closer look. While running the code I had `nvidia-smi pmon --options DT` running. When there is a high priority thread running, there are always gaps between actual GPU usage. Those gaps are big enough to have another proof with that size running.

And this is exactly what happens when none of the threads have a high priority. It just perfectly fits in to have them queued up on the GPU and have them run one after another.

I'll try to change the test a bit to be more realistic.

vmx:
I ran it with more threads. The locking clearly works. Though I'm not sure if it's worth it. I run the test a specific amount of time (5 mins) and look how many proofs I was able to perform. If I run one thread with high priority, this thread is able to run about 23 times.

 - With 1 lower priority thread, that one is run about 6 times
 - With 3 lower priority threads: each about 2 times
 - With 5 lower priority threads: each about 1 times

So the good news is, the number of times the higher priority thread is run is independent on how many lower priority threads run. That works as expected, that's good news.

Now to the bad news. If I don't use the priority stuff at all and run several threads, the total amount of proofs is way higher, as they max out the GPU more.

 - With 2 threads: each about 24 times
 - With 4 threads: each about 12 times
 - With 6 threads: each about 9 times

I think the goal of this whole effort should be to maximize the proofs run on the system, while making sure one has priority. The current version leads to overall less proofs.

vmx:
Once https://github.com/filecoin-project/bellman/pull/58 is merged, I'll update this PR, then it's ready to be merged.

vmx:
Ready for review!

Note that I've made some tests (where I think they are PoSts) to use the "high priority" code path as I thought it makes sense to use the same code path in tests as when running it.

PR #1026, opened 2020-02-05
feat: improved approach to do bit padding/calculate piece commitments

dignifiedquire:


vmx:
I didn't change anything, I just squashed some commits for easier future rebasing.

vmx:
The file I checked with, it returned the expected result. I can also confirm that no disk is used and the memory consumption is low.

The memory and disk consumption only spikes when it comes to the merkletree building. The disk usage can be brought down to zero if a vector store instead of the disk store is used.

PR #1027, opened 2020-02-05
feat: switch to poseidon by default

dignifiedquire:
This switches 
- column hashes and
- merkle trees

to use Poseidon

Outstanding are 
- switch to poseidon for the partial ticket (election post)
which will be done at a later point


storage-proofs/src/hasher/poseidon.rs
```diff
@@ -0,0 +1,453 @@
...
+            <Bls12 as ff::ScalarEngine>::Fr::from_repr(PoseidonDomain::from_slice(chunk).0).unwrap()
+        })
+        .collect::<Vec<_>>();
+
+    shared_hash_frs(&preimage)
+}
+
+fn shared_hash_frs(preimage: &[<Bls12 as ff::ScalarEngine>::Fr]) -> PoseidonDomain {
+    let mut p = Poseidon::new_with_preimage(&preimage, &*POSEIDON_CONSTANTS);
+    let fr: <Bls12 as ScalarEngine>::Fr = p.hash_in_mode(HashMode::OptimizedStatic);
```
porcuquine: It might be better to just call `p.hash()` here, as that is the default. That should probably be made the only public method, so don't have to continue to support that interface — which should probably be internal to the library.

dignifiedquire: Ah for some reason I thought that was not the default, will switch

storage-proofs/Cargo.toml
```diff
@@ -50,11 +50,13 @@ num_cpus = "1.10.1"
 rand_xorshift = "0.2.0"
 rand_chacha = "0.2.1"
 hex = "0.4.0"
-generic-array = "0.12"
+generic-array = "0.13.2"
 anyhow = "1.0.23"
 thiserror = "1.0.6"
 cpu-time = "1.0.0"
-gperftools = "0.2.0"
+neptune = "0.3.1"
```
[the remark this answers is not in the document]
dignifiedquire: fixed

storage-proofs/src/hasher/poseidon.rs
dignifiedquire: fixed

PR #1028, opened 2020-02-10
feat: switch to fff

dignifiedquire:
Forked version of ff, as well as latest versions of all snark deps


fil-proofs-tooling/Cargo.toml
```diff
@@ -42,7 +42,7 @@ uom = "0.26"
 merkletree = "0.14.0"
 bincode = "1.1.2"
 anyhow = "1.0.23"
-ff = "=0.5.0"
+ff = { version = "0.2.0", package = "fff" }
```
cryptonemo: @dignifiedquire Haven't seen this syntax before -- so it allows you to use the `fff` package while it's internally referred to as `ff` still?

dignifiedquire: yes, it's like a rename, and reduces the search and replace :D

cryptonemo: That's convenient, thanks!

PR #1030, opened 2020-02-12
feat: switch to quad trees in stacked

dignifiedquire:


storage-proofs/src/stacked/proof.rs
```diff
@@ -388,12 +393,30 @@ impl<'a, H: 'static + Hasher, G: 'static + Hasher> StackedDrg<'a, H, G> {
...
-        let mut tree_r_last_config =
-            StoreConfig::from_config(&config, CacheKey::CommRLastTree.to_string(), None);
-        let mut tree_c_config =
-            StoreConfig::from_config(&config, CacheKey::CommCTree.to_string(), None);
+        let mut tree_d_config = StoreConfig::from_config(
+            &config,
+            CacheKey::CommDTree.to_string(),
+            Some(StoreConfig::default_cached_above_base_layer(
+                data.len(),
+                BINARY_ARITY,
```
cryptonemo: Shouldn't all of the tree arities be matching?

storage-proofs/src/stacked/proof.rs
```diff
@@ -454,7 +477,7 @@ impl<'a, H: 'static + Hasher, G: 'static + Hasher> StackedDrg<'a, H, G> {
                 trace!("building merkle tree for the original data");
                 data.ensure_data()?;
                 measure_op(CommD, || {
-                    Self::build_tree::<G>(data.as_ref(), tree_d_config.clone())
+                    Self::build_binary_tree::<G>(data.as_ref(), tree_d_config.clone())
```
cryptonemo: I see that it's not here, just want to be sure it's intentional.

storage-proofs/src/stacked/proof.rs
dignifiedquire: no, comm_d will stay a binary tree

dignifiedquire: maybe will change

storage-proofs/src/stacked/proof.rs
dignifiedquire: yes it is for now

cryptonemo: Ok, thanks

PR #1034, opened 2020-02-17
feat: integrate external data reader for tree r

cryptonemo:


storage-proofs/src/election_post.rs
```diff
@@ -481,18 +486,40 @@ mod tests {
...
             assert_eq!(c, true);
 
+            // Open the replica for base layer data reads.
+            let reader = OpenOptions::new()
+                .read(true)
+                .open(&replica_path)
+                .expect("Failed to open replica (base layer data)");
+
+            // Configure an external reader for access to the base layer data.
+            let external_reader = ExternalReader {
```
dignifiedquire: this `ExternalReader` is always the same, I would suggest moving this impl into `.lcmerkle_tree` and just give it the path.

storage-proofs/src/merkle.rs
```diff
@@ -285,33 +288,48 @@ pub fn create_merkle_tree<H: Hasher, U: typenum::Unsigned>(
...
-        // iterator case.
-        H::Domain::try_from_bytes(d).expect("failed to convert node data to domain element")
+    // Open the replica for base layer data reads.
+    let reader = OpenOptions::new()
+        .read(true)
+        .open(replica_path)
+        .expect("Failed to open replica (base layer data)");
+
+    // Configure an external reader for access to the base layer data.
+    let external_reader = ExternalReader {
```
dignifiedquire: could use the new method here

storage-proofs/src/stacked/params.rs
```diff
@@ -362,9 +364,9 @@ impl<H: Hasher, G: Hasher> TemporaryAux<H, G> {
         self.labels.column(column_index)
     }
 
-    // "Compact" will discard all persisted data that is no longer
-    // required and compact the remaining "tree_r_last" merkle tree.
-    pub fn compact(t_aux: TemporaryAux<H, G>) -> Result<()> {
+    // 'delete' will discard all persisted merkle and layer data that
+    // is no longer required.
+    pub fn delete(t_aux: TemporaryAux<H, G>) -> Result<()> {
```
dignifiedquire: We still keep part of the cache, so I think calling it `compact` is still the right name

cryptonemo: We do, but that method only deletes now. Compaction is done elsewhere, immediately after the trees are built.

cryptonemo: :+1: Renamed to `clear_temp`

PR #1037, opened 2020-02-19
feat: add validation to check cache state and replica

### LINKED ISSUE #1022: feat: add api methods to validate cache
We have the request from the lotus team to add two methods

- `validate_cache_for_commit`
- `validate_cache_for_precommit_phase2`

These should check the passed in cache dir and make sure that everything is there that is expected for the upcoming phase.

cryptonemo:
Closes #1022

### DIFF
--- a/filecoin-proofs/src/api/mod.rs
+++ b/filecoin-proofs/src/api/mod.rs
@@ -4,15 +4,13 @@ use std::path::{Path, PathBuf};
 
 use anyhow::{anyhow, ensure, Context, Result};
 use bincode::deserialize;
-use log::trace;
 use merkletree::merkle::get_merkle_tree_leafs;
 use merkletree::store::{DiskStore, LevelCacheStore, StoreConfig};
 use storage_proofs::hasher::Hasher;
 use storage_proofs::porep::PoRep;
 use storage_proofs::sector::SectorId;
 use storage_proofs::stacked::{
-    generate_replica_id, CacheKey, PersistentAux, StackedDrg, TemporaryAux, BINARY_ARITY,
-    QUAD_ARITY,
+    generate_replica_id, CacheKey, PersistentAux, StackedDrg, TemporaryAux, BINARY_ARITY, OCT_ARITY,
 };
 use tempfile::tempfile;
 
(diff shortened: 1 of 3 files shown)


cryptonemo:
Addresses https://github.com/filecoin-project/rust-fil-proofs/issues/1022

filecoin-proofs/src/api/mod.rs
```diff
@@ -318,6 +323,153 @@ where
     add_piece(source, target, piece_size, Default::default())
 }
 
+// Verifies if a DiskStore specified by a config is consistent.
+fn verify_store(config: &StoreConfig, arity: usize) -> bool {
```
dignifiedquire: To make things a little more helpful, can we change this method to return `Result<()>`, and return things like `Missing store file: bla` or `Missing size` as errors if something doesn't match?

filecoin-proofs/src/api/mod.rs
```diff
@@ -318,6 +323,153 @@ where
...
+                &current_path,
+                &cache
+            );
+            return Ok(false);
+        }
+    }
+
+    Ok(verify_store(
+        &seal_precommit_phase1_output.config,
+        BINARY_ARITY,
```
dignifiedquire: is this only used for comm_d?

filecoin-proofs/src/api/mod.rs
```diff
@@ -318,6 +323,153 @@ where
...
+// Checks for the existence of the replica data and t_aux, which in
+// turn allows us to verify the tree d, tree r, tree c, and the
+// labels.
+pub fn validate_cache_for_commit<R, T>(cache_path: R, replica_path: T) -> Result<bool>
+where
+    R: AsRef<Path>,
+    T: AsRef<Path>,
+{
+    let replica = replica_path.as_ref().to_path_buf();
+    if !Path::new(&replica).exists() {
```
dignifiedquire: you should be able to write `replica_path.as_ref().exists()`

filecoin-proofs/src/api/mod.rs
```diff
@@ -318,6 +323,153 @@ where
...
+        trace!("replica {:?} does not exist", &replica);
+        return Ok(false);
+    }
+
+    for label in &seal_precommit_phase1_output.labels.labels {
```
dignifiedquire: I think it would a bit better to have this be a method no the `Labels` struct, so we don't forget to update the logic here, when that changes

filecoin-proofs/src/api/mod.rs
cryptonemo: Yes.  At that stage, I believe only labels and comm_d are present.

cryptonemo:
Resolves #1022 (just noting for future issue closing if/when this is merged).

filecoin-proofs/src/api/mod.rs
```diff
@@ -318,6 +321,131 @@ where
     add_piece(source, target, piece_size, Default::default())
 }
 
+// Verifies if a DiskStore specified by a config is consistent.
+fn verify_store(config: &StoreConfig, arity: usize) -> Result<()> {
+    let store_path = StoreConfig::data_path(&config.path, &config.id);
+    if !Path::new(&store_path).exists() {
+        return Err(anyhow!("Missing store file: {:?}", store_path));
```
dignifiedquire: you can write this easier I believe

```rust
ensure!(Path::new(&store_path).exists(), "Missing store file {}", store_path.display());
```

dignifiedquire: same for the other checks in this PR

PR #1039, opened 2020-02-20
feat(storage-proofs): speed up create_label

dignifiedquire:
- Improves perf of create_label by about 10-15% when used with sha intrinsics. 
- Introduces a new crate `sha2raw` which is optimized for this use case.
- Cleanup and unify `create_label` calls.
- Set core affinity during label generation.


storage-proofs/src/crypto/mod.rs
```diff
@@ -1,5 +1,5 @@
 pub mod aes;
-pub mod create_label;
+// pub mod create_label;
```
DrPeterVanNostrand: Remove this?

DrPeterVanNostrand:
Looks good to me too

deltazxm:
Slower 20% than before. Do you benchy it?@dignifiedquire

PR #1042, opened 2020-02-22
fix: update failed bench tests

cryptonemo:


fil-proofs-tooling/src/bin/benchy/shared.rs
```diff
@@ -203,6 +199,11 @@ pub fn create_replicas(
         .zip(priv_infos.into_iter())
         .zip(pub_infos.into_iter())
     {
+        println!(
+            "Inside create_replicas: Stored Replica Path {:?}, Exists? {}",
+            priv_info.replica_path(),
+            priv_info.replica_path().exists()
+        );
```
cryptonemo: The sealed files here are discarded after this method, which is the problem.  I'm not sure why.  The cache_dir still exists after.

cryptonemo: @dignifiedquire I tried many things trying to get this going, but without explicitly persisting the replica files (and removing them later), nothing was working locally.  This PR fixes the failed bench tests.  Please review/test.

merced: Sector sizes other than 2k are still failed in seal_pre_commit_phase2 with error of "Cannot compact with this configuration".

cryptonemo: @merced That's an unrelated issue, but I'll add a commit either here or open a new PR to show why.

cryptonemo:
@dignifiedquire @porcuquine Added tree validity checks and we see failures more clearly like this:

```
2020-02-23T14:11:29.437 INFO filecoin_proofs::api::seal > seal_pre_commit_phase1: start
2020-02-23T14:11:29.483 INFO filecoin_proofs::api::seal > building merkle tree for the original data
2020-02-23T14:11:29.483 TRACE filecoin_proofs::api::seal > seal phase 1: sector_size 16777216, tree size 1048575, tree leafs 524288, cached above base 7
2020-02-23T14:11:29.483 TRACE storage_proofs::merkle > create_merkle_tree called with size 524288
2020-02-23T14:11:29.635 INFO filecoin_proofs::api::seal > verifying pieces
2020-02-23T14:11:29.635 INFO filecoin_proofs::pieces > verifying 1 pieces
2020-02-23T14:11:29.635 INFO storage_proofs::stacked::proof > replicate_phase1
2020-02-23T14:11:29.635 INFO storage_proofs::stacked::proof > generate labels
2020-02-23T14:11:29.635 INFO storage_proofs::stacked::proof > generating layer: 1
2020-02-23T14:11:37.212 INFO storage_proofs::stacked::proof >   setting exp parents
2020-02-23T14:11:37.219 INFO storage_proofs::stacked::proof >   storing labels on disk
2020-02-23T14:11:37.219 TRACE storage_proofs::stacked::proof > is_merkle_tree_size_valid(524288, OCT_ARITY) = false
thread 'main' panicked at 'assertion failed: is_merkle_tree_size_valid(graph.size(), OCT_ARITY)', <::std::macros::panic macros>:2:4
```

This is a 16MiB sector size test.  The issue is that 16MiB would be fine as a leaf count, but it actually contains 524288 leaf nodes, which is not compatible with octrees.  Am I seeing this wrong?

fil-proofs-tooling/src/bin/benchy/shared.rs
cryptonemo: Updated :-)

porcuquine: @cryptonemo @dignifiedquire 

> This is a 16MiB sector size test. The issue is that 16MiB would be fine as a leaf count, but it actually contains 524288 leaf nodes, which is not compatible with octrees. Am I seeing this wrong?

That is the expected leaf count for 16 MiB, and it is true that this is not oct-tree compatible. Either we calculated these wrong in general, or a bad one slipped through. I think we should have a compile-time check so we cannot even generate parameters with incoherent sectors.

Reviewing the logic: 16MiB = 2^24 bytes. With 32-byte nodes, this is 2^19 leaf nodes. 19 % 3 != 0, so oct trees will not evenly cover this sector without a final compound tree.

porcuquine: NOTE: These responses belong to this comment of @cryptonemo's, but GItHub put them here: https://github.com/filecoin-project/rust-fil-proofs/pull/1042#issuecomment-590105475

I think the issue is that we calculated sector sizes based on quad trees initially, then moved to oct without recalculating.

For example: 2MiB sectors are good for quad (but not oct). 16MiB would be a valid extrapolation from 2MiB if that were valid for oct-trees, but it's not (because 16 % 3 != 0).

I'm hoping this issue only affects benchmarks and not sizes intended to be run in a network.

In addition to a compile-time check on sector coherence (as above), there should probably be a run-time error when trying to create such incoherent trees. (I don't know if there is or is not and am mostly responding to @cryptonemo's comment, not necessarily to the full context of this issue.)

dignifiedquire: 16MiB sectors are not supported anymore, you can see teh list of supported sectors here: https://github.com/filecoin-project/rust-fil-proofs/blob/master/filecoin-proofs/src/constants.rs#L11-L14

dignifiedquire:
works well, tested with 512MiB sectors

PR #1050, opened 2020-03-02
feat: add support for running multiple low priority threads

vmx:
Instead of having only a flag to run a single or two threads, change
the `parallel` option to a number, to define how many threads should
be run. The first thread/process is always a high priority one, the
other ones are low priority.

`1` means running only a single high priority proof (this is what
`--parallel false` was), `2` means running a high and a low priority
proof (this is what `--parallel true` was), higher numbers mean
one high priority proof and n-1 low priority ones.

This is the code I conducted my tests with for https://github.com/filecoin-project/bellman/pull/66#issuecomment-593459604.

### DIFF
--- a/fil-proofs-tooling/src/bin/gpu-cpu-test/main.rs
+++ b/fil-proofs-tooling/src/bin/gpu-cpu-test/main.rs
@@ -103,7 +103,7 @@ fn spawn_thread(
     (tx, handler)
 }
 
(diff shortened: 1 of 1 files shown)


vmx:
@dignifiedquire merge whenever you see fit. I'm happy to rebase it after the code re-org (that should be simple).

PR #1053, opened 2020-03-06
fix: quote dependence

gguoss:


storage-proofs/Cargo.toml
```diff
@@ -56,7 +56,7 @@ anyhow = "1.0.23"
 thiserror = "1.0.6"
 cpu-time = "1.0.0"
 neptune = "0.5.1"
-quote = "=1.0.2"
+quote = "1.0.2"
```
dignifiedquire: can be removed entirely, this was only an explicit dep as a workaround

dignifiedquire:
merged manually, thank you

PR #1054, opened 2020-03-06
feat: implement Winning and Window PoSt

dignifiedquire:
- [x] FallbackPoSt: Generic Compound Proof allowing for any number of sectors and challenges to be proven
- [x] Expose new api for Winning PoSt
- [x] Test for Winning PoSt
- [x] Expose new api for Window PoSt
- [x] Test for Window PoSt


filecoin-proofs/src/api/post.rs
```diff
@@ -477,7 +489,311 @@ pub fn verify_post(
...
+        partitions: None,
+        priority: false,
+    };
+    let pub_params: compound_proof::PublicParams<fallback::FallbackPoSt<DefaultTreeHasher>> =
+        fallback::FallbackPoStCompound::setup(&setup_params)?;
+
+    let verifying_key = get_post_verifying_key(&post_config)?;
+
+    let proof = MultiProof::new_from_reader(None, &proof[..], &verifying_key)?;
+    if proof.len() != 1 {
```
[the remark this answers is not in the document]
dignifiedquire: For now the information I have is to produce exactly one snark. Will remove if that changes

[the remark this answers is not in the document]
porcuquine: The proposal contains this: 
> Generate the SNARK proofsa using  WindowPoStChallenges:
> Set PartitionsAmount = RoundUp(len(SectorSet) / MaxSectorsPerSNARK)

I read this as suggesting that a `SectorSet` can contain more sectors than can be proved in a single SNARK.

storage-proofs/src/post/fallback/compound.rs
```diff
@@ -0,0 +1,361 @@
...
+
+        for (i, sector) in sectors.iter().enumerate() {
+            // 1. Inputs for verifying comm_r = H(comm_c || comm_r_last)
+            inputs.push(sector.comm_r.into());
+
+            // 2. Inputs for verifying inclusion paths
+            for n in 0..pub_params.challenge_count {
+                let challenge_index = (partition_index * pub_params.sector_count
+                    + i * pub_params.challenge_count
+                    + n) as u64;
```
porcuquine: I'm not sure this is right. Shouldn't it be `(partition_index * pub_params.sector_count + i) * pub_params.challenge_count + n`? Maybe I got that wrong, or maybe I'm missing something else — so let me also describe what I think you're trying to do. I think you're trying to assign a unique, sequential index to every challenge — starting from the first sector of the first partition, incrementing through all the challenges of that sector, and proceeding through the remaining sectors, then doing the same for each partition.

If that sounds right, what you have above seems off. I think the above will duplicate `challenge_index`es.

storage-proofs/src/post/fallback/compound.rs
```diff
@@ -0,0 +1,361 @@
...
+                    &por_pub_params,
+                    partition_k,
+                )?;
+
+                inputs.extend(por_inputs);
+            }
+        }
+        let num_inputs_per_sector = inputs.len() / sectors.len();
+
+        // dupliate last one if too little sectors available
```
porcuquine: typo: 'duplicate'

storage-proofs/src/post/fallback/vanilla.rs
```diff
@@ -0,0 +1,571 @@
...
+    Ok(challenged_range_index)
+}
+
+impl<'a, H: 'a + Hasher> ProofScheme<'a> for FallbackPoSt<'a, H> {
+    type PublicParams = PublicParams;
+    type SetupParams = SetupParams;
+    type PublicInputs = PublicInputs<'a, H::Domain>;
+    type PrivateInputs = PrivateInputs<'a, H>;
+    type Proof = Proof<H>;
+    type Requirements = NoRequirements;
```
porcuquine: I think this should actually be `ChallengeRequirements` not `NoRequirements`. Similar to PoRep, we need to ensure that a set of provided PoSt circuit proofs actually contains enough proofs.

I know that is being checked in a different way here in the vanilla proof, but we need to do it also in the circuit `MultiProof` case, and that is the intended purpose of `ChallengeRequirements`.

filecoin-proofs/src/api/post.rs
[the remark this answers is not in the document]
dignifiedquire: yes, that is handled now

filecoin-proofs/src/constants.rs
```diff
@@ -13,8 +13,13 @@ pub const SECTOR_SIZE_8_MIB: u64 = 1 << 23;
 pub const SECTOR_SIZE_512_MIB: u64 = 1 << 29;
 pub const SECTOR_SIZE_32_GIB: u64 = 1 << 35;
 
-pub const POST_CHALLENGE_COUNT: usize = 65;
-pub const POST_CHALLENGED_NODES: usize = 1;
+pub const ELECTION_POST_CHALLENGE_COUNT: usize = 65;
+pub const ELECTION_POST_CHALLENGED_NODES: usize = 1;
+
+pub const WINNING_POST_CHALLENGE_COUNT: usize = 66;
```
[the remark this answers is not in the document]
dignifiedquire: @nicola told me to switch to 66. Folks can't make up their mind..

filecoin-proofs/src/constants.rs
```diff
@@ -55,6 +60,17 @@ lazy_static! {
         .copied()
         .collect()
     );
+    pub static ref WINDOW_POST_SECTOR_COUNT: RwLock<HashMap<u64, usize>> = RwLock::new(
+        [
+            (SECTOR_SIZE_2_KIB, 2),
+            (SECTOR_SIZE_8_MIB, 2),
+            (SECTOR_SIZE_512_MIB, 2),
+            (SECTOR_SIZE_32_GIB, 32) // TODO: correct value
```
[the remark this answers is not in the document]
dignifiedquire: correct, I will need to do measurements and update once I have run those

storage-proofs/src/post/fallback/compound.rs
dignifiedquire: cood catch, fixing

storage-proofs/src/post/fallback/vanilla.rs
dignifiedquire: fixed, though I am not sure it does verify everything we need,

storage-proofs/src/post/fallback/compound.rs
dignifiedquire: fixed

dignifiedquire:
> This question still stands. Is there a reason we don't put all this into the PoStConfigs then treat the different PoSt types uniformly?

There is, I do not want that number to be set from the outside.

filecoin-proofs/src/constants.rs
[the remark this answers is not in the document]
dignifiedquire: measured and updated

filecoin-proofs/src/constants.rs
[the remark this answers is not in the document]
porcuquine: I have added comments in the doc, requesting clarification. This PR is either blocked on confirmation or else finalization of the work is, in which case, please add a clearly-marked TODO and plan on another version to be required before we release this.

storage-proofs/src/post/fallback/vanilla.rs
porcuquine: I don't think you updated `POREP_MINIMUM_CHALLENGES` in `constants.rs` can you confirm that 138 is the correct number of minimum challenges for 32GiB? (cc: @nicola)

What else do you think needs to be verified here that might be missing? Let's think it through and extend the relevant implementation of `ProofScheme::satisfies_requirements` (and `ChallengeRequirements`, if necessary) to ensure everything necessary *is* verified.

porcuquine: UPDATE 2: I see now that you did implement `satisfies_requirements`, but you must not ignore the number of partitions in the input — since that is what needs to be verified.

filecoin-proofs/src/api/post.rs
```diff
@@ -662,7 +662,7 @@ pub fn verify_winning_post(
         &pub_inputs,
         &proof,
         &fallback::ChallengeRequirements {
-            challenge_count: param_challenge_count,
+            challenge_count: post_config.challenge_count * replicas.len(),
```
porcuquine: I would slightly rather see this using `post_config.sector_count` than `replicas.len()` (although I understand you have asserted them to be equal above). The former expresses exactly what we mean to require here.

storage-proofs/src/post/fallback/vanilla.rs
```diff
@@ -443,12 +444,10 @@ impl<'a, H: 'a + Hasher> ProofScheme<'a> for FallbackPoSt<'a, H> {
...
         requirements: &Self::Requirements,
-        _partitions: usize,
+        partitions: usize,
     ) -> bool {
-        dbg!(&requirements);
-        dbg!(&public_params);
-        // Every partition must have the exact same number of challenges.
-        requirements.challenge_count == public_params.challenge_count
+        partitions * public_params.sector_count * public_params.challenge_count
+            >= requirements.challenge_count
```
porcuquine: If we use `>=` here, then we *should* call it `minimum_challenges` (as in PoRep). If we call it `challenge_count` we should require that it be exactly `==`.

storage-proofs/src/post/fallback/vanilla.rs
dignifiedquire: updated again

storage-proofs/src/post/fallback/vanilla.rs
dignifiedquire: fixed

filecoin-proofs/src/api/post.rs
dignifiedquire: fixed

PR #1060, opened 2020-03-17
fix(sha2raw): use runtime detection for sha-ni correctly

dignifiedquire:


cryptonemo:
Looks good (except CI and that pesky commit message ;-)

PR #1063, opened 2020-03-23
refactor(filecoin-proofs): pad_reader::PadReader fr32_reader::Fr32Reader

rvagg:
For clarity, since padding input data to achieve correct ^2 output size prior to CommP calculation is also required.

Ref #1061 

tbh I've been messing with and around Fr32 padding for a couple of months now but I still don't know what it stands for, is it an acronym? Therefore, I don't know for sure that using the name "Fr32Reader" makes as much sense as it could.

Is this "feat", "fix", "chore", or something else? It's a breaking change but it was only exposed publicly a couple of days ago so maybe a "fix"?

### DIFF
--- a/filecoin-proofs/src/lib.rs
+++ b/filecoin-proofs/src/lib.rs
@@ -6,7 +6,7 @@ mod commitment_reader;
 
 pub mod constants;
 pub mod fr32;
-pub mod pad_reader;
+pub mod fr32_reader;
 pub mod param;
 pub mod parameters;
 pub mod pieces;
@@ -17,7 +17,6 @@ pub mod types;
 pub use self::api::*;
 pub use self::commitment_reader::*;
 pub use self::constants::SINGLE_PARTITION_PROOF_LEN;
-pub use self::pad_reader::*;
 pub use self::param::{ParameterData, ParameterMap};
 pub use self::types::*;
 
--- a/filecoin-proofs/benches/preprocessing.rs
+++ b/filecoin-proofs/benches/preprocessing.rs
@@ -2,6 +2,7 @@ use std::io::{self, Read};
 use std::time::Duration;
 
 use criterion::{criterion_group, criterion_main, Criterion, ParameterizedBenchmark, Throughput};
+use filecoin_proofs::fr32_reader::Fr32Reader;
 use rand::{thread_rng, Rng};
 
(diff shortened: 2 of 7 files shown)


filecoin-proofs/src/lib.rs
```diff
@@ -17,7 +17,6 @@ pub mod types;
 pub use self::api::*;
 pub use self::commitment_reader::*;
 pub use self::constants::SINGLE_PARTITION_PROOF_LEN;
-pub use self::pad_reader::*;
```
rvagg: Not exactly sure what the purpose of this line is, doesn't _seem_ to be needed, tests pass and I can use `fr32_reader::Fr32Reader` in an external library. Maybe I should have removed it in #1061.

dignifiedquire:
> Is this "feat", "fix", "chore", or something else? It's a breaking change but it was only exposed publicly a couple of days ago so maybe a "fix"?

I'll probably merge it as `"refactor"`

dignifiedquire:
@rvagg thank you, looks like the benchmarks need some updates too according to CI.

dignifiedquire:
> tbh I've been messing with and around Fr32 padding for a couple of months now but I still don't know what it stands for, is it an acronym? Therefore, I don't know for sure that using the name "Fr32Reader" makes as much sense as it could.

It is not a great name `Fr` comes from the fact that the finite field elements that we are referring to are called `Fr` in pairing: https://github.com/filecoin-project/pairing/blob/master/src/bls12_381/fr.rs#L11

The `32` comes from the fact that the size we are padding them to is 32bytes

porcuquine:
Just a little more context on the name. This won't actually make any more sense of the name itself, but the reason we *have* a name at all is that we created a type alias with the intent that it should only ever be used for values which are known to fit in the field. This could because a) they were checked; b) they were truncated to guarnatee it; c) they were padded/extended from something short enough to guarantee it; or d) they were produced as a result of a field operation guaranteed to return field elements. This is not enforced, but it is a program-level rule.

rvagg:
ahhhh

> In zk-SNARK schemes, we need to manipulate very large polynomials over the scalar field F<sub>r</sub>.

👍 

Pushed a fixup! that fixes the benchmarks and a variable name.
Will let you handle the squash and commit msg rename.

Issue #1067, opened 2020-03-25
Simplify `BucketSample[n, m]` construction


schomatis:
Described in page 22 of this [TR](https://web.stanford.edu/~bfisch/porep_short.pdf), this construction expands on the [`DRSample`](https://eprint.iacr.org/2017/443.pdf) algorithm to allow in-degrees bigger than two (the immediate predecessor and one random parent) by increasing the number of nodes and clustering them in groups of `m`, where `m` is the target in-degree. Over that meta-graph the original `DRSample` is called (`BucketSample[n]` in the TR, which only deals with the random parent, the immediate predecessor seems implicit in the construction) to produce many parents for the same meta-node.

This expanded algorithm has two issues in my view:

1. It's overly complicated, it introduces a new layer of abstraction (the meta-graph and the logic associated with navigating these expanded meta-nodes).

2. It allows a new undesired edge case that `DRSample` does not suffer from: a node can have itself as its parent. This happens when one of the nodes in the expanded meta-node is (randomly) assigned another node (close in distance) in its same meta-node (blue edge in this [diagram](https://i.imgur.com/ALt4fMx.png)).

Proposed solution: simplify `BucketSample[n, m]` as just a `m` repetitive calls to `BucketSample[n]` (`DRSample`). Some observations:

* `BucketSample[n, m]`/`BucketSample[n]` should then be converted to a call that doesn't generate the entire graph (of size `n`) but just the parents of a particular `i` node. This is already what our current implementation does (generate parents for only one node per call as needed).

* We [inject the node index `i` in the seed](https://github.com/filecoin-project/rust-fil-proofs/blob/14870d715f1f6019aba3f72772659e38184378bf/storage-proofs/src/drgraph.rs#L185) to have deterministic parents independent of call order (due to the way we implement `BucketSample[n, m]`, see point above). We won't need to inject the index [`k`](https://github.com/filecoin-project/rust-fil-proofs/blob/14870d715f1f6019aba3f72772659e38184378bf/storage-proofs/src/drgraph.rs#L188) of the nodes in the meta-node because we always request *all* the parents of a node at once, so the `m` *sequential* calls to `DRSample` will always have the same outcome.

[...]

PR #1069, opened 2020-03-25
fix: DRG parent generation

### LINKED ISSUE #1065: Fix the off-by-one error in the number of buckets in the `BucketGraph`
### Description

The number of buckets in the bucket graph should be `ceil(log2(node))` not [`floor(log2(node))`](https://github.com/filecoin-project/rust-fil-proofs/blob/master/storage-proofs/src/drgraph.rs#L192). This issue only affects `BucketGraph::parents()`. The variable names used during parent generation should also be cleaned up.

DrPeterVanNostrand:
This PR makes the following changes to DRG parent generation:

- fixes an off by one error in the number of buckets
- removes unnecessary and dead code related to parent number addition
- fixes bugs introduced by improper casting
- makes explicit the type differences between DRG node indexes `u32` and metagraph node indexes `u64`, neither should be `usize`
- replaces `rng.gen_range()`
- changes the rng from `ChaCha20` to `ChaCha8`
- clarifies variable names

Closes #1065, #1067

### DIFF
--- a/storage-proofs/src/drgraph.rs
+++ b/storage-proofs/src/drgraph.rs
@@ -6,7 +6,7 @@ use anyhow::ensure;
 use generic_array::typenum;
 use merkletree::store::StoreConfig;
 use rand::{rngs::OsRng, Rng, SeedableRng};
-use rand_chacha::ChaChaRng;
+use rand_chacha::ChaCha8Rng;
 use sha2::{Digest, Sha256};
 
 use crate::error::*;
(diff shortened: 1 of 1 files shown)


schomatis:
As discussed this change makes sense to me and the comments and renaming of the variables are very welcome, nice, just two requests:

1. Decouple the logic change from the refactoring in the commit to make sure we don't inadvertently change anything else. The idea is to have an isolated commit with the floor/ceil change that explains why we're deviating from the theoretical construction.

2. Add a simple test (like the basic example you showed me to illustrate your point) to make sure this has the desired effect.

nicola:
I reviewed this with [these notes](https://hackmd.io/i8wDaOLZSiCauDKyaM5KOg) which seem to have the same conclusions.

This review is a review of the change and not the correctness of the algorithm as a whole. I will need longer time to do this.

- In R2, I did note the floor error and I added a +1. I think that using ceil is the right fix.
- On the deletion of k, it seems correct.

Ok to merge, more review will be needed in terms of algorithm correctness.

nicola:
One way to gain confidence is to re-run the attacks.

PR #1070, opened 2020-03-26
fix: check updated store path on verify after precommit phase1

cryptonemo:


filecoin-proofs/src/api/mod.rs
```diff
@@ -314,7 +314,15 @@ where
         .labels
         .verify_stores(verify_store, &cache)?;
```
cryptonemo: The labels path here also needs to be updated, or I can modify the verify to only use the provided cache dir, rather than cross checking.  I'll look at the latter option for now.

storage-proofs/src/porep/stacked/vanilla/params.rs
```diff
@@ -504,17 +504,10 @@ impl<H: Hasher> Labels<H> {
...
-            let current_path = StoreConfig::data_path(&label.path, &label.id);
-            if !current_path.starts_with(&cache_dir) {
-                return Err(anyhow!(
-                    "Label path mis-match [label vs cache_dir: {:?}/{:?}]",
-                    current_path,
-                    cache_dir
-                ));
-            }
+        let updated_path_labels = self.labels.clone();
+        for mut label in updated_path_labels {
```
dignifiedquire: there is already `fn update_root<P: AsRef<Path>>(&mut self, root: P)` down below for exactly this

cryptonemo: Yes, if we clone the entire `Labels` object, it could be re-used.  This seemed simpler in this case.

dignifiedquire: no need to clone, you could just change this function signature to take `&mut self` and call `update_root` as the first step

cryptonemo: Ok.  We also update_root in the next phase.

dignifiedquire: Or alternatively call `update_root` manually before calling this. Otherwise this gives a false sense of correctness, as this checks against a different cache path that the struct that the caller actually holds

cryptonemo: Just to be clear, the `validate_cache_for_precommit_phase2` will change signature to take a mutable `SealPreCommitPhase1Output`, correct?

dignifiedquire: hmmmmmmmmm

cryptonemo: > Otherwise this gives a false sense of correctness, as this checks against a different cache path that the struct that the caller actually holds

Agreed, that's kind of the point of being able to set the cache_path though.  We do more or less ignore what's in the output, in favor of the updated path (e.g. in the next phase)

dignifiedquire: Your call, I dislike all solutions proposed, pick what you think is the least bad.

cryptonemo: I also dislike the solutions, but if I'm understanding correctly, the phases are doing the right thing, but the validate is not because it is not respecting the updated cache path.  By being able to specify a path at all in the validate, it seems it should be used (always) in favor of whatever's stored, as the caller may have even locally moved the data.  We can chat about briefly on our call today, but I think as-is should be correct.  The previous code ensured that the cache path matched what was in the phase output, but we know that won't be the case when files have moved, so I'd say that was incorrect(ly based on non-moving assumptions).

dignifiedquire: alright, lets ship it

cryptonemo: Not opposed, but would like to hear from @magik6k if it resolves the issue.  Just want to be sure I didn't miss anything.  If it's easier to test from master, let's ship.

cryptonemo:
I prefer this to #1071, but it's also available as an option.

cryptonemo:
@dignifiedquire Confirmed to resolve the initial issue.  I closed #1071 since it sounded like this one is better for now.  Please review/approve/merge when CI is happy (rebased against latest).

PR #1071, opened 2020-03-26
fix: check updated store path on verify after precommit phase1

cryptonemo:
Alternative to https://github.com/filecoin-project/rust-fil-proofs/pull/1070

Updates phase1 output in place by setting the new cache_dir specified during the validate (in addition to the precommit phase 2, as validate isn't required to be called).


cryptonemo:
I think this path will not be pursued in favor of #1070

PR #1077, opened 2020-04-07
refactor: split storage-proof into subcrates

dignifiedquire:
This helps compile times and makes dev cycles quicker.

There are now
- storage-proofs-core - everything shared
- storage-proofs-porep - all poreps
- storage-proofs-post - all posts
- storage-proofs - same api as before, reexports the others


vmx:
Out of interest. Was this all manual work or are there tools to do something like "fix my imports"?

dignifiedquire:
> Out of interest. Was this all manual work or are there tools to do something like "fix my imports"?

All manual labor, though I hear that rust-analyzer has some import fixing tools.

PR #1078, opened 2020-04-07
publish v25 params

dignifiedquire:


filecoin-proofs/parameters.json
```diff
@@ -1,82 +1,122 @@
...
     "sector_size": 536870912
   },
-  "v24-proof-of-spacetime-election-PoseidonHasher-5916054ae98e28fc2f0470d1fb58eb875a6865be86f0b8c4e302d55f13217fef.params": {
-    "cid": "QmSXMF85mdGLQfAY98zVL4dUBpGPFFUPDmFzdc1NZrVFdh",
-    "digest": "a93de0f8cfb04af5d21f66ef48ee59a8",
+  "v25-proof-of-spacetime-fallback-MerkleTree<PoseidonHasher, 8, 0, 0>-3ea05428c9d11689f23529cde32fd30aabd50f7d2c93657c1d3650bca3e8ea9e.params": {
+    "cid": "QmQ2HrKCWbtWQNNQiBj3BFE8QrqMyed8P5Vw5vyyzuSMsF",
+    "digest": "2e15ec3fbff51abf66d241252fb8babd",
     "sector_size": 2048
   },
```
porcuquine, line 26: This one and the one above (""v25-proof-of-spacetime-fallback-MerkleTree<PoseidonHasher, 8, 0, 0>-0170db1f394b35d995252228ee359194b13199d259380541dc529fb0099096b0.params") seem to be duplicated but for the digest part of the cache key. Is it possible that two different circuit-code versions got included in this manifest?

dignifiedquire: I believe this is correct, there are two fallback posts per sector size, one for window and one for winning.

dignifiedquire: There should be and there are 24 entrie in total

- 4 sector sizes
- 3 different proofs per sector size
- 2 files (`.params`, `.vk`) per proof

porcuquine: Okay. They really should have distinct names, in my opinion.

porcuquine: By that, I mean there should be enough metadata surfaced in their cache keys to make it clear which is which.

dignifiedquire: they use the same circuit, so this is not easily possible

porcuquine: Don't they differ in public parameters, though (they should)? The cache identifier is a function of public parameters for that reason, so it seems like it should be possible. What am I missing?

PR #1081, opened 2020-04-08
feat(winning-post): avoid passing the full list of sectors

dignifiedquire:
This was a request from the node implementations, to make this interface more performant. The burden of checking the correct sectors is now on the callers.


porcuquine:
In particular, this review doesn't give any attention to the question of whether this change of behavior is a good idea or whether it might violate some other assumption about the proofs taken as a whole. So this needs to be considered at the protocol/spec/security level also. (I'm not saying it *is* an issue, just that I'm only approving this as a code change and don't think such a PR would be the place to endorse anything else either.) Can we please track this decision elsewhere? (cc: @anorth @nicola )

PR #1083, opened 2020-04-13
ci: fix file descriptor limits on macOS

ognots:
fixes errors in CircleCI mac builds, such as
```
---- fallback::vanilla::tests::fallback_post_poseidon_two_partitions_smaller_top_8_8_2 stdout ----
thread 'fallback::vanilla::tests::fallback_post_poseidon_two_partitions_smaller_top_8_8_2' panicked at 'called `Result::unwrap()` on an `Err` value: Too many open files (os error 24)', /Users/distiller/crate/storage-proofs/core/src/merkle/builders.rs:411:13
```

Fix sets ulimit for system, users (`sudo ulimit` calls) and then user sets ulimit (without `sudo`)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -278,7 +278,8 @@ jobs:
       - run: rustup default $(cat rust-toolchain)
       - run: cargo update
       - run: cargo fetch
-      - run: ulimit -n 512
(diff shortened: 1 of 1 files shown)


dignifiedquire:
Thank you

PR #1085, opened 2020-04-27
feat: add some more timing results to window post bench

cryptonemo:
Outputs some additional timings.

### DIFF
--- a/fil-proofs-tooling/src/bin/benchy/window_post.rs
+++ b/fil-proofs-tooling/src/bin/benchy/window_post.rs
@@ -34,6 +34,18 @@ struct Inputs {
 #[derive(Serialize)]
 #[serde(rename_all = "kebab-case")]
 struct Outputs {
+    seal_pre_commit_phase1_cpu_time_ms: u64,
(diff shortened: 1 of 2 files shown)


fil-proofs-tooling/src/bin/benchy/window_post.rs
```diff
@@ -105,49 +117,67 @@ pub fn run_window_post_bench<Tree: 'static + MerkleTreeTrait>(
...
+            cache_dir.path(),
+            staged_file.path(),
+            sealed_file.path(),
+            PROVER_ID,
+            sector_id,
+            TICKET_BYTES,
+            &piece_infos,
+        )
+    })
+    .expect("failed in seal_pre_commit_phase1");
```
vmx, line 132: This function returns already an anyhow result, hence you could use `.context()?` instead of `.expect()` which then would return an error instead of panicking. The applies to all uses of `.expect()`.

dignifiedquire:
merging, non critical code, we can always improve, thansk @cryptonemo

PR #1086, opened 2020-04-27
feat: lowercase and remove special chars on parameter files

cryptonemo:
Updates name output where used in parameter file path generation.

### DIFF
--- a/storage-proofs/core/src/hasher/poseidon.rs
+++ b/storage-proofs/core/src/hasher/poseidon.rs
@@ -28,7 +28,7 @@ impl Hasher for PoseidonHasher {
     type Function = PoseidonFunction;
 
     fn name() -> String {
-        "poseidonhasher".into()
+        "poseidon_hasher".into()
     }
 
     #[inline]
--- a/storage-proofs/core/src/hasher/poseidon.rs
+++ b/storage-proofs/core/src/hasher/poseidon.rs
@@ -28,7 +28,7 @@ impl Hasher for PoseidonHasher {
     type Function = PoseidonFunction;
 
     fn name() -> String {
-        "PoseidonHasher".into()
+        "poseidonhasher".into()
     }
 
     #[inline]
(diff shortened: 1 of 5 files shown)


storage-proofs/core/src/hasher/sha256.rs
```diff
@@ -24,7 +24,7 @@ impl Hasher for Sha256Hasher {
     type Function = Sha256Function;
 
     fn name() -> String {
-        "Sha256Hasher".into()
+        "sha256hasher".into()
```
dignifiedquire, line 27: lets use `sha256_hasher` for better readability

storage-proofs/core/src/hasher/poseidon.rs
```diff
@@ -28,7 +28,7 @@ impl Hasher for PoseidonHasher {
     type Function = PoseidonFunction;
 
     fn name() -> String {
-        "PoseidonHasher".into()
+        "poseidonhasher".into()
```
dignifiedquire, line 31: lets use `poseidon_hasher` for better readability

dignifiedquire:
sad CI is sad

cryptonemo:
> sad CI is sad

CI doesn't like that we renamed parameter files on it ;-)

cryptonemo:
> > sad CI is sad
> 
> CI doesn't like that we renamed parameter files on it ;-)

@dignifiedquire What's the proper way to get new params generated on CI for this kind of thing?  The test *should* pass when the params are found.  Also, are there any other changes required before params are regenerated, and if so, maybe those could be merged here so we can do them all at once?

dignifiedquire:
@cryptonemo you need to bump the cache versions. Both in circleci and the main cache identifier, as this is a breaking change as far as the paramchache is concerend

cryptonemo:
@dignifiedquire There's still something here I'm not understanding :-(

PR #1087, opened 2020-04-29
feat(storage-proofs): integrate neptune's column tree builder

cryptonemo:


storage-proofs/porep/src/stacked/vanilla/proof.rs
```diff
@@ -480,66 +725,103 @@ impl<'a, Tree: 'static + MerkleTreeTrait, G: 'static + Hasher> StackedDrg<'a, Tr
...
-                info!("building base tree_c {}/{}", i + 1, tree_count);
-                trees.push(DiskTree::<
-                    Tree::Hasher,
-                    Tree::Arity,
-                    typenum::U0,
-                    typenum::U0,
-                >::from_par_iter_with_config(
-                    hashes.into_par_iter(), config.clone()
-                )?);
+        let tree_c_root = match layers {
```
cryptonemo, line 728: @dignifiedquire If you get a chance, I could use some help making this part manageable.  The double layer matching is cumbersome at best ...

.circleci/config.yml
```diff
@@ -26,7 +26,7 @@ jobs:
           no_output_timeout: 30m
       - restore_cache:
           keys:
-            - cargo-v26b-{{ checksum "rust-toolchain" }}-{{ checksum "Cargo.toml" }}-{{ checksum "Cargo.lock" }}-{{ arch }}
+            - cargo-v27a-{{ checksum "rust-toolchain" }}-{{ checksum "Cargo.toml" }}-{{ checksum "Cargo.lock" }}-{{ arch }}
```
dignifiedquire, line 29: lets stay at 26, just go to 26c, also this shouldn't change the groth params, or am I missing sth?

cryptonemo: I tried 26c, but it didn't work (missing params), so I bumped to v27 just to see if I could get that working.  The params are updated because this PR bumps some of the test sectors to have min of 2 layers (instead of 1 for the smaller test ones).

cryptonemo:
@porcuquine These appear to be real errors (I can reproduce locally) -- would any of this work (layer changes, params, etc) cause these to change?  If not, I'll need some help better understanding what's going on there. https://circleci.com/gh/filecoin-project/rust-fil-proofs/45078?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

.circleci/config.yml
dignifiedquire: we only need to change the full version if we switch from already published ones, because v26 is not yet published, there is no need to change that version yet

filecoin-proofs/src/bin/paramcache.rs
```diff
@@ -24,6 +24,7 @@ const PUBLISHED_SECTOR_SIZES: [u64; 10] = [
     SECTOR_SIZE_16_MIB,
     SECTOR_SIZE_512_MIB,
     SECTOR_SIZE_1_GIB,
+    SECTOR_SIZE_4_GIB,
```
dignifiedquire, line 27: why are we publishing this size?

storage-proofs/porep/Cargo.toml
```diff
@@ -27,10 +27,15 @@ pretty_assertions = "0.6.1"
...
 
 [dev-dependencies]
 tempdir = "0.3.7"
 tempfile = "3"
 rand_xorshift = "0.2.0"
+
+[features]
+default = ["column-builder-gpu"]
+#default = []
+column-builder-gpu = []
```
dignifiedquire, line 41: given that we always build this, we might as well not add a feature

storage-proofs/porep/src/stacked/vanilla/proof.rs
```diff
@@ -356,6 +366,243 @@ impl<'a, Tree: 'static + MerkleTreeTrait, G: 'static + Hasher> StackedDrg<'a, Tr
...
+                    );
+                    let mut columns: Vec<GenericArray<Fr, ColumnArity>> =
+                        vec![
+                            GenericArray::<Fr, ColumnArity>::generate(|_i: usize| Fr::zero());
+                            chunked_nodes_count
+                        ];
+
+                    rayon::scope(|s| {
+                        // spawn n = num_cpus * 2 threads
+                        let n = num_cpus::get() * 2;
```
dignifiedquire, line 453: `num_cpus::get()` already returns hyperthreaded, so probably we should not multiply by 2

filecoin-proofs/src/bin/paramcache.rs
cryptonemo: This was just to help GPU testing 4GiB trees.  A bit quicker than 32GiB ;-)  I don't mind either way if it stays or goes though.

storage-proofs/porep/Cargo.toml
cryptonemo: Right, that changed -- I'll update that.

PR #1088, opened 2020-04-29
fix(storage-proofs-porep): add missing public input for replica-id

dignifiedquire:
Thanks to @starli-trapdoor for pointing this out.

TODOs

- [x] Decide if we need `to_bits_le` or `to_bits_le_strict`


gakonst:
How could this be exploited?

woodstock-xkx:
The attacker can provide arbitrary `replica_id_bits` as these variables are not constrainted by public input `replica_id`.

StarLI-Trapdoor:
@gakonst If replica_id is NOT public, the current circuit only proves that one sector calculation (SDR) is "constrained" by some replica_id, not a specified replica_id. Simply put, one fixed replica_id can be used for SDR calculation. That's to say, all sectors can share one SDR replica.

storage-proofs/porep/src/stacked/circuit/params.rs
```diff
@@ -63,57 +84,164 @@ impl<Tree: MerkleTreeTrait, G: 'static + Hasher> Proof<Tree, G> {
...
-                    raw.map(Into::into)
-                        .ok_or_else(|| SynthesisError::AssignmentMissing)
-                })?;
+        // stores the labels of the challenged column
+        let mut column_labels = Vec::new();
 
-            proof.synthesize(
-                cs.namespace(|| format!("labeling_proof_{}", layer)),
+        // PrivateInput: challenge index
+        let challenge_num = uint64::UInt64::alloc(cs.namespace(|| "challenge"), challenge)?;
```
porcuquine, line 158: Why is this private? This must correspond to the public input for a given challenge, but I don't see where that is enforced. I think we either need to use the supplied public input or add a constraint enforcing equality.

porcuquine: I pushed a commit which I believe addresses this issue. cc: @dignifiedquire @nicola

nicola: yes, @porcuquine is correct - it must be a public input

storage-proofs/porep/src/stacked/circuit/params.rs
```diff
@@ -154,8 +154,9 @@ impl<Tree: MerkleTreeTrait, G: 'static + Hasher> Proof<Tree, G> {
         // stores the labels of the challenged column
         let mut column_labels = Vec::new();
 
-        // PrivateInput: challenge index
+        // PublicInput: challenge index
         let challenge_num = uint64::UInt64::alloc(cs.namespace(|| "challenge"), challenge)?;
+        challenge_num.pack_into_input(cs.namespace(|| "challenge input"))?;
```
nicola, line 159: @porcuquine why are you using pack_into_input (?), you should just call inputize like we did on other parts

porcuquine: `UInt64` is represented as allocated `Boolean`s, and we need to constrain them to decompose a value passed as `Fr` input. This accomplishes that. If `UInt64` were represented by an allocated `Num`, then we would use `inputize`. It's just one constraint, if that's what you're concerned about.

StarLI-Trapdoor:
More explanation about the vulnerability: 

https://medium.com/@starli/filecoin-one-porep-vulnerability-found-by-trapdoor-tech-7fc7beb4557b

Issue #1079, opened 2020-04-08
Improve parampublish


dignifiedquire:
Currently `parampublish` simply publishes all files it finds in the cache directory when passing the `--all` flag. 

What is needed

- scan the directory
- detect the different versions available, ask the user to select which version to publish
- verify that for each sector size there are exactly 3 matching files for each proof, one `.vk` and one `.params` and one `.meta` file
- fixed list of sector sizes should be there, which all need to be present to publish, publishing by all or sector size should be disable
  - this list should be as of now
    - SECTOR_SIZE_2_KIB,
    - SECTOR_SIZE_8_MIB,
    - SECTOR_SIZE_512_MIB,
    - SECTOR_SIZE_32_GIB,
    - SECTOR_SIZE_64_GIB,

porcuquine:
Consider also checking timestamps and requiring an extra check if a user wants to use the older version.

Also considering extending paramcache to capture git commit information as metadata (perhaps in a directory-local manifest) in order to help know which parameter file corresponds to which code. Otherwise it may be hard to know which version should actually be used when this situation arises (without just deleting everything and regenerating).

dignifiedquire:
Closing as the most important things were done.

PR #1090, opened 2020-05-01
feat(parampublish): Select version

vmx:
With the `--all` flag you now get a select menu to select which version
of parameters you want to publish.

That's just a part of https://github.com/filecoin-project/rust-fil-proofs/issues/1079, but it's a start.
I had hoped I would get further, but it makes sense to
do it properly, which takes longer than expected.


dignifiedquire:
nice, seems CI is broken though

vmx:
I'm closing this one as I split the commits in different ways and push it to https://github.com/filecoin-project/rust-fil-proofs/pull/1092.

PR #1092, opened 2020-05-01
feat(filecoin-proofs): improve parampublish

vmx:
This PR fixes parts of https://github.com/filecoin-project/rust-fil-proofs/issues/1079.

I originally misunderstood some of the requirements, hence it got fancier then probably anticipated. You can now select the sizes you want to publish, with having the defaults pre-selected.

This PR can also be reviewed commit per commit, I tried to make them self-contained (the commit message should explain some of the reasons for some of the changes).

@DrPeterVanNostrand I've added you for review as I updated dialoguer, which also needed a little change in the phase2 bin code. I just want to make sure you are aware of that change, in case something breaks (which I won't expect).

This PR supersedes #1090.


vmx:
Yay, all tests passed. Ready for review.

PR #1093, opened 2020-05-01
[WIP] improve sdr circuit

dignifiedquire:


storage-proofs/porep/src/stacked/circuit/proof.rs
```diff
@@ -257,6 +254,9 @@ impl<'a, Tree: 'static + MerkleTreeTrait, G: 'static + Hasher>
 
             // final replica layer
             inputs.extend(generate_inclusion_inputs(challenge)?);
+
+            // c_x
+            inputs.extend(generate_inclusion_inputs(challenge)?);
         }
```
porcuquine, line 260: Why is this duplicated as both 'final replica layer' and 'c_x'? If these inputs really are guaranteed identical, then we should be able to reuse the inputs (which is more economical).

dignifiedquire: One is an inclusion proof in comm_r_last and one in comm_c, so they can not be combined asfaict

dignifiedquire:
Merged manually

PR #1095, opened 2020-05-02
feat: update cache to allow 32GiB and 64GiB sectors

cryptonemo:


storage-proofs/porep/src/stacked/vanilla/graph.rs
```diff
@@ -177,10 +177,10 @@ where
...
         if use_cache {
             info!("using parents cache of unlimited size");
-            assert_eq!(
-                nodes,
-                (32 * 1024 * 1024 * 1024) / NODE_SIZE,
-                "Cache is only available for 32GiB sectors"
+            const GIB: usize = 1024 * 1024 * 1024;
+            ensure!(
+                ((nodes * NODE_SIZE == 32 * GIB) || (nodes * NODE_SIZE == 64 * GIB)),
+                "Cache is only available for 32GiB and 64GiB sectors"
```
dignifiedquire, line 183: This won't work, because we have a global cache, and the same program could attempt to seal both 32 and 64 GIB sectors, need to create a second static variabel for this other cache.

Alternatively we could forbid that, and reuse the cache, but then we need to check the size of the existing cache probably

cryptonemo: I added something for this, let me know if something like that will work or if I'm still missing something there.

dignifiedquire: looks good now

PR #1098, opened 2020-05-03
fix(filecoin-proofs): use 2 layers for 8 and 16 MiB sectors

dignifiedquire:


dignifiedquire:
@cryptonemo please check if this is okay, or if we need to whitelist 8 layer column hashes

cryptonemo:
> @cryptonemo please check if this is okay, or if we need to whitelist 8 layer column hashes

I think it should be fine.  I had bumped those up to 8 only because 2/8/11 are the supported GPU arities.  Moving them back to 2 should do no harm.

PR #1100, opened 2020-05-07
feat: integrate filecoin-proofs fuzzers

cryptonemo:


cryptonemo:
A couple of the commands are stubbed in and incomplete, as it appears that they will require some more low level types across dependencies to support the arbitrary derive. Examples include: `generate_window_post` and `seal_commit_phase2`.

PR #1101, opened 2020-05-11
fix: use single replica file when using sub-trees

cryptonemo:


storage-proofs/porep/src/stacked/vanilla/params.rs
```diff
@@ -503,15 +503,16 @@ impl<Tree: MerkleTreeTrait, G: Hasher> TemporaryAuxCache<Tree, G> {
             DiskTree<Tree::Hasher, Tree::Arity, Tree::SubTreeArity, Tree::TopTreeArity>,
         >(tree_c_size, &configs)?;
 
-        let (configs, replica_paths) = split_config_and_replica(
+        // tree_c_size stored in the config is the base tree size
```
dignifiedquire, line 506: why does it talk about tree_c_size here?

cryptonemo: Bad comment, will update

PR #1103, opened 2020-05-13
tooling: refactor create_replicas()

vmx:
In the fil-proofs-tooling it's a common operation to create replicas. This creation is now properly refactored so that it can be used by several binaries.

This commit doesn't change any functionality it's a pure refactoring.

The commits are self-contained so this PR might be easier to review on a commit per commit basis.


vmx:
@dignifiedquire Please merge whenever you like, you know better when. I don't want to cause more rebasing work.

PR #1104, opened 2020-05-13
style: add myself as codeowner

cryptonemo:


cryptonemo:
> This should be `@cryptonemo` and not `@nemo`

Just updated

PR #1116, opened 2020-05-19
unseal using Read and Write instead of Path or PathBuf

laser:
## Why does this PR exist?

@magik6k has indicated that he'd like to be able to unseal a sector to a file descriptor. While both file descriptors and regular files have paths, the more ergonomic way to do this is to use `Read` and `Write` instead of concrete types.

## What's in this PR?

This PR introduces a new function, `unseal_range`, which reads bytes from a `Read` and writes output to a `Write`. It then implements `get_unsealed_range` in terms of `unseal_range`.

### DIFF
--- a/filecoin-proofs/src/api/mod.rs
+++ b/filecoin-proofs/src/api/mod.rs
@@ -37,6 +37,58 @@ pub use self::seal::*;
 
 use storage_proofs::pieces::generate_piece_commitment_bytes_from_source;
 
+/// Unseals the sector at `sealed_path` and returns the bytes for a piece
+/// whose first (unpadded) byte begins at `offset` and ends at `offset` plus
+/// `num_bytes`, inclusive. Note that the entire sector is unsealed each time
+/// this function is called.
+///
+/// # Arguments
+///
+/// * `porep_config` - porep configuration containing the sector size.
+/// * `cache_path` - path to the directory in which the sector data's Merkle Tree is written.
+/// * `sealed_path` - path to the sealed sector file that we will unseal and read a byte range.
+/// * `output_path` - path to a file that we will write the requested byte range to.
+/// * `prover_id` - the prover-id that sealed the sector.
+/// * `sector_id` - the sector-id of the sealed sector.
+/// * `comm_d` - the commitment to the sector's data.
+/// * `ticket` - the ticket that was used to generate the sector's replica-id.
+/// * `offset` - the byte index in the unsealed sector of the first byte that we want to read.
+/// * `num_bytes` - the number of bytes that we want to read.
+#[allow(clippy::too_many_arguments)]
+pub fn get_unsealed_range<T: Into<PathBuf> + AsRef<Path>, Tree: 'static + MerkleTreeTrait>(
+    porep_config: PoRepConfig,
+    cache_path: T,
+    sealed_path: T,
+    output_path: T,
+    prover_id: ProverId,
+    sector_id: SectorId,
+    comm_d: Commitment,
+    ticket: Ticket,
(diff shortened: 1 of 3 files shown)


filecoin-proofs/src/api/mod.rs
```diff
@@ -111,12 +116,64 @@ pub fn get_unsealed_range<T: Into<PathBuf> + AsRef<Path>, Tree: 'static + Merkle
     // If the call to `extract_range` was successful, the `unsealed` vector must
     // have a length which equals `num_bytes_padded`. The byte at its 0-index
     // byte will be the the byte at index `offset_padded` in the sealed sector.
-    let written = write_unpadded(unsealed, &mut buf_writer, 0, num_bytes.into())
-        .with_context(|| format!("could not write to output_path={:?}", output_path.as_ref()))?;
+    let written = write_unpadded(unsealed, &mut unsealed_output, 0, num_bytes.into())
+        .with_context(|| "write_unpadded failed")?;
```
vmx, line 172: Just FYI: you if you don't use any variable, but just a string, you can use `context()` instead of `with_context()` =>

```rust
.context("write_unpadded failed")?;
```

storage-proofs/core/benches/pedersen.rs
```diff
@@ -157,6 +157,7 @@ fn pedersen_circuit_benchmark(c: &mut Criterion) {
     );
 }
 
+#[allow(dead_code)]
```
vmx, line 160: Rather than adding this, I'd put it into `criterion_group!()` at line 220.

filecoin-proofs/src/api/mod.rs
```diff
@@ -67,20 +67,77 @@ pub fn get_unsealed_range<T: Into<PathBuf> + AsRef<Path>, Tree: 'static + Merkle
...
+) -> Result<UnpaddedBytesAmount>
+where
+    P: Into<PathBuf> + AsRef<Path>,
+    R: Read,
+    W: Write,
+    Tree: 'static + MerkleTreeTrait,
+{
     ensure!(comm_d != [0; 32], "Invalid all zero commitment (comm_d)");
 
+    let mut sealed_sector = sealed_sector;
```
vmx, line 130: Is there a difference between this and making the function parameters `mut`?

storage-proofs/core/benches/pedersen.rs
laser: @vmx - what would that do? will that change the output of some benchmarking tool in some meaningful / surprising way?

filecoin-proofs/src/api/mod.rs
laser: done

storage-proofs/core/benches/pedersen.rs
vmx: Good point. It would run more benchmarks. Though I would then also not add the `allow(dead_code)` as I don't expect that it is intentional, so there should be a warning that the code isn't used until someone puts in time to properly fix it.

filecoin-proofs/src/api/mod.rs
laser: Negative. I'll remove the shadowing let-bindings and make the parameters mutable.

laser:
Sounds good!

storage-proofs/core/benches/pedersen.rs
laser: Sounds good. I've removed the `dead_code` as per your suggestion :)

laser:
@cryptonemo or @porcuquine - Would you mind reviewing this PR - or passing along to someone who can?

PR #1122, opened 2020-05-21
Remove and prevent usage of of unwrap()

tcharding:
Builds on top of https://github.com/filecoin-project/rust-fil-proofs/pull/1133 because we need version 1.45 of cargo in order to use the `unwrap_used` lint.

Using `expect()` with a useful argument string assists when debugging.

Remove all instances of `unwrap()`, typically this is done by using `expect()`. On occasion the pattern `unsure()` followed by `unwrap()` is replaced with an explicit match and the usage of anyhow's `bail` macro.
```
ensure!(foo.is_some(), "foo is empty");
foo = foo.unwrap();
```
Is replaced with
```
let foo = match foo {
    Some(foo) => foo,
    None => bail!("foo is empty"),
};
```

Close: #390


tcharding:
I'm going to go ahead and remove the `unwrap` calls in `storage-proofs/` and `filecoin-proofs/` this week. If I'm doing anything in this PR that you guys would like to see done differently please holla at me.

vmx:
@tcharding Please also check if there are cases, where an an unwrap should actually be a proper error instead on panicking. Those might still exist (I did run trough those a few months ago, but I might have missed some/there might be new ones). Though though there are also some hard/almost impossible ones when iterators are used.

tcharding:
> @tcharding Please also check if there are cases, where an an unwrap should actually be a proper error instead on panicking. Those might still exist (I did run trough those a few months ago, but I might have missed some/there might be new ones). Though though there are also some hard/almost impossible ones when iterators are used.

Thanks, I'll re-check over theses and keep my eyes open during the rest. Converting this PR to draft until I have the others done and the clippy rule added.

tcharding:
> Please also check if there are cases, where an an unwrap should actually be a proper error instead on panicking.

@vmx I'm seeing a bunch of `unwraps` where the function they are in could be changed to return a `Result`, electing _not_ to do this because it seems to be out of scope for this ticket. Also since I'm super new around here I don't want to go around changing all the callsites of a whole bunch of functions willy-nilly. Happy to do that after this if its deemed useful.

vmx:
> electing _not_ to do this because it seems to be out of scope for this ticket

Yes sounds good. I think for benchmarking, testing code it isn't important to have results (which is why there are so many `unwraps()`. Though if you find `unwrap()` in code which already returns a `Result`, it might make sense to have a closer look/taking notes somewhere to revisit those calls.

tcharding:
> > electing _not_ to do this because it seems to be out of scope for this ticket
> 
> Yes sounds good. I think for benchmarking, testing code it isn't important to have results (which is why there are so many `unwraps()`. Though if you find `unwrap()` in code which already returns a `Result`, it might make sense to have a closer look/taking notes somewhere to revisit those calls.

Got it, will do. Thanks.

tcharding:
This PR was seriously boring grunt work, bit sad to not see it get merged.  The issue it fixes is still open. I can't be bothered rebasing it any more.

vmx:
@cryptonemo I think we should rebase this PR and get it merged. It's indeed a lot of work and it certainly makes things better. If you agree and have the time to review/merge it, I'm happy to rebase it next week.

For the benchmarking code we might not enable the clippy flag it if turns out that it slows development down, but I think for the main part of the code it makes sense (during development you can easily comment the clippy flag out).

vmx:
@tcharding finally it got merged with https://github.com/filecoin-project/rust-fil-proofs/pull/1260. Thanks again for working on this!

porcuquine:
Me too — thanks for putting in the effort and sorry it took so long to land.

tcharding:
> Me too — thanks for putting in the effort and sorry it took so long to land.

No sweat, cheers.

PR #1125, opened 2020-05-27
optimize oct and quad insertion

porcuquine:
## WARNING: this PR reduces constraints and is therefore a breaking circuit change. Only merge once last v26 release has been cut. cc: @cryptonemo 

This PR updates `insertion.rs` to use hand-optimized versions of insertion when the size of the output array is either 4 or 8. These are the sizes used for our merkle trees and for which optimization will help. By taking advantage of redundancy in the output for each position, we can avoid the full generality otherwise required. This allows us to avoid the quadratic cost of `size * (size - 1)` constraints.

In order to aid in legibility and for consistency with internal spec audit (the motivation for this change), I have introduced a macro, `witness!` which allows declarative assignment of constraints to allocated numbers based on the value of an allocated boolean value.

Previous behavior is preserved and verified in tests — so bit input for insertion is still little endian. Bit and position indices are 0-indexed, and witness names rely on these conventions systematically.

For reference, here are the constraint definitions used to compute the oct insertion result:
```
    witness!(p1_x00 <== if b0 { a } else { b });
    witness!(p1_xx0 <== if b1 { c } else { &p1_x00 });
    witness!(p1 <== if b2 { c } else { &p1_xx0 });

    witness!(p2_x10 <== if b0 { d } else { a });
    witness!(p2_xx0 <== if b1 { &p2_x10 } else { c });
    witness!(p2 <== if b2 { d } else { &p2_xx0 });

    witness!(p3_xx0 <== if &b0_and_b1 { a } else { d });
    witness!(p3 <== if b2 { e } else { &p3_xx0 });

    witness!(p4_xx1 <== if &b0_nor_b1 { a } else { f });
    witness!(p4 <== if b2 { &p4_xx1 } else { e });

    witness!(p5_x01 <== if b0 { a } else { f });
    witness!(p5_xx1 <== if b1 { g } else { &p5_x01 });
    witness!(p5 <== if b2 { &p5_xx1 } else { f });

    witness!(p6_x11 <== if b0 { h } else { a });
    witness!(p6_xx1 <== if b1 { &p6_x11 } else { g });
    witness!(p6 <== if b2 { &p6_xx1 } else { g });

    witness!(p7_xx1 <== if &b0_and_b1 { a } else { h });
    witness!(p7 <== if b2 { &p7_xx1 } else { h });

    Ok(vec![p0, p1, p2, p3, p4, p5, p6, p7])
```

### DIFF
--- a/storage-proofs/core/src/gadgets/insertion.rs
+++ b/storage-proofs/core/src/gadgets/insertion.rs
@@ -3,7 +3,7 @@
 //! Insert an `AllocatedNum` into a sequence of `AllocatedNums` at an arbitrary position.
 //! This can be thought of as a generalization of `AllocatedNum::conditionally_reverse` and reduces to it in the binary case.
 
-use bellperson::gadgets::boolean::Boolean;
+use bellperson::gadgets::boolean::{AllocatedBit, Boolean};
 use bellperson::gadgets::num::AllocatedNum;
 use bellperson::{ConstraintSystem, SynthesisError};
 use ff::Field;
@@ -20,6 +20,22 @@ pub fn insert<E: Engine, CS: ConstraintSystem<E>>(
 ) -> Result<Vec<AllocatedNum<E>>, SynthesisError> {
     let size = elements.len() + 1;
     assert_eq!(1 << bits.len(), size);
+
+    // For the sizes we know we need, we can take advantage of redundancy in the candidate selection at each position.
+    // This allows us to accomplish insertion with fewer constraints, if we hand-optimize.
+    // We don't need a special case for size 2 because the general algorithm
+    // collapses to `conditionally_reverse` when size = 2.
+    //
+    // If no special cases have been hand-coded, use the general algorithm.
+    // This costs size * (size - 1) constraints.
+    //
+    // Future work: In theory, we could compile arbitrary lookup tables to minimize constraints and avoid
+    // the most general case except when actually required — which it never is for simple insertion.
+    if size == 4 {
+        return insert_4(cs, element, bits, elements);
+    } else if size == 8 {
+        return insert_8(cs, element, bits, elements);
+    };
 
     // Running example choices, represent inserting x into [1, 2, 3].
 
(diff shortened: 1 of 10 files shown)


storage-proofs/core/src/gadgets/insertion.rs
```diff
@@ -69,6 +86,172 @@ pub fn insert<E: Engine, CS: ConstraintSystem<E>>(
     Ok(result)
 }
 
+pub fn insert_4<E: Engine, CS: ConstraintSystem<E>>(
```
nicola, line 89: when do we use quads?

storage-proofs/core/src/gadgets/insertion.rs
```diff
@@ -21,6 +21,23 @@ pub fn insert<E: Engine, CS: ConstraintSystem<E>>(
     let size = elements.len() + 1;
     assert_eq!(1 << bits.len(), size);
 
+    match size {
```
DrPeterVanNostrand, line 24: Maybe just write this as
```rust
if size == 4 {
    return ...
} else if size == 8 {
    return ...
}
```
to avoid the empty match arm and the `if`s in pattern matching?

storage-proofs/core/src/gadgets/insertion.rs
porcuquine: Quads are needed any time we need to build a tree which cannot be constructed either with:
- only oct trees
- only oct and binary trees

Put differently, we need them whenever log N mod 3 = 2, where N is the number of 32-byte leaves. Although not currently supported, this would include either 16 or 128GiB sectors. Since we have only currently generated parameters for 32 and 64GiB sectors, quad hashing is dormant. But it's important that we preserve the freedom to support any power of two sector size — so the audit needs to cover quad hashes.

storage-proofs/core/src/gadgets/insertion.rs
```diff
@@ -69,6 +86,172 @@ pub fn insert<E: Engine, CS: ConstraintSystem<E>>(
...
+
+    A = element
+    b = elements[0]
+    c = elements[1]
+    d = elements[2]
+     */
+    let (b0, b1) = (&bits[0], &bits[1]);
+    let (a, b, c, d) = (&element, &elements[0], &elements[1], &elements[2]);
+
+    /// Define constrain macro to allow legible definition of positional constraints.
```
DrPeterVanNostrand, line 114: I think that this is a nice use of macros, however (imho) this adds code without expressing new ideas or clarifying preexisting ones. The "assign and constrain" operator looks (imo) out of place with respect to bellman's view of how gadgets should be defined and values allocated. Can the condition bit ever not be a reference?

storage-proofs/core/src/gadgets/insertion.rs
```diff
@@ -69,6 +86,172 @@ pub fn insert<E: Engine, CS: ConstraintSystem<E>>(
...
+        };
+
+        // Handle conditions which are references.
+        ( $var:ident <== if &$cond:ident { $a:expr } else { $b:expr }) => {
+            let $var = pick(cs.namespace(|| stringify!($var)), &$cond, $a, $b)?;
+        };
+    }
+
+    // Declaration:
+    witness!(p0_x0 <== if b0 { b } else { a });
```
DrPeterVanNostrand, line 128: Maybe change "p" to "pick" in the variable names and clarify what "x0" means?

storage-proofs/core/src/gadgets/insertion.rs
```diff
@@ -69,6 +86,172 @@ pub fn insert<E: Engine, CS: ConstraintSystem<E>>(
...
+    let b0_nor_b1 = match (b0, b1) {
+        (Boolean::Is(ref b0), Boolean::Is(ref b1)) => {
+            Boolean::Is(AllocatedBit::nor(cs.namespace(|| "b0 nor b1"), b0, b1)?)
+        }
+        _ => panic!("bits must be allocated and unnegated"),
+    };
+
+    // true if booleans b0 and b1 are both true: `b0 and b1`
+    // b0 * b1 = 1
+    let b0_and_b1 = match (&bits[0], &bits[1]) {
```
DrPeterVanNostrand, line 200: Maybe change this variable to "b0_nor_b1" or just "b1_and_b2_are_zero"?

storage-proofs/core/src/gadgets/insertion.rs
```diff
@@ -125,6 +308,11 @@ where
         }
     })?;
 
+    // Enforce (b - a) * cond = (b - c)
```
DrPeterVanNostrand, line 311: I think that this comment is unnecessary and confuses the point. Annotating a quadratic constraint using algebraic manipulations implies (imo) that r1cs performs that algebra and/or that the linear terms are able to commute.

storage-proofs/core/src/gadgets/insertion.rs
```diff
@@ -256,12 +445,15 @@ mod tests {
                 for i in 0..size - 1 {
                     let a = elements[i].get_value();
                     let b = inserted[i].get_value();
-
                     assert_eq!(a, b)
                 }
 
                 // One selection for each element of the result.
-                let expected_constraints = size * (size - 1);
+                let expected_constraints = match size {
```
DrPeterVanNostrand, line 452:
```rust
match size {
    4 => ... ,
    8 => ... ,
    _ => ... ,
}
```

storage-proofs/core/src/gadgets/insertion.rs
porcuquine: I added it to clarify for the internal auditors, and I think at least the first line should stay (and is consistent with Bellman internal documentation). I can remove the algebra and replace it with a prose summary n the Bellman style.

I'm not sure what you mean by 'linear terms commute', but we can discuss that elsewhere.

@nicola If you do want the original algebra to remain, please say so — otherwise, I'll remove per @DrPeterVanNostrand's suggestion.

storage-proofs/core/src/gadgets/insertion.rs
porcuquine: Done.

storage-proofs/core/src/gadgets/insertion.rs
porcuquine: Yeah… fixed.

storage-proofs/core/src/gadgets/insertion.rs
DrPeterVanNostrand: The first line in the comment is good with me. My comment was meant express that the algebra was misleading because the `=`'s sign could be interpreted as equivalency of circuit wiring, not just arithmetic. The current comment is good with me.

storage-proofs/core/src/gadgets/insertion.rs
porcuquine: 'p' is for position, actually. I added an example reading of the witness naming to set the reader on the right path.

PR #1126, opened 2020-05-27
feat: upgrade some dependencies

koushiro:
Signed-off-by: koushiro <koushiro.cqx@gmail.com>


koushiro:
Can someone review and merge this PR?

vmx:
Thanks for your PR. We are a bit hesitate to merge it right away as this repo is really a core piece of Filecoin. It will take us some time to look into all the implications those dependency updates have. But it will save us time as we can just take your changes, once we decide on an upgrade.

Do you know if the test failure is related to your change or whether it's unrelated?

koushiro:
> Do you know if the test failure is related to your change or whether it's unrelated?

The stable rust version is too old.

vmx:
I've assigned this one to me. I'll first tackle https://github.com/filecoin-project/rust-fil-proofs/issues/1142 and then look over the upgraded dependencies and look into the impact.

fil-proofs-tooling/Cargo.toml
```diff
@@ -31,14 +31,14 @@ storage-proofs = { path = "../storage-proofs"}
 filecoin-proofs = { path = "../filecoin-proofs"}
 tempfile = "3.0.8"
 cpu-time = "1.0.0"
-git2 = "0.10.1"
-heim = "0.0.9"
-futures-preview = "0.3.0-alpha.17"
+git2 = "0.13.6"
+heim = { version = "0.1.0-alpha.1", features = ["full", "runtime-tokio"] }
```
vmx, line 35: There seems to be a new version of `heim`. Also it looks like `heim` has support for async-std. In Rust Filecoin stuff we use async-std, rather than tokio. Do you think it would be hard to use the async-std instead?

PS: I'm not fully done with the review yet, but I thought I give early feedback on things I found.

vmx:
@koushiro As we now use a newer version of stable Rust, the tests should pass. Could you please rebase your changes. Ideal would be if you put all your changes into a single commit and rebase that on master.

In case you don't have the time, please let me know and I'll make those changes to this PR.

koushiro:
@vmx PTAL

vmx:
In my previous comment I forgot to mention that changing from a fixed version of `regex` to a version range is fine. I clarified that with @porcuquine, it was only fixed due to some upstream issue that made the tests fail, which seems to be fixed now.

This PR needs another rebase. But I'd wait for that until we are right before the merge, else we keep rebasing all the time.

@cryptonemo This PR looks good to me. Please let @koushiro know when merging it comes close, so that he can rebase it one last time (as before I'm happy to do the rebase as well).

PR #1131, opened 2020-05-29
Pretty print in benchy

nicola:


cryptonemo:
@nicola Can you fix the clippy complaints?

nicola:
Hey people just for reference, this was done for an internal audit and for future reference.
I am not planning to do further work on this, feel free to change it as you like or to drop it and close it.

vmx:
I just made that one requested change, I leave it to the maintainers to merge is or not. If it would be me, I'd take it, it seems useful.

porcuquine:
It's worth having. I'll make the change I want *at some point* when I have time to run it and see how output looks.

vmx:
The main pretty print code now lives in [Bellman](https://github.com/filecoin-project/bellman/blob/1c96dce05886f26f1c6ddb6f1227fc64ac30cd7d/src/util_cs/metric_cs.rs).

It wasn't really intended to be merged (https://github.com/filecoin-project/rust-fil-proofs/pull/1131#issuecomment-643192050), hence I propose that we just close the issue and not spending time on wiring it up properly.

cryptonemo:
> The main pretty print code now lives in [Bellman](https://github.com/filecoin-project/bellman/blob/1c96dce05886f26f1c6ddb6f1227fc64ac30cd7d/src/util_cs/metric_cs.rs).
> 
> It wasn't really intended to be merged ([#1131 (comment)](https://github.com/filecoin-project/rust-fil-proofs/pull/1131#issuecomment-643192050)), hence I propose that we just close the issue and not spending time on wiring it up properly.

I'm good with that.  If @porcuquine agrees, please close or signal to close.

porcuquine:
Right… I forgot we moved those. The best answer would be to port these changes over to Bellman then. It has non-zero value, and otherwise people will keep hacking it up in an ad-hoc way as needed. I'm fine with closing this one but would rather see it integrated there.

porcuquine:
> It wasn't really intended to be merged (#1131 (comment)), 

That was @nicola's tactful way of saying he wouldn't respond to review comments — not that he didn't want the code saved.

vmx:
@porcuquine Should then all this also get wired up to benchy, or just pretty printing added to bellman?

This was parted of the stacked benchmarks, which don't even exist anymore. So we'd need to find a new place for it in benchy (if we want that).

porcuquine:
I think supporting the underlying pretty-printing through the appropriate constraint system (in Bellman) sounds like the easier step worth doing.

If we can also add it as a general option for benchy, that would be great. (I'm showing my ignorance of benchy's structure here, though.)

The middle ground might be a simple sample usage *somewhere* in benchy (if the general solution is too much work for now).

vmx:
Benchy these days (at least from looking at the code) is just for printing out timings and nothing else. Hence I suggest that we merge https://github.com/filecoin-project/bellman/pull/106 into bellman in case someone needs the pretty printing, but I don't integrate it into benchy.

If we ever need pretty printing of circuits I'd create a separate tool for that (though there are currently no plans to do so).

Given all that I'm closing this issue, if you disagree please re-open.

porcuquine:
That's fine. For clarity, benchy still allows measuring the size (in constraints) of circuits, right?

vmx:
> For clarity, benchy still allows measuring the size (in constraints) of circuits, right?

I don't think so. From looking at the code it seems that it's really only about timing (which I think makes sense):

 - https://github.com/filecoin-project/rust-fil-proofs/blob/feb0862a12d936e8a7f7b7b21776236d69baa931/fil-proofs-tooling/src/bin/benchy/window_post.rs#L49-L66
 - https://github.com/filecoin-project/rust-fil-proofs/blob/feb0862a12d936e8a7f7b7b21776236d69baa931/fil-proofs-tooling/src/bin/benchy/winning_post.rs#L24-L31

porcuquine:
I think prodbench is part of benchy, and seems to be the answer to my next question: then how are we supposed to count the constraints (which really needs to remain possible): https://github.com/filecoin-project/rust-fil-proofs/blob/6383bb44e9ea577073ffae7abdcb2dda2ef95d85/fil-proofs-tooling/src/bin/benchy/prodbench.rs#L268-L272

So maybe that is also the answer to 'where would we hook up circuit inspection', when that is needed again.

PR #1133, opened 2020-05-31
Cargo bump version

tcharding:
While working on https://github.com/filecoin-project/rust-fil-proofs/issues/390 it was discovered that we need to use a lint that is not supported by the current toolchain.

Bump cargo version to the newest available version (1.44 -> 1.45) and clear the new clippy warnings.


vmx:
Given https://github.com/filecoin-project/rust-fil-proofs/issues/1142, I think we should close that one and address it when we are going for 1.45. @tcharding What do you think?

tcharding:
Sure, no worries - someone should probably close https://github.com/filecoin-project/rust-fil-proofs/issues/390 that the lint required needs cargo v1.45

PR #1135, opened 2020-05-31
use partial parent cache and other methods to reduce the mem usage for sdr

dtynn:
tested on a machine
- AMD Ryzen 9 3900X 12-Core Processor
- 62.9GiB physical mem
- 10GiB swap on ssd
- 8T hdd (7200rpm) for staged file, sealed file & cache dir

with
- 13:32:51 ~ 20:11:44 for seal_pre_commit_phase1
- 20:11:44 ~ 23:04:04 for seal_pre_commit_phase2

TODOs:
- [x] patch bellman to serialize circuit synthesizing & calculating: [bellman pr #81](https://github.com/filecoin-project/bellman/pull/81)
- [ ] use mixed-base-layer to see if we can reduce some more mem for seal_pre_commit_phase1
- [ ] copying base layer to exp cost ~9min in the last 4 layers, with a high ratio of SWAPIN, which I think, can be optimized

[see log here](https://gist.github.com/dtynn/80cbdaaf893644a7830d93722a86fb0d)

### DIFF
--- a/storage-proofs/porep/src/stacked/vanilla/cache.rs
+++ b/storage-proofs/porep/src/stacked/vanilla/cache.rs
@@ -1,0 +1,15 @@
+use std::env;
+use std::path::{Path, PathBuf};
+
+pub const SDR_CACHE_ENV_VAR: &str = "FIL_PROOFS_SDR_CACHE";
+pub const PARAMETER_CACHE_DIR: &str = "/var/tmp/filecoin-proof-sdr-cache/";
+
+pub fn dir_name() -> PathBuf {
+    env::var(SDR_CACHE_ENV_VAR)
+        .map(PathBuf::from)
+        .unwrap_or(PathBuf::from(PARAMETER_CACHE_DIR))
+}
+
+pub fn file_path<P: AsRef<Path>>(name: P) -> PathBuf {
+    dir_name().join(name)
+}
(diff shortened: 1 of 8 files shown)


storage-proofs/porep/src/stacked/vanilla/cache.rs
```diff
@@ -0,0 +1,15 @@
...
+use std::path::{Path, PathBuf};
+
+pub const SDR_CACHE_ENV_VAR: &str = "FIL_PROOFS_SDR_CACHE";
+pub const PARAMETER_CACHE_DIR: &str = "/var/tmp/filecoin-proof-sdr-cache/";
+
+pub fn dir_name() -> PathBuf {
+    env::var(SDR_CACHE_ENV_VAR)
+        .map(PathBuf::from)
+        .unwrap_or(PathBuf::from(PARAMETER_CACHE_DIR))
+}
```
porcuquine, line 11: For env var configuration, please use the mechanism in `settinsgs.rs` for consistency and to consolidate in one place developers and users can keep track of.

dignifiedquire:
@dtynn what are the numbers you are seeing on that setup with the current implementation, so we have a comparison.

dtynn:
> @dtynn what are the numbers you are seeing on that setup with the current implementation, so we have a comparison.

considering for seal_pre_commit_phase1 only:

62.9GiB physical + 10GiB swap, both full used  VS  56GiB (parent cache) + 64GiB, with 50% ~ 80% more time cost for each layer

if we use the official implementation without full cached parent indexes(FIL_PROOFS_MAXIMIZE_CACHING=0), mem usage reduces, but the time would be much longer (don't remember the exact numbers)

dignifiedquire:
@dtynn for the numbers you reported, what value for `partial_num` did you use?

dtynn:
> @dtynn for the numbers you reported, what value for `partial_num` did you use?

128

dignifiedquire:
Thanks a lot for this @dtynn! I have used this code to build https://github.com/filecoin-project/rust-fil-proofs/pull/1163 which has similar properties, but better integrated into the machinery.

PR #1138, opened 2020-06-01
feat: allow tree_r_last to be built on gpu

cryptonemo:
feat: attempt to improve gpu tree_c layer retrieval


cryptonemo:
Work in progress.  Still uses CPU compat mode for tree builders.  Todo: verify poseidon standard/strengthened on tests.  Param generation bump required?

porcuquine:
Yes, this will require bumping the parameter version.

cryptonemo:
@porcuquine Can you confirm if poseidon constraints (tests) should have to be updated with this change (updated neptune, etc)?

porcuquine:
Yes, and they should go down.

cryptonemo:
Note: This comes out of draft status when neptune v1 is released.

storage-proofs/porep/src/stacked/vanilla/proof.rs
```diff
@@ -451,39 +462,41 @@ impl<'a, Tree: 'static + MerkleTreeTrait, G: 'static + Hasher> StackedDrg<'a, Tr
...
-                    rayon::scope(|s| {
-                        let n = num_cpus::get();
-
-                        // only split if we have at least two elements per thread
-                        let num_chunks = if n > chunked_nodes_count * 2 { 1 } else { n };
+                    // Allocate layer data array and insert a placeholder for each layer.
+                    let mut layer_data = Vec::with_capacity(layers);
+                    for _ in 0..layers {
+                        layer_data.push(Vec::new());
+                    }
```
dignifiedquire, line 469: could be written as 

```rust
let mut layer_data = vec![vec![]; layers];
```

storage-proofs/porep/src/stacked/vanilla/proof.rs
```diff
@@ -451,39 +462,41 @@ impl<'a, Tree: 'static + MerkleTreeTrait, G: 'static + Hasher> StackedDrg<'a, Tr
...
+                                let end = start + chunked_nodes_count;
+                                let elements: Vec<<Tree::Hasher as Hasher>::Domain> =
+                                    store.read_range(std::ops::Range { start, end }).unwrap();
+
+                                let fr_elements: Vec<_> =
+                                    elements.into_iter().map(|el| el.into()).collect();
+                                layer_data[k - 1] = fr_elements;
+                            }
+                        });
+                    });
```
dignifiedquire, line 490: could be 

```rust
for layer_elements in layer_data.iter_mut() {
    // ... 
    layer_elements.extend(elements.into_iter().map(Into::into));
}
```

which should avoid some allocations

storage-proofs/porep/src/stacked/vanilla/proof.rs
```diff
@@ -497,28 +510,78 @@ impl<'a, Tree: 'static + MerkleTreeTrait, G: 'static + Hasher> StackedDrg<'a, Tr
...
+                        base_data
+                            .into_par_iter()
+                            .chunks(column_write_batch_size)
+                            .enumerate()
+                            .try_for_each(|(index, fr_elements)| {
+                                let mut buf = Vec::with_capacity(batch_size * NODE_SIZE);
 
-                        let flat_tree_data: Vec<_> = tree_data
+                                for fr in fr_elements {
+                                    buf.extend(fr_into_bytes(&fr));
```
porcuquine, line 555: This will still allocate a buffer into which to write the conversion. I think you should add a new version of this function which writes directly into `buf`.

cryptonemo: For this instance (of tree data), it's taking about a half second on a 32GiB run.  The base data flattening is now taking about 2-3 seconds.

cryptonemo: The buf being extended has been pre-allocated above that.  I'm not sure if you're saying to avoid that allocation, or the one that extend would do had it not been pre-allocated.

storage-proofs/porep/src/stacked/vanilla/proof.rs
cryptonemo: The inner vec is entirely replaced

storage-proofs/porep/src/stacked/vanilla/proof.rs
cryptonemo: I'll take a look at this, thanks

storage-proofs/porep/src/stacked/vanilla/proof.rs
```diff
@@ -497,28 +510,78 @@ impl<'a, Tree: 'static + MerkleTreeTrait, G: 'static + Hasher> StackedDrg<'a, Tr
...
-                        // Persist the data to the store based on the current config.
-                        DiskStore::<<Tree::Hasher as Hasher>::Domain>::new_from_slice_with_config(
-                            tree_data_len,
-                            Tree::Arity::to_usize(),
-                            &flat_tree_data[..],
-                            config.clone(),
-                        )?;
+                                for fr in fr_elements {
+                                    buf.extend(fr_into_bytes(&fr));
+                                }
```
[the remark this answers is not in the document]
cryptonemo: I think I want to avoid this.  This looks eerily like the construction that was originally in place that bottlenecked hard.

storage-proofs/porep/src/stacked/vanilla/proof.rs
porcuquine: I was talking about the allocation inside `fr_into_bytes`.

cryptonemo: :+1: Got it, thanks.  Given the overall speed improvements, we can wait on that a bit since this is no longer bottlenecking.

storage-proofs/porep/src/stacked/vanilla/proof.rs
cryptonemo: I see what you're saying now (along with the above comment).  Thanks!

PR #1132, opened 2020-05-29
feat: update tests to accomodate strengthened poseidon

cryptonemo:

PR #1139, opened 2020-06-02
Include layer index before node when creating label preimage.

porcuquine:
This adds constraints, which slightly surprised me — since it does not increase the length of the preimage being hashed. ~I believe the reason for this is that the padding bits being replaced were in fact boolean constants. Some initial operations in the sha256 computation involving them may therefore have been optimized away at synthesis time.~

*UPDATE:* I don't think my explanation above makes sense. I think the actual source of the increase is the booleanity constraints introduced when allocating the bits.

*UPDATE 2*: Although my update is accurate, there should not have been an allocation. The latest should now correctly use constants everywhere.

### DIFF
--- a/storage-proofs/core/src/gadgets/por.rs
+++ b/storage-proofs/core/src/gadgets/por.rs
@@ -901,6 +901,11 @@ mod tests {
         test_private_por_input_circuit::<TestTree<PoseidonHasher, typenum::U4>>(1_172);
     }
 
+    #[test]
+    fn test_private_por_input_circuit_poseidon_oct() {
+        test_private_por_input_circuit::<TestTree<PoseidonHasher, typenum::U8>>(1_068);
+    }
+
     fn test_private_por_input_circuit<Tree: MerkleTreeTrait>(num_constraints: usize) {
         let rng = &mut XorShiftRng::from_seed(crate::TEST_SEED);
 
--- a/storage-proofs/porep/src/stacked/circuit/create_label.rs
+++ b/storage-proofs/porep/src/stacked/circuit/create_label.rs
@@ -1,20 +1,21 @@
 use bellperson::gadgets::{
     boolean::Boolean,
     sha256::sha256 as sha256_circuit,
-    {multipack, num},
+    uint32, {multipack, num},
 };
 use bellperson::{ConstraintSystem, SynthesisError};
 use ff::PrimeField;
 use fil_sapling_crypto::jubjub::JubjubEngine;
 use storage_proofs_core::gadgets::uint64;
 
-use super::super::vanilla::TOTAL_PARENTS;
+use crate::stacked::vanilla::TOTAL_PARENTS;
 
 /// Compute a single label.
 pub fn create_label_circuit<E, CS>(
     mut cs: CS,
     replica_id: &[Boolean],
     parents: Vec<Vec<Boolean>>,
+    layer_index: uint32::UInt32,
     node: uint64::UInt64,
 ) -> Result<num::AllocatedNum<E>, SynthesisError>
 where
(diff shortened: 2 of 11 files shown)


storage-proofs/core/src/gadgets/por.rs
```diff
@@ -901,6 +901,11 @@ mod tests {
         test_private_por_input_circuit::<TestTree<PoseidonHasher, typenum::U4>>(1_172);
     }
 
+    #[test]
+    fn test_private_por_input_circuit_poseidon_oct() {
+        test_private_por_input_circuit::<TestTree<PoseidonHasher, typenum::U8>>(1_068);
+    }
+
```
porcuquine, line 908: This was a missing test I added while I was editing circuit tests. It is not otherwise related to this PR.

porcuquine:
I'm not sure why the seal lifecycle tests are failing on CI. With fresh groth parameters, they (at least 2Kib) pass locally for me. Maybe I accidentally reused a CI parameter cache key. Will poke at it more tomorrow.

storage-proofs/core/src/gadgets/por.rs
cryptonemo: Yes, I also made that mistake, so v27c as a key has been burned first here: https://github.com/filecoin-project/rust-fil-proofs/pull/1132  Then here: https://github.com/filecoin-project/rust-fil-proofs/pull/1138

Go ahead and claim v27d and I'll update mine for things other than that.

cryptonemo:
> I'm not sure why the seal lifecycle tests are failing on CI. With fresh groth parameters, they (at least 2Kib) pass locally for me. Maybe I accidentally reused a CI parameter cache key. Will poke at it more tomorrow.

Yes, I also made that mistake, so v27c as a key has been burned first here: https://github.com/filecoin-project/rust-fil-proofs/pull/1132  Then here: https://github.com/filecoin-project/rust-fil-proofs/pull/1138

Go ahead and claim v27d and I'll update mine for things other than that.

storage-proofs/core/src/parameter_cache.rs
```diff
@@ -16,7 +16,7 @@ use std::io::{self, SeekFrom};
 use std::path::{Path, PathBuf};
 
 /// Bump this when circuits change to invalidate the cache.
-pub const VERSION: usize = 27;
+pub const VERSION: usize = 28;
```
cryptonemo, line 19: Oh, this requires a version bump here to 28?

dignifiedquire: this version shouldn’t change, only when we have published 27 and need to change sth

storage-proofs/porep/src/stacked/circuit/create_label.rs
```diff
@@ -177,8 +182,15 @@ mod tests {
         assert_eq!(cs.num_constraints(), 532_024);
 
         let (l1, l2) = data.split_at_mut(size * NODE_SIZE);
-        super::super::super::vanilla::create_label_exp(&graph, &id_fr.into(), &*l2, l1, node)
-            .unwrap();
+        super::super::super::vanilla::create_label_exp(
```
DrPeterVanNostrand, line 185: Super not cool

cryptonemo: I believe you mean Super not chocolate ...

storage-proofs/core/src/parameter_cache.rs
porcuquine: Hmmmm. I guess we have concretized a more recent policy. I'll go with that but note it means developers will more often get caught with stale parameters.

storage-proofs/porep/src/stacked/circuit/create_label.rs
porcuquine: Dammit, why did this change have to touch that line? @DrPeterVanNostrand I'll look into it.

Issue #869, opened 2019-09-10
Blake2 digest shouldn't be truncated


Kubuxu:
Blake2 digest length is a parameter of the hash function (which changes the hash output).
The `blake2b_simd` supports it: https://docs.rs/blake2b_simd/0.5.8/blake2b_simd/struct.Params.html#method.hash_length

Explicit digest length should be used instead of truncating hex output to 32 bytes.

https://github.com/filecoin-project/rust-fil-proofs/blob/e6fa4232404641bb4be1ffc42944d2e734ab9748/filecoin-proofs/src/param.rs#L36-L44

tcharding:
After investigating this I've found that the issue description may be, if I am correct, slightly misleading. The current code takes a slice of 32 ascii characters representing _hex digits_ not 32 bytes, 32 hex digits is 16 bytes (4 bits per digit). The issue remains however that we are truncating the hash since we are using default params for the hasher which uses a `hash_length` of 8 * wordsize bytes (so 8 * `size_of<u64>` == 64 , on a 64 bit architecture). Of the 64 available bytes we are using just 16.

The question is, do we want to create a hash that is just 16 bytes long and stick with the 32 character string or use a string that is 128 characters long and use the default 64 byte hash length?

In order to close this issue https://github.com/filecoin-project/rust-fil-proofs/compare/master...tcharding:869-hash-length?expand=1 configures blake2 to use a hash length of 16 bytes, maintaining the current hex string length.

dignifiedquire:
If we do this, we will need to have a migration routine which renames existing parameters and update a lot of tooling in lotus. While it would be desirable, it does not actually impact anything at the moment, so I will close for now.

PR #1140, opened 2020-06-02
fix(filecoin-proofs): explicitly use 16 byte hash length

tcharding:
Currently we are hashing filenames using blake2 configured with a hash
length of 64 bytes. We are the converting this to hex digits and
truncating to take the first 32 digits - this gives us a string that
represents 16 bytes from the hash.

Blake2 can be configured with a hash length, if we want 16
bytes (in order to maintain the current string length) we should
explicitly configure blake2 to use a 16 byte hash length.

Closes: #869


tcharding:
> if we change this we should use blake3, given the size of params.
> this is a breaking change that needs to update more places in other repos like go-paramfetch and possibly others, so this needs to be well coordinated with the node implementations

Thanks for the review. Is this something we want done? Happy to do the work, conscious of not causing too much churn while I'm still new around the code base. A less intrusive fix could be just to move the `take 32 characters from the hex string' to a public scope function, use it in the tests also, and document whats going on fully and why (including potential breakage sites if it gets changed)? Would that be preferable?

tcharding:
Closing, no forward progres.

PR #1143, opened 2020-06-02
Constrain bits to result when creating label.

porcuquine:
When creating a label, the bits resulting from the SHA256 circuit were not being constrained to the returned result — although that result was correctly computed. This PR corrects that.

- Use `pack_bits` instead of `compute_multipacking`. This adds one constraint per call to `create_label`.
- Use a general utility function (`reverse_bit_numbering` was `fixup_bits`) rather than ad-hoc transformation of the sha256-result bit array.
- Rename `fixup_bits`, an inscrutable and domain-specific name, to the transparent and general-purpose `reverse_bit_numbering`. NOTE: I chose 'bit numbering' (https://en.wikipedia.org/wiki/Bit_numbering) rather than 'bit endianness' to avoid potential ambiguity when mentioning 'endianness', since this function preserves byte order (when the bit array is viewed as a flattened sequence of octets).
- Generalize the renamed function to not require 255-bit input but instead zero-pad to fulfill the functional requirement that octets should be full before reversal.

### DIFF
--- a/storage-proofs/porep/src/stacked/circuit/create_label.rs
+++ b/storage-proofs/porep/src/stacked/circuit/create_label.rs
@@ -1,12 +1,8 @@
-use bellperson::gadgets::{
-    boolean::Boolean,
-    sha256::sha256 as sha256_circuit,
-    uint32, {multipack, num},
-};
+use bellperson::gadgets::{boolean::Boolean, num, sha256::sha256 as sha256_circuit, uint32};
 use bellperson::{ConstraintSystem, SynthesisError};
 use ff::PrimeField;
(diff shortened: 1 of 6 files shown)


storage-proofs/porep/src/stacked/circuit/create_label.rs
```diff
@@ -181,7 +169,7 @@ mod tests {
         .expect("key derivation function failed");
 
         assert!(cs.is_satisfied(), "constraints not satisfied");
-        assert_eq!(cs.num_constraints(), 532_024);
+        assert_eq!(cs.num_constraints(), 532_025);
```
porcuquine, line 173: NOTE: constraint count increases by 1.

PR #1144, opened 2020-06-03
Add opaque porep_id as public input and use when constructing replica_id

porcuquine:
This PR adds a new 32-byte public input, `porep_id`, to SDR. This value is used to construct `replica_id`, the drg seed, and the Feistel keys. As a public input, it must be supplied (with identical value) by both prover and verifier. The intention at the protocol level is that proofs consumers should pass a value derived from the `RegisteredProof` — since it is guaranteed to distinguish all extant PoReps. However, from the perspective of the `rust-fil-proofs` the `porep_id` is treated as opaque and is required only to be used consistently. If consumers choose to duplicate `porep_id`s, they take responsibility for the possibility that `replica_id` or DRG graph might be reused between the set of concrete proofs the duplicated id is used to create. If the Filecoin spec is modified to require a distinct `porep_id` for every distinct concrete type, such usage will violate the protocol.

*UPDATE: now supports Feistel keys*
- `porep_id` is a PublicParam
- It is passed to `derive_porep_domain_seed`, which also takes domain separation tag.
- In this way, the DRG seed and Feistel keys are created when the graphs are instantiated.

### DIFF
--- a/storage-proofs/core/src/drgraph.rs
+++ b/storage-proofs/core/src/drgraph.rs
@@ -7,6 +7,7 @@ use rand::{rngs::OsRng, Rng, SeedableRng};
 use rand_chacha::ChaCha8Rng;
 use sha2::{Digest, Sha256};
 
+use crate::crypto::{derive_porep_domain_seed, DRSAMPLE_DST};
 use crate::error::*;
 use crate::fr32::bytes_into_fr_repr_safe;
 use crate::hasher::{Hasher, PoseidonArity};
@@ -54,7 +55,7 @@ pub trait Graph<H: Hasher>: ::std::fmt::Debug + Clone + PartialEq + Eq {
         nodes: usize,
         base_degree: usize,
         expansion_degree: usize,
-        seed: [u8; 28],
+        porep_id: [u8; 32],
     ) -> Result<Self>;
     fn seed(&self) -> [u8; 28];
 
@@ -203,7 +204,7 @@ impl<H: Hasher> Graph<H> for BucketGraph<H> {
         nodes: usize,
         base_degree: usize,
         expansion_degree: usize,
-        seed: [u8; 28],
+        porep_id: [u8; 32],
     ) -> Result<Self> {
         ensure!(expansion_degree == 0, "Expension degree must be zero.");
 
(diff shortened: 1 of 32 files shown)


nicola:
It's ok to be opaque, but it's important the at verification time, the chain correctly verifies the correct porep-id. Does this live in CompoundProof?

porcuquine:
It is a public input. It can be derived from the `RegisteredProof` at the API layer, which handles routing and accounts for versioning. As long as proofs consumers provide the correct `RegisteredProof` at proof and verification time (without which verification could not be expected to work in any case), the correct `porep_id` can be attached and forwarded to the inner API.

nicola:
Perfect. Future review of integration of proofs into Filecoin should catch this. I will make a note.

nicola:
Just a quick note: this PR does not directly address issue "20. FeistelKeys, Seed_D should be specific to a particular Proof of Replication e.g. H(RegisteredProofID || random nonce)", but only "21. ReplicaID should contain the RegisteredProofID". Correct?

porcuquine:
No, it address both. See the second commit: https://github.com/filecoin-project/rust-fil-proofs/pull/1144/commits/5cb433151d5ff52afe3cbfa95dfb73f24b02cac9.

porcuquine:
Reading more closely, it does handle the DRG part, but not the Feistel keys. I will add.

porcuquine:
@nicola Updated with Feistel keys and DSTs.

porcuquine:
FYI, @cryptonemo: this change is going to require work in `rust-filecoin-api` — since `porep_id` will need to be set in the `PorepConfig`. This may also require some work to ensure the concrete enum values of `RegisteredProof`s agree between the spec (actors) and the source. I will skip discussion here of further, related changes which may be required but wanted to at least let you know there will be more than the minimum to change in `rust-filecoin-api`, and that this may end up affecting `filecoin-ffi` and spec actors as well. (cc: @laser )

storage-proofs/core/src/drgraph.rs
```diff
@@ -216,10 +217,14 @@ impl<H: Hasher> Graph<H> for BucketGraph<H> {
             "The number of metagraph nodes must be precisely castable to `f64`"
         );
 
+        let mut drg_seed = [0; 28];
+        let raw_seed = derive_porep_domain_seed("Filecoin_DRSample", porep_id);
```
dignifiedquire, line 221: this string should be a constant, not inline

storage-proofs/core/src/drgraph.rs
```diff
@@ -216,10 +217,14 @@ impl<H: Hasher> Graph<H> for BucketGraph<H> {
...
+        drg_seed.copy_from_slice(&raw_seed[..28]);
```
dignifiedquire, line 222: hmmmmm this seems very arbitrary, that we use the first 28 bytes

porcuquine: I think you wrote the code using 28 bytes to seed `ChaChaRng`. If you think it should be changed, let me know how, please.

PR #1145, opened 2020-06-03
Help auditors by special-casing binary insert.

porcuquine:
This PR special-cases insertion when the array size is 2. It is structurally equivalent to the hand-optimized cases already provided for sizes 4 and 8. However, the resulting circuit synthesized is identical to what it replaces, since binary insertion is already optimal and reduces to the case made explicit here.

Note that neither tests counting constraints (throughout the repo) or the CI parameter cache key are modified by this PR. That is because it does not affect circuits.


storage-proofs/core/src/gadgets/insertion.rs
```diff
@@ -381,7 +408,7 @@ mod tests {
 
     #[test]
     fn test_insert() {
-        for log_size in 1..=3 {
+        for log_size in 1..=4 {
```
porcuquine, line 411: Test one more size, 16, so the general form is still exercised.

PR #1146, opened 2020-06-03
Improve names in PoR circuit.

porcuquine:


storage-proofs/core/src/gadgets/por.rs
```diff
@@ -163,7 +163,7 @@ impl<H: Hasher, Arity: 'static + PoseidonArity> SubPath<H, Arity> {
             auth_path_bits.extend_from_slice(&index_bits);
 
             // Witness the authentication path elements adjacent at this depth.
-            let path_elements = elements
+            let hash_nums = hashes
                 .iter()
```
dignifiedquire, line 167: hashes is very generic, should keep the path reference in the name

porcuquine: Done.

Issue #1123, opened 2020-05-25
Fix all clippy warning


vmx:
### Description

Currently CI isn't running the same clippy command as other repos (e.g. https://github.com/filecoin-project/filecoin-ffi/blob/6109b6ad2fa9968941c206161dd01ac059011d4e/.circleci/config.yml#L138). To catch all warnning this should be run:

    cargo clippy --all-targets --all-features -- -D warnings

In order to enable that on CI, all issues need to be fixed first

### Acceptance criteria

    cargo clippy --all-targets --all-features -- -D warnings

passes without any warning or errors

### Risks + pitfalls

This changes code everywhere, there might be merge conflicts with other open PRs.

### Where to begin

Where the first warning/error is.

cryptonemo:
Thanks!

PR #1147, opened 2020-06-03
refactor(storage-proofs): fix clippy warnings

tcharding:
Fix all clippy warnings produced with the following invocation:
```
cargo clippy --all-targets --all-features -- -D warnings
```

None of the fixes are particularly noteworthy except for:

- Usage of combinators instead of array index, arguably makes the test code harder to read: https://github.com/filecoin-project/rust-fil-proofs/pull/1147/commits/3592d8c78f70040c765ddaeb923e811dbe2c24c8
- Add test to criterion_group (see discussion below): https://github.com/filecoin-project/rust-fil-proofs/pull/1147/commits/889f4ad72e9d69f313ccf04b8fb2635e85a05fcf
- Allow `unit_arg` lint for usage of `black_box`: https://github.com/filecoin-project/rust-fil-proofs/pull/1147/commits/5d67be339a34ae21eb9a4d4cc5b2d475762d8465

Closes: #1123

### DIFF
--- a/storage-proofs/core/benches/drgraph.rs
+++ b/storage-proofs/core/benches/drgraph.rs
@@ -2,6 +2,7 @@ use criterion::{black_box, criterion_group, criterion_main, Criterion, Parameter
 use storage_proofs_core::drgraph::*;
 use storage_proofs_core::hasher::pedersen::*;
 
+#[allow(clippy::unit_arg)]
 fn drgraph(c: &mut Criterion) {
     let params = vec![12, 24, 128, 1024];
 
--- a/storage-proofs/core/benches/pedersen.rs
+++ b/storage-proofs/core/benches/pedersen.rs
@@ -15,7 +15,7 @@ impl<'a> Circuit<Bls12> for PedersenExample<'a> {
     fn synthesize<CS: ConstraintSystem<Bls12>>(self, cs: &mut CS) -> Result<(), SynthesisError> {
         let data: Vec<Boolean> = self
             .data
-            .into_iter()
+            .iter()
             .enumerate()
             .map(|(i, b)| {
                 Ok(Boolean::from(boolean::AllocatedBit::alloc(
@@ -45,7 +45,7 @@ impl<'a> Circuit<Bls12> for PedersenMdExample<'a> {
     fn synthesize<CS: ConstraintSystem<Bls12>>(self, cs: &mut CS) -> Result<(), SynthesisError> {
         let data: Vec<Boolean> = self
             .data
-            .into_iter()
+            .iter()
             .enumerate()
             .map(|(i, b)| {
                 Ok(Boolean::from(boolean::AllocatedBit::alloc(
--- a/storage-proofs/core/benches/pedersen.rs
+++ b/storage-proofs/core/benches/pedersen.rs
@@ -212,5 +212,6 @@ criterion_group!(
     pedersen_benchmark,
     pedersen_md_benchmark,
     pedersen_circuit_benchmark,
+    pedersen_md_circuit_benchmark,
 );
 criterion_main!(benches);
--- a/storage-proofs/core/benches/pedersen.rs
+++ b/storage-proofs/core/benches/pedersen.rs
@@ -106,9 +106,7 @@ fn pedersen_md_benchmark(c: &mut Criterion) {
 fn pedersen_circuit_benchmark(c: &mut Criterion) {
     let mut rng1 = thread_rng();
     let groth_params = generate_random_parameters::<Bls12, _, _>(
-        PedersenExample {
-            data: &vec![None; 256],
-        },
+        PedersenExample { data: &[None; 256] },
         &mut rng1,
     )
     .unwrap();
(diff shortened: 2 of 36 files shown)


storage-proofs/core/benches/drgraph.rs
```diff
@@ -15,7 +15,8 @@ fn drgraph(c: &mut Criterion) {
 
                 b.iter(|| {
                     let mut parents = vec![0; 6];
-                    black_box(graph.parents(2, &mut parents).unwrap());
+                    // parents() never fails, call is_ok() to quieten clippy.
```
vmx, line 18: It's out of scope for this PR, but wouldn't it make sense create a new PR which changes `parents()` to not return a `Result`, but just nothing?

storage-proofs/core/benches/pedersen.rs
```diff
@@ -157,60 +155,6 @@ fn pedersen_circuit_benchmark(c: &mut Criterion) {
     );
 }
 
-fn pedersen_md_circuit_benchmark(c: &mut Criterion) {
```
vmx, line 160: Are you sure you want to remove this one? I'd rather add this to the `criterion_group!()` call below. You could check with the original author, if it was a mistake not adding it, or if it should be removed.

storage-proofs/core/src/crypto/pedersen.rs
```diff
@@ -405,8 +405,8 @@ mod tests {
             let hashed = pedersen_md_no_padding(&flat);
 
             let mut hasher = Hasher::new(&x[0]).unwrap();
-            for k in 1..5 {
-                hasher.update(&x[k]).unwrap();
+            for val in x.iter().take(5).skip(1) {
```
vmx, line 408: Just a personal nitpick (no need to change it). I would write it as `x.iter().skip(1).take(5)`. I needed a while to parse this one, I read the current version as "take the first five elements, ohhh wait, skip the first one".

storage-proofs/core/benches/drgraph.rs
tcharding: I initially thought that (even had that written in the commit log at one stage), also during the remove unwrap work this came up. `parents` is a trait method, I didn't dig into it but assumed the result was used by other implementers of the the trait, otherwise why would it be there in the first place? I can definitely look into this further now you mentioned it though.

storage-proofs/core/benches/pedersen.rs
tcharding: Good point, @laser `git blame` highlights you as the developer who added the macro call `criterion_group`. Do you remember if there was a reason why `pedersen_md_circuit_benchmark()` was not added to it please?

storage-proofs/core/src/crypto/pedersen.rs
tcharding: Agreed.  Highlighting that you are correct is the fact that the `5` is misleading because you translated it to `x.iter().skip(1).take(5)` which iis different, we actually only want 4 items. I'll use `x.iter().skip(1).take(4)` for this and I'll use this order for this sort of combinator in future, thanks for the lesson.

tcharding:
Force pushed the combinator change, will wait on further comment before fixing the unused function change. Please do not merge (do you guys have a process for this, would you prefer me to change this to a draft PR to sort it out or is the discussion thread enough?) Thanks.

tcharding:
While we are on the topic of process, do you guys have a preference for force pushing / not force pushing? Do you follow the 'don't change history' philosophy or the 'maintain clean discrete patch sets' philosophy?

tcharding:
Oh, and thanks for the review @vmx, appreciate it.

laser:
Hi Tobin,

Nope - I have no idea why it wasn’t added. Probably an accidental omission / oversight. 

Erin

vmx:
> Do you follow the 'don't change history' philosophy or the 'maintain clean discrete patch sets' philosophy?

I can't speak for the whole team. But I personally prefer clean discrete patch sets. Though it also has it's limits. If you e.g. do a rebase and a change and force push, it might make it hard for reviewers what actually changes. What I do in such cases is that I just push a commit with review comments addressed and once everything is reviewed a forced push with a clean history.

Havind said that, we also happily squash commits, so feel free to just keep adding things and we'll do a squash commit at the end.

tcharding:
> Hi Tobin, Nope - I have no idea why it wasn’t added. Probably an accidental omission / oversight. Erin

Cool, thanks.

tcharding:
> > Do you follow the 'don't change history' philosophy or the 'maintain clean discrete patch sets' philosophy?
> 
> I can't speak for the whole team. But I personally prefer clean discrete patch sets. Though it also has it's limits. If you e.g. do a rebase and a change and force push, it might make it hard for reviewers what actually changes. What I do in such cases is that I just push a commit with review comments addressed and once everything is reviewed a forced push with a clean history.
> 
> Havind said that, we also happily squash commits, so feel free to just keep adding things and we'll do a squash commit at the end.

Awesome, thanks.

tcharding:
I have updated the patch order, and amended the `criterion_group` patch based on responses above. Force pushed the whole thing and updated the initial PR comment to reflect this.

Thanks.

tcharding:
Thanks for the explanation @vmx, no worries at all. No rush, when you are ready just holla at me and I can rebase. I'm going to keep chipping away at little issues as I can, if there is anyway I can make your lives (as reviewers) easier just let me know.

tcharding:
lol you beat me to it, was just working on it now :)

PR #1149, opened 2020-06-03
fix: run on stable rust and update CI

vmx:
The CI is now testing only the version that is specified in the `rust_toolchain` file.
Previously that version plus stable was tested, but now as we are on stable, there is
not need to run tests twice.

The `--all` flag for `cargo` is deprecated, instead one should use `--workspace`. This
commit also makes this adjustment.


vmx:
@cryptonemo I think the FFI is happy, I removed those targets with this commit, so it can well be that those are kind of screwed up.

If you think it's ready to be merged, could you please merge this one together with https://github.com/filecoin-project/rust-filecoin-proofs-api/pull/20 and https://github.com/filecoin-project/filecoin-ffi/pull/102, you know better than I what other things are going on and when is a good time to do the merge.

PR #1153, opened 2020-06-05
Audit: Update total challenge count and increase partitions.

porcuquine:
The internal audit found we should be using more total challenges. This PR raises the minimum challenges to 180 and the partitions to 10, for both 32GiB and 64GiB sectors. The partition count increases so circuit size will not.


cryptonemo:
@porcuquine I'll approve once rebased

PR #1154, opened 2020-06-05
Build and load columns in separate threads.

porcuquine:
@cryptonemo I can't test this, because for some reason I get an error when running with `FIL_PROOFS_USE_GPU_COLUMN_BUILDER` set. This happens even on `master` when I run on the machine I use for GPU testing. So it's likely this doesn't work quite right. [To clarify: I mean likely this PR doesn't work right. I don't have any reason to assume the problem I'm seeing isn't with my environment — though it is puzzling and annoying.]

Nevertheless, I think something generally like this should (might?) work to eliminate the serialization of loading and tree building which seems to be happening now.


porcuquine:
Closed in favor of #1155.

PR #1158, opened 2020-06-07
feat: publish v27 params

cryptonemo:


cryptonemo:
So looking at the hashes, something seems to have gone wrong?

cryptonemo:
> So looking at the hashes, something seems to have gone wrong?

Or is it the case that the hash in the parameter file name only changes on tree shape or hasher change?

> So looking at the hashes, something seems to have gone wrong?

Oh, nevermind, it looks like that hash doesn't change based on the identifier not changing.  Please sanity check that though.

porcuquine:
As far as I can tell, this is all as expected. However, I can't parse exactly what you wrote well enough to also confirm your logic. To state mine: I expect both the digest and the CID to change — since changed file content should affect both.

cryptonemo:
Thanks for checking, yes, I think we're seeing the same expected results.  I had a minor panic earlier, but it all looked good when I tracked it down further.

I just rebased this against neptune, so I think after this is merged, we're ready for the final release PR.

dignifiedquire:
Looks like I would expect it to look like. The identifiers shouldn't change, unless we changed parameters of the trees

PR #1165, opened 2020-06-15
fix: change default rows_to_discard for oct trees

cryptonemo:
feat: move default value from merkle_light to internal

### DIFF
--- a/storage-proofs/core/src/settings.rs
+++ b/storage-proofs/core/src/settings.rs
@@ -21,7 +21,7 @@ pub struct Settings {
     pub column_write_batch_size: u32,
     pub use_gpu_tree_builder: bool,
     pub max_gpu_tree_batch_size: u32,
-    pub oct_rows_to_discard: u32,
+    pub rows_to_discard: u32,
 }
 
 impl Default for Settings {
@@ -34,7 +34,7 @@ impl Default for Settings {
             column_write_batch_size: 262_144,
             use_gpu_tree_builder: false,
             max_gpu_tree_batch_size: 700_000,
-            oct_rows_to_discard: 2,
+            rows_to_discard: 2,
         }
     }
 }
--- a/storage-proofs/core/src/settings.rs
+++ b/storage-proofs/core/src/settings.rs
@@ -21,7 +21,7 @@ pub struct Settings {
     pub column_write_batch_size: u32,
     pub use_gpu_tree_builder: bool,
     pub max_gpu_tree_batch_size: u32,
-    pub rows_to_discard: u32,
+    pub oct_rows_to_discard: u32,
 }
 
 impl Default for Settings {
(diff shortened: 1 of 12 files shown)


cryptonemo:
An alternative to https://github.com/filecoin-project/merkle_light/pull/88, giving this library fuller control over the values going forward.

storage-proofs/core/src/settings.rs
```diff
@@ -21,7 +21,7 @@ pub struct Settings {
     pub column_write_batch_size: u32,
     pub use_gpu_tree_builder: bool,
     pub max_gpu_tree_batch_size: u32,
-    pub rows_to_discard: u32,
+    pub oct_rows_to_discard: u32,
```
porcuquine, line 24: I think we don't need to expose the arity here, since we use oct everywhere we discard tree rows.

storage-proofs/core/src/util.rs
```diff
@@ -142,6 +145,36 @@ pub fn reverse_bit_numbering(bits: Vec<boolean::Boolean>) -> Vec<boolean::Boolea
...
+    let rows_to_discard = settings::SETTINGS.lock().unwrap().oct_rows_to_discard as usize;
+
+    // Discard at most 'constant value' rows (coded below,
+    // differing by arity) while respecting the max number that
+    // the tree can support discarding.
+    match arity {
+        2 => std::cmp::min(max_rows_to_discard, 7),
+        4 => std::cmp::min(max_rows_to_discard, 5),
+        _ => std::cmp::min(max_rows_to_discard, rows_to_discard),
+    }
```
porcuquine, line 175: Along with a single `rows_to_discard` default, we could probably get rid of this match — unless it simplifies testing.

storage-proofs/porep/src/stacked/vanilla/proof.rs
```diff
@@ -973,34 +973,31 @@ impl<'a, Tree: 'static + MerkleTreeTrait, G: 'static + Hasher> StackedDrg<'a, Tr
...
             CacheKey::CommRLastTree.to_string(),
             Some(get_merkle_tree_len(nodes_count, Tree::Arity::to_usize())?),
         );
 
-        // A default 'rows_to_discard' value will be chosen for
-        // tree_r_last, unless the user overrides this value via the
-        // environment setting (FIL_PROOFS_ROWS_TO_DISCARD).  If this
-        // value is specified, no checking is done on it and it may
+        // A default 'rows_to_discard' value will be chosen for tree_r_last, unless the user overrides this value via the
+        // environment setting (FIL_PROOFS_OCT_ROWS_TO_DISCARD).  If this value is specified, no checking is done on it and it may
```
porcuquine, line 985: It's still not clear to me whether this is converging on what I think the right answer should be:

- If nothing is specified by 'settings', default to discarding zero rows.
- If a setting is specified by the user (env var or config file), use that.
- If nothing is specified by the user, use a default hard-coded in settings.

cryptonemo: This doesn't address the first case at all, as something is always specified by the settings (either the default value, or a user provided value).  The last 2 cases should be satisfied.

PR #1169, opened 2020-06-16
feat(storage-proofs): NSE - gpu implementation of encode_with_trees

keyvank:


storage-proofs/porep/src/nse/vanilla/labels.rs
```diff
@@ -493,6 +495,66 @@ fn tree_from_slice<Tree: MerkleTreeTrait>(
...
+        let store_config = store_configs.remove(0);
+        let tree_data: Vec<u8> = lo
+            .base
+            .0
+            .iter()
+            .chain(lo.tree.iter())
+            .flat_map(|node| fr_into_bytes(&node.0))
+            .collect();
+        trees.push(
+            MerkleTree::<GPUTree>::from_tree_slice_with_config(
```
keyvank, line 546: @dignifiedquire This is not working properly atm, how can I build a `MerkleTree::<GPUTree>` given a `Vec<Fr>` returned from Neptune? (And what should I set `discard_rows` in case I want to do this?

storage-proofs/porep/src/nse/vanilla/labels.rs
```diff
@@ -486,17 +495,144 @@ fn lc_tree_from_slice<Tree: MerkleTreeTrait>(
...
+        || pool.is_err()
+    {
+        debug!("encoding trees on cpu");
+
+        inps.map(|(store_configs, window_index, replica_id, data)| {
+            encode_with_trees::<Tree>(conf, store_configs, window_index, replica_id, data)
+        })
+        .collect()
+    } else {
+        debug!("encoding trees on gpu");
```
porcuquine, line 565: This block is a bit of a monster. It's probably worth *at least* separating it out into something like `encode_with_trees_gpu` to make the structure clear.

storage-proofs/porep/src/nse/vanilla/labels.rs
```diff
@@ -486,17 +495,144 @@ fn lc_tree_from_slice<Tree: MerkleTreeTrait>(
...
+        .map_err(|e| -> gpu::NSEError { e.into() })
+        .and_then(|devs| {
+            Ok(gpu::SealerPool::new(
+                devs,
+                gpu_conf,
+                gpu::TreeOptions::Enabled { rows_to_discard },
+            )?)
+        });
+
+    if !settings::SETTINGS.lock().unwrap().use_gpu_nse
```
porcuquine, line 554: Instead of checking the setting value here, we should pass it as an input to `encode_with_trees_all`. We might need another wrapper function just for that purpose. This will make it easier to test both cases for equivalence.

storage-proofs/porep/src/nse/vanilla/labels.rs
```diff
@@ -701,4 +837,80 @@ mod tests {
             ]
         );
     }
+
+    #[test]
+    fn test_gpu_cpu_consistency() {
+        let rng = &mut XorShiftRng::from_seed(crate::TEST_SEED);
```
porcuquine, line 843: This test is very repetitive, which makes it hard to verify that CPU and GPU code paths are equivalent.

Can you restructure (perhaps including some of the suggested refactorings of the implementation) so it becomes obvious?

Something roughly like this would be good:

```
let test_fn = |use_gpu, …| { … };

let (a_cpu, b_cpu, c_cpu) = test_fn(false);
let (a_gpu, b_gpu, c_gpu) = test_fn(true);

assert_eq!(a_cpu, a_gpu);
assert_eq!(b_cpu, b_gpu);
assert_eq!(c_cpu, c_gpu);
```

storage-proofs/porep/src/nse/vanilla/labels.rs
```diff
@@ -486,17 +495,144 @@ fn lc_tree_from_slice<Tree: MerkleTreeTrait>(
...
+        Item = (
+            Vec<StoreConfig>,
+            u32,
+            &'a <Tree::Hasher as Hasher>::Domain,
+            &'a mut [u8],
+        ),
+    >,
+{
+    let gpu_conf = to_gpu_config(conf);
+    let pool = gpu::utils::all_devices()
```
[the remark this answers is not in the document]
keyvank: hmm i don't see a `gpu` feature in `porep` crate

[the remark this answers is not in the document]
porcuquine: I mean if the setting doesn't specify to use GPU for NSE.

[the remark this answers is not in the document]
keyvank: oh i see, ok

dignifiedquire:
Merged manualy

PR #1170, opened 2020-06-16
feat: rebuild tree_r_last trees from a replica file

cryptonemo:
This PR adds a tool for rebuilding tree_r_last trees from a replica file.  The intended usage is for replacing existing trees with an updated `rows_to_discard` setting, as trees with different values for that setting are incompatible.

OPTIONAL (one time sanity check, but can be skipped entirely): To verify that the tool works as expected, let's use the 'old' default of rows_to_discard as 3 for 32GiB sectors and simply rebuild them to make sure the data produced exactly matches what's already present:

```
time RUST_LOG=trace RUSTFLAGS="-C target-cpu=native" FIL_PROOFS_ROWS_TO_DISCARD=3 cargo run --release --bin update_tree_r_cache rebuild --size 34359738368 --cache /some/new/output/dir --replica /path/to/existing/sealed-file
```

OPTIONAL (continued): After this completes, use md5sum to match the tree_r_last files produced in `/some/new/output/dir` against the ones already present in the path containing the existing tree_r_last files.  If these match exactly, the real conversion step is run as follows:

Example usage:

```
time RUST_LOG=trace RUSTFLAGS="-C target-cpu=native" FIL_PROOFS_ROWS_TO_DISCARD=2 cargo run --release --bin update_tree_r_cache rebuild --size 34359738368 --cache /some/new/output/dir --replica /path/to/existing/sealed-file
```

Note that in this case, `FIL_PROOFS_ROWS_TO_DISCARD` is changed from 3 to 2.  Alternatively, you could unset that env var, as 2 is the new default value.

The output trees produced in this step should be bigger, as less data is discarded.  While it's recommended that you preserve the old tree_r_last files (just until we're sure everything starts up properly with the new code), the newly produced files should be placed where the 'old' ones were present.

### DIFF
--- a/fil-proofs-tooling/Cargo.toml
+++ b/fil-proofs-tooling/Cargo.toml
@@ -48,6 +48,7 @@ bytefmt = "0.1.7"
 rayon = "1.3.0"
 flexi_logger = "0.14.7"
 typenum = "1.11.2"
+generic-array = "0.13.2"
 
 [features]
 default = ["gpu", "measurements"]
--- a/fil-proofs-tooling/src/bin/update_tree_r_cache/main.rs
+++ b/fil-proofs-tooling/src/bin/update_tree_r_cache/main.rs
@@ -1,0 +1,136 @@
+use std::fs::{create_dir_all, OpenOptions};
+use std::path::{Path, PathBuf};
+
+use anyhow::{Context, Result};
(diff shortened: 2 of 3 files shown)


porcuquine:
FYI, @travisperson.

Issue #943, opened 2019-11-12
More gracefully handling out of disk space errors


whyrusleeping:
Currently, the code panics if it hits an out of disk space error. We can all agree that this isnt the right way to handle the problem. 

Simply returning the error back up to the user isnt the right thing to do either. The right way to handle this would be to wrap all your write operations with something that preallocates the length of the write, and if that fails *truncate* the file back down to the original size. truncation is important here, and not just deleting the file because CoW filesystems actually require additional space to perform a delete, where truncations free up space first. This doesnt *always* work, but it should tend towards not leaving the filesystem in a super borked state most of the time.

dignifiedquire:
The way we should likely implement this is by creating a wrapper type around `std::fs::File`, which roughly looks like this


```rust
pub struct SafeFile(std::fs::File);

impl SafeFile {
    pub fn new(size: usize) -> Self {
    }
}

**Comment by dignifiedquire** [comment]
> preallocates the length of the write, 

Pretty much all of our operations already do this, the issue is that the creation in the first place fails for the most part.

proposed construction

```rust

pub trait FileExt {
  fn set_len_safe(&self, len: usize) -> io::Result<()>;
}

impl FileExt for std::fs::File {
  fn set_len_safe(&self, len: usize) -> io::Result<()> {
    // 1. stat for current len
	// 2. try calling `set_len` with new len
    // 3. if it try `set_len` with old len
  }
}
```

dignifiedquire:
Files that we create in the `tmpfile` already get cleaned up using OS level calls, so that should not be an issue: see https://docs.rs/tempfile/ for details

tcharding:
Noob question: do folks run this software as non-priveleged users? If so we could wrap all writes with a call to `fs2::available_space()`, and return early if there is not enough space on the filesystem to do the write.

As well as the suggested trait extension above we will likely need a wrapper for `fs::copy()`.

I came up with this while trying to find out if the `Metadata`  method `set_len()` calls `fallocate`, I was unable to ascertain this.

ref: https://docs.rs/fs2/0.4.3/fs2/

cryptonemo:
Unclear if still needed and prior work on this was discarded in #1171.  Open a new issue if still required.

PR #1171, opened 2020-06-17
Safe write

tcharding:
Makes an effort at adding safe write functionality. Draft because currently this does not cover _all_ writes in the repo.

Adds a trait
```
pub trait SafeFileExt {
    fn safe_write_all(&mut self, buf: &[u8]) -> io::Result<()>;
    fn safe_set_len(&self, size: u64) -> io::Result<()>;
}
```
And also a `io::copy` wrapper
```
pub fn safe_copy<P: AsRef<Path>, Q: AsRef<Path>>(from: P, to: Q) -> io::Result<u64>
```

Done as part of resolving: https://github.com/filecoin-project/rust-fil-proofs/issues/943

### DIFF
--- a/filecoin-proofs/src/api/seal.rs
+++ b/filecoin-proofs/src/api/seal.rs
@@ -1,5 +1,4 @@
 use std::fs::{self, metadata, File, OpenOptions};
-use std::io::prelude::*;
 use std::path::{Path, PathBuf};
 
 use anyhow::{ensure, Context, Result};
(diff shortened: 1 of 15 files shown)


tcharding:
Closing. I've been thinking about this for a while and its not sitting well. I do not think this is the correct place to be doing this 'safe' check. The reason is that many writes are done via a parameter that implements `Write`, this makes it the destination opaque.  Having a 'safe' write just for cases where we know we have a file seems like wrong thing to do since it does not do the job fully. 

I think this issue (disk getting filled) would be better handled by some tool outside of filecoin i.e., monitor the disk and then tell the filecoin process if disk space is low. I do not know if the filecoin software can run with different disk usage under different conditions which would be required for this to be any benefit I guess.

PR #1174, opened 2020-06-18
Synthesize Window PoSt circuit in parallel when generating Groth proofs.

porcuquine:
This just needs dependencies updated after https://github.com/filecoin-project/bellman/pull/86 merges and the intermediate packages are updated/released.


dignifiedquire:
needs rebase

Cargo.toml
```diff
@@ -9,3 +9,7 @@ members = [
   "fil-proofs-tooling",
   "sha2raw"
 ]
+
+[patch.crates-io]
```
dignifiedquire, line 13: (just noting so it doesn't get lost) this needs to be removed before merging

PR #1175, opened 2020-06-18
feat: extend update_tree_r_cache command with new features

cryptonemo:
Added are 'inspect' and 'verify' options.  Inspect makes sure the cache is consistent with itself and verify rebuilds trees and then verifies against the cached data (which can detect an error in the replica data).

### DIFF
--- a/fil-proofs-tooling/src/bin/update_tree_r_cache/main.rs
+++ b/fil-proofs-tooling/src/bin/update_tree_r_cache/main.rs
@@ -1,20 +1,28 @@
-use std::fs::{create_dir_all, OpenOptions};
+use std::fs::{create_dir_all, remove_dir_all, OpenOptions};
 use std::path::{Path, PathBuf};
 
-use anyhow::{Context, Result};
+use anyhow::{ensure, Context, Result};
+use bincode::deserialize;
 use clap::{value_t, App, Arg, SubCommand};
-use generic_array::typenum::{U0, U8};
+use generic_array::typenum::{self, Unsigned};
 use memmap::MmapOptions;
-use merkletree::store::StoreConfig;
+use merkletree::merkle::get_merkle_tree_len;
+use merkletree::store::{ExternalReader, ReplicaConfig, Store, StoreConfig};
+use tempdir::TempDir;
+use typenum::{U0, U8};
 
 use filecoin_proofs::constants::*;
 use filecoin_proofs::types::*;
 use storage_proofs::cache_key::CacheKey;
-use storage_proofs::merkle::LCTree;
-use storage_proofs::merkle::{get_base_tree_count, split_config_and_replica};
+use storage_proofs::merkle::{create_lc_tree, get_base_tree_count, split_config_and_replica};
+use storage_proofs::merkle::{LCStore, LCTree, MerkleTreeTrait};
 use storage_proofs::util::{default_rows_to_discard, NODE_SIZE};
 
-fn run_update(sector_size: usize, cache: PathBuf, replica_path: PathBuf) -> Result<()> {
+fn get_tree_r_info(
+    sector_size: usize,
+    cache: &PathBuf,
+    replica_path: &PathBuf,
+) -> Result<(usize, usize, Vec<StoreConfig>, ReplicaConfig)> {
     let tree_count = match sector_size as u64 {
         SECTOR_SIZE_2_KIB => get_base_tree_count::<SectorShape2KiB>(),
         SECTOR_SIZE_4_KIB => get_base_tree_count::<SectorShape4KiB>(),
@@ -25,8 +33,9 @@ fn run_update(sector_size: usize, cache: PathBuf, replica_path: PathBuf) -> Resu
         SECTOR_SIZE_64_GIB => get_base_tree_count::<SectorShape64GiB>(),
         _ => panic!("Unsupported sector size"),
     };
+
     // Number of nodes per base tree
-    let nodes_count = sector_size / NODE_SIZE / tree_count;
+    let base_tree_leafs = sector_size / NODE_SIZE / tree_count;
 
     // If the cache dir doesn't exist, create it
     if !Path::new(&cache).exists() {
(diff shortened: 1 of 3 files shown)


dignifiedquire:
I am sorry, the CI seems to be flaky since adding the new caching :( I still quite understand why though :/

cryptonemo:
> I am sorry, the CI seems to be flaky since adding the new caching :( I still quite understand why though :/

Ah, I was sanity checking here, and of course things pass locally.  Good to know it's not just me then :-)

fil-proofs-tooling/src/bin/update_tree_r_cache/main.rs
```diff
@@ -64,27 +132,206 @@ fn run_update(sector_size: usize, cache: PathBuf, replica_path: PathBuf) -> Resu
...
+
+    println!("CommRLast from p_aux: {:?}", p_aux.comm_r_last);
+    println!(
+        "CommRLast [cached tree_r_last root]: {:?}",
+        tree_r_last_root
+    );
+    let status = if tree_r_last_root == p_aux.comm_r_last {
+        "MATCH"
+    } else {
+        "MIS-MATCH"
```
porcuquine, line 232: Any chance I can convince you to use the word "MISMATCH"?

fil-proofs-tooling/src/bin/update_tree_r_cache/main.rs
```diff
@@ -64,27 +132,206 @@ fn run_update(sector_size: usize, cache: PathBuf, replica_path: PathBuf) -> Resu
...
+        let p_aux_path = cache.join(CacheKey::PAux.to_string());
+        let p_aux_bytes = std::fs::read(&p_aux_path)
+            .with_context(|| format!("could not read file p_aux={:?}", p_aux_path))?;
+
+        deserialize(&p_aux_bytes)
+    }?;
+
+    // Rebuild each of the tree_r_last base trees (in a new temp dir so as not to interfere
+    // with any existing ones on disk) and check if the roots match what's cached on disk
+    let tmp_dir = TempDir::new("tree-r-last-verify")?;
```
porcuquine, line 279: I might not completely follow the intended usage, so this question may not be quite right.

Under what circumstances would I want to go to the trouble of rebuilding the tree — in order to check the stored versions of the tree — but then discard the tree if the stored one is wrong?

If the answer is, "You wouldn't. If there is a mismatch, we will repair the existing trees by replacing with the correct ones just generated," then great. If not, perhaps we should consider that?

We should at least make sure there's no intended workflow in which a problem would be detected as above, and the solution would then involve rebuilding the tree *again* to replace.

porcuquine:
Regarding CI failures: I lack the context to know why we should feel comfortable ignoring what have the appearance of real test failures, which — if tests are accurate — would be significant.

I don't think we should develop a habit of skipping or removing tests because they are failing. If we start down that road, it will be very hard to recover if a real problem goes undetected.

Have we seen these failures elsewhere? We need to diagnose/debug this unless we have high confidence about what is going on and why it's not an immediate problem. (I can accept this might be true if someone presents the case to me.)

porcuquine:
After reading previous comments more closely: if the new caching made CI flaky, we should consider reverting the new caching until we can ascertain that this is a CI problem (and fix it). If there is any chance that the new caching is itself flaky and this is being exposed by the tests, we should *not* release it. If we want to claim this is not possible, we should consider what method we *would* use to detect intermittent incorrectness of the caching code.

fil-proofs-tooling/src/bin/update_tree_r_cache/main.rs
cryptonemo: This is a verify command (separate from rebuild).

If the replica data is bad and the cached trees mis-match, there's no point in rebuilding at all.  

This command verifies that the replica data has not changed and that the cached trees match based on that assumption.  If there's a mis-match, we don't know which is correct, but will report that mis-match.

fil-proofs-tooling/src/bin/update_tree_r_cache/main.rs
```diff
@@ -64,27 +132,206 @@ fn run_update(sector_size: usize, cache: PathBuf, replica_path: PathBuf) -> Resu
...
+        SECTOR_SIZE_32_GIB => {
+            let (tree_r_last, base_tree_roots) =
+                build_tree_r_last::<SectorShape32GiB>(sector_size, &cache, &replica_path)?;
+            (tree_r_last.root(), base_tree_roots)
+        }
+        SECTOR_SIZE_64_GIB => {
+            let (tree_r_last, base_tree_roots) =
+                build_tree_r_last::<SectorShape64GiB>(sector_size, &cache, &replica_path)?;
+            (tree_r_last.root(), base_tree_roots)
+        }
```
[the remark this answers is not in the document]
cryptonemo: Can look at, forgot about that macro.

filecoin-proofs/src/constants.rs
```diff
@@ -128,6 +128,13 @@ pub type DefaultBinaryTree = storage_proofs::merkle::BinaryMerkleTree<DefaultTre
 pub type DefaultOctTree = storage_proofs::merkle::OctMerkleTree<DefaultTreeHasher>;
 pub type DefaultOctLCTree = storage_proofs::merkle::OctLCMerkleTree<DefaultTreeHasher>;
 
+// Generic shapes
+pub type SectorShapeBase = LCTree<DefaultTreeHasher, U8, U0, U0>;
+pub type SectorShapeSub2 = LCTree<DefaultTreeHasher, U8, U2, U0>;
+pub type SectorShapeSub8 = LCTree<DefaultTreeHasher, U8, U8, U0>;
+pub type SectorShapeTop2 = LCTree<DefaultTreeHasher, U8, U8, U2>;
+
```
porcuquine, line 136: The one further step I would take is to use these types in the definitions directly below. That will also make the correctness of the following predicates easiest to verify and keep in sync.

filecoin-proofs/src/constants.rs
```diff
@@ -134,17 +134,20 @@ pub type SectorShapeSub2 = LCTree<DefaultTreeHasher, U8, U2, U0>;
...
+pub type SectorShape4KiB = SectorShapeSub2;
+pub type SectorShape16MiB = SectorShapeSub2;
+pub type SectorShape1GiB = SectorShapeSub2;
+
+pub type SectorShape16KiB = SectorShapeSub8;
+pub type SectorShape32GiB = SectorShapeSub8;
+
+pub type SectorShape32KiB = SectorShapeTop2;
+pub type SectorShape64GiB = SectorShapeTop2;
```
porcuquine, line 150: Perfect, so much better. Thank you.

PR #1179, opened 2020-06-19
fix(storage-proofs-porep): use file locking for cache generation

dignifiedquire:


storage-proofs/core/src/parameter_cache.rs
```diff
@@ -25,7 +25,7 @@ pub const PARAMETER_METADATA_EXT: &str = "meta";
 pub const VERIFYING_KEY_EXT: &str = "vk";
 
 #[derive(Debug)]
-struct LockedFile(File);
+pub struct LockedFile(File);
```
porcuquine, line 28: This might belong in a more general place if being used elsewhere now, too. Let's consider moving later.

PR #1186, opened 2020-06-24
fix: allow parameters map to be accessible externally

cryptonemo:
The motivation for this is because of a broken test that attempts to use this from rust-filecoin-proofs-api, which always fails due to the parameter file not being available

### DIFF
--- a/filecoin-proofs/src/constants.rs
+++ b/filecoin-proofs/src/constants.rs
@@ -28,9 +28,11 @@ pub const WINDOW_POST_CHALLENGE_COUNT: usize = 10;
 pub const DRG_DEGREE: usize = storage_proofs::drgraph::BASE_DEGREE;
 pub const EXP_DEGREE: usize = storage_proofs::porep::stacked::EXP_DEGREE;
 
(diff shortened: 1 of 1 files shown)


filecoin-proofs/src/constants.rs
```diff
@@ -192,6 +194,19 @@ fn parameter_id(cache_id: &str) -> String {
...
 }
 
+/// Get the correct verifying key data for a given cache id.
+pub fn get_verifying_key_data(cache_id: &str) -> Option<&ParameterData> {
+    PARAMETERS.get(&verifying_key_id(cache_id))
+}
+
+fn verifying_key_id(cache_id: &str) -> String {
+    format!(
+        "v{}-{}.vk",
```
porcuquine, line 204: Does this format need to be consistent with one from elsewhere? If so, a comment noting where so they can be kept in sync (or a refactor to enforce it) might be helpful.

cryptonemo: No, nothing is exposed how it's needed here, but could be part of a future re-factor.

PR #1187, opened 2020-06-25
Implement Circuit and CompoundProof for NSE

dignifiedquire:
- [x] Scaffolding
	- [x] Circuit
	- [x] CompoundProof
	- [x] Tests
- [x] Implementation
  - [x] Shared components (`comm_r`, `replica_id`, ..)
  - [x] LayerProof
  - [x] NodeProof


storage-proofs/porep/src/nse/circuit/compound.rs
```diff
@@ -25,11 +28,129 @@ impl<'a, Tree: 'static + MerkleTreeTrait, G: 'static + Hasher>
...
+                let c = layer_challenge.first_layer_challenge.absolute_index;
+
+                // challenge input
+                inputs.push(u64_into_fr(c));
+
+                // comm_d inclusion proof for the data leaf
+                inputs.extend(data_inclusion_inputs(c)?);
+
+                // layer_inclusion proof
+                inputs.extend(layer_inclusion_inputs(c)?);
```
porcuquine, line 81: @dignifiedquire This is somewhat redundant, right? This will result in `c` being passed three times, I think. I am not saying it is necessarily convenient to do otherwise (although it might be worth trying) — just checking to make sure I am tracking what you have here.

storage-proofs/porep/src/nse/vanilla/labels.rs
```diff
@@ -214,7 +214,9 @@ fn mask_layer<D: Domain>(
         .par_chunks_mut(NODE_SIZE)
         .enumerate()
         .for_each(|(node_index, node)| {
-            let prefix = hash_prefix(LAYER_INDEX, node_index as u32, window_index);
+            let node_absolute_index =
+                window_index as u64 * config.num_nodes_window as u64 + node_index as u64;
+            let prefix = hash_prefix(LAYER_INDEX, node_absolute_index);
             let hash = Sha256::digest(&[&prefix[..], AsRef::<[u8]>::as_ref(replica_id)]);
```
porcuquine, line 220: @dignifiedquire Is this a change to the actual encoding/sealing results? Changed tests below suggest it will be.

If that is correct, why is such a change introduced in the circuit implementation? Is this just an independent change/fix being  bundled here for convenience?

(cc: @keyvank in case the GPU implementation will need to change to agree with this)

storage-proofs/porep/src/nse/circuit/hash.rs
```diff
@@ -0,0 +1,276 @@
...
+        &bits[0..(Fr::CAPACITY as usize)],
+    )
+}
+
+/// Calculates the common prefix for the hash.
+fn hash_prefix(challenge_index: &UInt64, layer: UInt32) -> Result<Vec<Boolean>, SynthesisError> {
+    // The prefix is 32 bytes long
+    const PREFIX_LEN: usize = 32 * 8;
+
+    let mut ciphertexts = Vec::with_capacity(PREFIX_LEN);
```
porcuquine, line 131: Why is this called `ciphertexts`?

storage-proofs/porep/src/nse/circuit/hash.rs
```diff
@@ -0,0 +1,276 @@
...
+    mut cs: CS,
+    replica_id: &[Boolean],
+    challenge_index_num: &UInt64,
+    layer: u32,
+) -> Result<num::AllocatedNum<Bls12>, SynthesisError> {
+    // TODO: can we reuse this allocation accross challenges?
+    let layer_index = UInt32::alloc(cs.namespace(|| "layer_index"), Some(layer))?;
+
+    // prefix
+    let mut ciphertexts = hash_prefix(challenge_index_num, layer_index)?;
```
porcuquine, line 24: Is the word 'ciphertext' being used here for any reason other than because it has gradually leaked into the code base over time? I think it was originally used for the output of Sloth, which later became just XOR. In that context, it still made sense because a key was being used to transform node values.

By SDR, this was no longer really what was happening, but I think the name stuck around. Now it seems like we're using it for hash input.

I've recapitulated a probably poor history as a guess about how this naming came to be. If there's a different reason for it, please let me know. Otherwise, I propose we adopt new and less confusing naming moving forward.

storage-proofs/porep/src/nse/vanilla/proof_scheme.rs
```diff
@@ -339,7 +349,9 @@ impl<'a, 'c, Tree: 'static + MerkleTreeTrait, G: 'static + Hasher> ProofScheme<'
         _requirements: &ChallengeRequirements,
         _partitions: usize,
     ) -> bool {
-        todo!()
+        // TODO:
```
porcuquine, line 352: This is still missing and needs to be addressed, although not necessarily in this circuit PR.

storage-proofs/porep/src/nse/circuit/circuit.rs
```diff
@@ -21,6 +38,471 @@ impl<'a, Tree: 'static + MerkleTreeTrait, G: 'static + Hasher> Circuit<Bls12>
...
+                || Ok(Fr::zero()),
+            )?);
+        }
+
+        // Verify hash(comm_layers) == comm_r
+        {
+            let hash_num = PoseidonFunction::hash_md_circuit::<_>(
+                &mut cs.namespace(|| "comm_layers_hash"),
+                &comm_layers_nums_padded,
+            )?;
```
porcuquine, line 112: BLOCKING: Whoops, per Poseidon audit, I think we shouldn't be using the Poseidon HashMD function as-is. We can probably just use the specific arity needed (is it 14, 15?) instead. Alternately, if we need variable length (which I don't think is the case here), we can use one of the other constructions the auditors proposed — one of which might include a small tweak to the extant MD function. I don't have access to the document at the moment, but will look into this and deal with it next week.

storage-proofs/porep/src/nse/vanilla/proof_scheme.rs
dignifiedquire: Yes, lets do this at a later point, I only switched to a manual todo, to be able to test the rest.

storage-proofs/porep/src/nse/circuit/circuit.rs
dignifiedquire: > Whoops, per Poseidon audit, I think we shouldn't be using the Poseidon HashMD function as-is.

:(  that is unfortunate, yes the hashing is a specific arity, but we do vary that in the tests quite a bit currently. But technically we can fix it to either a hardcoded number, or simply add a method which uses an existing artiy up to the one we need.

porcuquine: I'll look at the options and try to implement one of those which accommodates variable length. I think the easiest version is not bad, and the other might be worth doing too. Understood about the tests — makes sense.

storage-proofs/porep/src/nse/circuit/hash.rs
dignifiedquire: legacy, and because I like it as a name for sth encoded

storage-proofs/porep/src/nse/vanilla/labels.rs
dignifiedquire: - yes this is a change to the actual hashing
- this is a bug fix
- it is bundled with this, as I realised this bug when trying to implement the circuit, and couldn't move forward without fixing it

storage-proofs/porep/src/nse/circuit/compound.rs
dignifiedquire: While `c` is being used to generate the inputs, it is only allocated in the circuit once (the first one where we call `u64_into_fr` (as far as I remember)

storage-proofs/porep/src/nse/vanilla/labels.rs
porcuquine: Okay, thanks. That's what I thought. @keyvank please take note that when this PR merges, the GPU implementation will need to be updated to synchronize results again.

storage-proofs/porep/src/nse/circuit/circuit.rs
porcuquine: @dignifiedquire And I discussed this, with the resolution that we will change to use a fixed arity hash equal to the value actually required for real circuits (15). When fewer layers exist because an insecure circuit is used for faster/easier testing, we will hardcode the padding in the circuit.

porcuquine:
@dignifiedquire @cryptonemo I started trying to rebase this but suffer from not being very familiar with either side of the conflicts. Can one of you take this on so we can get this merged into the mainline NSE please?

storage-proofs/porep/src/nse/circuit/circuit.rs
porcuquine: After more discussion with Poseidon auditors, I will probably end up implementing a different scheme for this. That will require calling a new function here. @dignifiedquire I think you should proceed as planned, and when the new API function is available, I will replace the usage in rust-fil-proofs.

PR #1194, opened 2020-07-01
Add fauxrep to API for fake sealing.

porcuquine:
This PR adds a new API function, `fauxrep`, which allows fake sealing for testing purposes. `fauxrep` takes a `PoRepConfig`, a replica path, and a cache path — and returns a new, random-but-valid `CommR` and writes all artifacts needed to legitimately pass PoSt. No PoRep is performed, and no proof is returned. If `FIL_PROOFS_USE_GPU_TREE_BUILDER` is set, the replica tree will be built on GPU.

All fake replicas consist entirely of null bytes, and the implementation depends on this assumption for simplicity. This means all such replicas have interchangeable base layers, partial trees, and `comm_r_last` values. They differ only in the value of their `comm_c`, which is stored in the `p_aux` file. This value is randomly generated and is the reason fake replicas have unique `comm_r`s.

The PoST API tests have been enhanced to use both real and fake sealing. This preserves their value in establishing the correctness of seal and PoSt in conjunction — and also demonstrates that fauxrep produces sectors indistinguishable from those produced by seal as far as PoSt is concerned. (Of course, they do not come with a seal proof so are no good for mining.)

### DIFF
--- a/filecoin-proofs/src/api/seal.rs
+++ b/filecoin-proofs/src/api/seal.rs
@@ -758,3 +758,52 @@ pub fn verify_batch_seal<Tree: 'static + MerkleTreeTrait>(
     info!("verify_batch_seal:finish");
     result
 }
+
+pub fn fauxrep<R: AsRef<Path>, S: AsRef<Path>, Tree: 'static + MerkleTreeTrait>(
+    porep_config: PoRepConfig,
+    cache_path: R,
+    out_path: S,
+) -> Result<Commitment> {
+    let mut rng = rand::thread_rng();
+    fauxrep_aux::<_, R, S, Tree>(&mut rng, porep_config, cache_path, out_path)
+}
+
+pub fn fauxrep_aux<
+    Rng: rand::Rng,
+    R: AsRef<Path>,
+    S: AsRef<Path>,
+    Tree: 'static + MerkleTreeTrait,
+>(
+    mut rng: &mut Rng,
+    porep_config: PoRepConfig,
+    cache_path: R,
+    out_path: S,
+) -> Result<Commitment> {
+    let sector_bytes = PaddedBytesAmount::from(porep_config).0;
+
+    {
+        // Create a sector full of null bytes at `out_path`.
+        let file = File::create(&out_path)?;
+        file.set_len(sector_bytes)?;
+    }
+
+    let fake_comm_c = <Tree::Hasher as Hasher>::Domain::random(&mut rng);
+    let (comm_r, p_aux) = StackedDrg::<Tree, DefaultPieceHasher>::fake_replicate_phase2(
+        fake_comm_c,
+        out_path,
+        &cache_path,
+        sector_bytes as usize,
+    )?;
+
+    let p_aux_path = cache_path.as_ref().join(CacheKey::PAux.to_string());
+    let mut f_p_aux = File::create(&p_aux_path)
+        .with_context(|| format!("could not create file p_aux={:?}", p_aux_path))?;
+    let p_aux_bytes = serialize(&p_aux)?;
+    f_p_aux
+        .write_all(&p_aux_bytes)
+        .with_context(|| format!("could not write to file p_aux={:?}", p_aux_path))?;
+
+    let mut commitment = [0u8; 32];
+    commitment[..].copy_from_slice(&comm_r.into_bytes()[..]);
+    Ok(commitment)
+}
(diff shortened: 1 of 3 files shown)


filecoin-proofs/src/api/seal.rs
```diff
@@ -758,3 +759,64 @@ pub fn verify_batch_seal<Tree: 'static + MerkleTreeTrait>(
...
+>(
+    mut rng: &mut Rng,
+    porep_config: PoRepConfig,
+    cache_path: R,
+    out_path: S,
+) -> Result<Commitment> {
+    let sector_bytes = PaddedBytesAmount::from(porep_config).0 as usize;
+
+    {
+        // Create a sector full of null bytes at `out_path`.
```
cryptonemo, line 786: I think we can use this instead of allocating/looping: https://doc.rust-lang.org/std/fs/struct.File.html#method.set_len

storage-proofs/porep/src/stacked/vanilla/proof.rs
```diff
@@ -1161,6 +1161,167 @@ impl<'a, Tree: 'static + MerkleTreeTrait, G: 'static + Hasher> StackedDrg<'a, Tr
 
         Ok((tau, (paux, taux)))
     }
+
+    // Assumes data is all zeros.
+    // Replica path is used to create configs, but is not read.
+    // Instead new zeros are provided (hence the need for replica to be all zeros).
+    fn generate_fake_tree_r_last<TreeArity>(
```
cryptonemo, line 1168: Oops, this comment didn't make it: Is it possible to use the 'real' tree_r_last code with a zeroed replica file and have it work as normal?  At a quick glance, this code looks largely duplicated.

cryptonemo: I see the encoding is dropped in this version.  If it's just that and not beneficial to use a null encoder (or whatever is needed), you can disregard this comment.

filecoin-proofs/tests/api.rs
```diff
@@ -81,25 +81,29 @@ fn seal_lifecycle<Tree: 'static + MerkleTreeTrait>(sector_size: u64) -> Result<(
 #[test]
 #[ignore]
 fn test_winning_post_2kib_base_8() -> Result<()> {
-    winning_post::<SectorShape2KiB>(SECTOR_SIZE_2_KIB)
+    winning_post::<SectorShape2KiB>(SECTOR_SIZE_2_KIB, false);
+    winning_post::<SectorShape2KiB>(SECTOR_SIZE_2_KIB, true)
```
cryptonemo, line 85: Seeing warnings about the `Result` here not being consumed (from each of the first calls)

cryptonemo:
Hrm, I tried adding in just a fauxrep test for 512MiB and I'm seeing failures.  Any reason you can think of that might happen? (`proof did not verify`).  I created new params for all of the test sector sizes used and 2/4/16/32KiB were all good.

storage-proofs/porep/src/stacked/vanilla/proof.rs
porcuquine: I started out trying to reuse. Encoding became a pain. And there's no need to indirect through a file since we know the content. Although it might be possible to refactor to share some code, I decided it was fine to duplicate the minimum needed for this specialized path. This also minimizes risk/change for the real and tested code path. I generally agree with your sentiments, though.

porcuquine:
> Hrm, I tried adding in just a fauxrep test for 512MiB and I'm seeing failures. Any reason you can think of that might happen? (proof did not verify). I created new params for all of the test sector sizes used and 2/4/16/32KiB were all good.

Not sure, but I will try it and debug. Should do the same for the real sector sizes, too to verify before finalizing this.

filecoin-proofs/src/api/seal.rs
porcuquine: Good call. I pushed that fix and also one to reuse the vector when building on GPU.

porcuquine:
@cryptonemo I could not reproduce the problem with 512MiB sectors. I verified it working for me for both Window and Winning PoSt — on CPU and GPU. Given that you are seeing a failure of the proof itself, I suspect something is wrong with your params.

cryptonemo:
> @cryptonemo I could not reproduce the problem with 512MiB sectors. I verified it working for me for both Window and Winning PoSt — on CPU and GPU. Given that you are seeing a failure of the proof itself, I suspect something is wrong with your params.

32GiB sector test worked fine.  I agree about the params.

PR #1198, opened 2020-07-02
opt: Stacked DRG: swap buffers instead of memcpy #1197

### LINKED ISSUE #1197: Security Audit FPS-19
### Description

The function generate_labels() iterates though the layers of the Stacked DRG, creating labels for each node based off their parents in both this layer and the previous one.

The function stores an array of the labels consisting of the current layer and the previous layer as
labels_buffer of length n ∗ 2 where n is the size of a layer. The first n nodes (i.e. [0, n)) represent the current
layer and the second n (i.e. [n, 2n)) nodes represent the previous layer.

At the end of the iteration of each layer there is a memcopy of the current nodes to the previous nodes (i.e. the
first n nodes are copied over the second n nodes). Subsequently, the next iteration will overwrite the current
nodes (i.e. the first n nodes) with the new current layer nodes.

These new current layer nodes overwrite the first n nodes without reading them. Thus, the memcopy can be
avoided.

It can be avoided by, at the end of the first iteration, switching the previous layer nodes to be the first n nodes
and the current layer nodes to be the second n nodes. After the second iteration, switch back such that the first
n are the current layer nodes and the second n nodes are the previous layer nodes. This process can be repeated
to reduce the need for a memcopy .

This is a significant optimisation as a layer will contain a sector size worth of nodes which may be as large as
64GiB.

### Where to begin

1) Determine if the problem description as reported is accurate.

2) Consider implementing the optimisation described above to reduce copying large quantities of memory.

keyvank:
Resolves #1197

### DIFF
--- a/storage-proofs/porep/src/stacked/vanilla/proof.rs
+++ b/storage-proofs/porep/src/stacked/vanilla/proof.rs
@@ -299,8 +299,8 @@ impl<'a, Tree: 'static + MerkleTreeTrait, G: 'static + Hasher> StackedDrg<'a, Tr
 
         let layer_size = graph.size() * NODE_SIZE;
         // NOTE: this means we currently keep 2x sector size around, to improve speed.
-        let mut layer_labels = vec![0u8; layer_size];
-        let mut exp_labels = vec![0u8; layer_size];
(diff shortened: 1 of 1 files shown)


keyvank:
@porcuquine LGTM

porcuquine:
Needs rebase.

PR #1202, opened 2020-07-06
Make pieces::Stack methods private.

porcuquine:
The `pieces::Stack` struct is only used in the implementation of `compute_comm_d`, and accurate usage of its helper methods is required for correctness. For example, `peek()` and `peek2` must not be called when the stack is not large enough. This correctness is guaranteed by usage of `Stack` as a helper data structure within `compute_comm_d` as written, but there is no reason to expose these methods to callers who *might* misuse them. Put differently, correctness of the `Stack` data structure depends on these methods being private. This allows a reader to determine — by only inspecting  the `pieces` module — that these methods are never called incorrectly.

### DIFF
--- a/filecoin-proofs/src/pieces.rs
+++ b/filecoin-proofs/src/pieces.rs
@@ -142,31 +142,31 @@ struct Stack(Vec<PieceInfo>);
 
 impl Stack {
     /// Creates a new stack.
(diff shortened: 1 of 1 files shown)


porcuquine:
This is a further response to Security Audit FPS-11.

PR #1204, opened 2020-07-07
feat: allow for compilation on aarch64

dignifiedquire:
Depends on 

- [x] https://github.com/filecoin-project/sapling-crypto/pull/18
- [x] https://github.com/filecoin-project/ff/pull/7

### DIFF
--- a/fil-proofs-tooling/Cargo.toml
+++ b/fil-proofs-tooling/Cargo.toml
@@ -25,7 +25,7 @@ chrono = { version = "0.4.7", features = ["serde"] }
 memmap = "0.7.0"
 bellperson = "0.9.1"
 paired = "0.20.0"
-fil-sapling-crypto = "0.6.0"
+fil-sapling-crypto = "0.6.3"
 rand = "0.7"
 storage-proofs = { path = "../storage-proofs"}
 filecoin-proofs = { path = "../filecoin-proofs"}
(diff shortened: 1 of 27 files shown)


fil-proofs-tooling/src/metadata.rs
```diff
@@ -109,6 +110,33 @@ impl SystemMetadata {
             memory_total_bytes: memory.total().get::<uom::si::information::byte>(),
         })
     }
+
+    #[cfg(not(target_arch = "x86_64"))]
```
vmx, line 114: I personally prefer keeping the `cfg()` differences as small as possible. We could create a methos that returns the host, memory and the cpu_logical/physical that is shared between the `new()`s. This would also make it more apparent what the difference between the cfgs is. Though it''s a matter of test, hence just a comment and nothing that needs to change/blocking a merge.

dignifiedquire: I agree this could be nicer, but I mostly wanted to get this to compile

porcuquine: @keyvank

fil-proofs-tooling/Cargo.toml
```diff
@@ -25,7 +25,7 @@ chrono = { version = "0.4.7", features = ["serde"] }
 memmap = "0.7.0"
 bellperson = "0.9.1"
 paired = "0.20.0"
-fil-sapling-crypto = "0.6.0"
+fil-sapling-crypto = "0.6.3"
```
cryptonemo, line 28: Rust oddity, these individual patch version updates can be omitted.  They don't hurt though ;-)

PR #1207, opened 2020-07-10
Default cache paths to use FIL_PROOFS_CACHE_DIR if set.

### LINKED ISSUE #1206: Respect TMPDIR environment variable
### Description
The two lines [here](https://github.com/filecoin-project/rust-fil-proofs/blob/master/storage-proofs/core/src/settings.rs#L44) hardcode the location of the parameter cache and the parent cache into the `/tmp` directory. For many reasons, the root partition of most linux installations is not usually given a substantial amount of storage. This also prevents the user from making a more informed choice and placing the caches on faster drives that may not be the drive backing the root partition.

### Acceptance criteria
Code respects the TMPDIR environment variable allowing the user to determine where the two caches are stored. 

### Risks + pitfalls
N/A

### Where to begin
https://github.com/filecoin-project/rust-fil-proofs/blob/master/storage-proofs/core/src/settings.rs#L44

porcuquine:
Closes #1206.

~NOTE: Do not merge until a decision is made. This PR was a proposal for @lanzafame's approval, but merging it may cause as much or more trouble as the originally reported problem.~

UPDATE: We have settled on solution: separate actual cache usage from canonical parameter location.

*In particular, we must ensure the location of the Groth parameter cache does not change for any user.* (cc: @travisperson)

This is untested, but I wanted to push something quickly for @lanzafame to review, since this is high priority for him.

@lanzafame Please approve if this seems to satisfy your requirements. If so, ~we'll~ we might get this into the upcoming release after making sure it's solid.

TODO:
- [ ] document this change in README.
- [x] document usage/policy for future cache settings in source.

### DIFF
--- a/storage-proofs/core/src/settings.rs
+++ b/storage-proofs/core/src/settings.rs
@@ -1,3 +1,4 @@
+use std::env;
 use std::sync::Mutex;
 
 use config::{Config, ConfigError, Environment, File};
@@ -10,6 +11,7 @@ lazy_static! {
 }
 
 const SETTINGS_PATH: &str = "./rust-fil-proofs.config.toml";
+const PREFIX: &str = "FIL_PROOFS";
 
 #[derive(Debug, Serialize, Deserialize)]
 #[serde(default)]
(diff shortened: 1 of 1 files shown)


travisperson:
Just to clarify (and in preparation for change log) I believe this is how it works now:

If `FIL_PROOFS_PARENT_CACHE` is **set** cached parents trees will be stored under `$FIL_PROOFS_PARENT_CACHE`. Setting `FIL_PROOFS_PARENT_CACHE` has no affect.
If  `FIL_PROOFS_PARENT_CACHE` is **unset** and `FIL_PROOFS_CACHE_DIR` is **set**, cached parents trees will be stored under `$FIL_PROOFS_CACHE_DIR/filecoin-parents`.
If If `FIL_PROOFS_CACHE_DIR` is **unset** and `FIL_PROOFS_PARENT_CACHE` is **unset**, the directory will be `/var/tmp/filecoin-parents`

The behaviour of the parameter cache is unchanged. Setting `FIL_PROOFS_CACHE_DIR` has no affect on their location.

porcuquine:
@travisperson I think this looks right except for one thing:

>If FIL_PROOFS_PARENT_CACHE is set cached parents trees will be stored under $FIL_PROOFS_PARENT_CACHE. Setting FIL_PROOFS_PARENT_CACHE has no affect.

Did you mean setting *`FIL_PROOFS_CACHE_DIR`* has no effect?

travisperson:
Yes

cryptonemo:
Approved, although no longer related to TMPDIR.  Will that part cause confusion?

porcuquine:
@cryptonemo Fixed reference to TMPDIR. If you have bandwidth to propagate changes everywhere after #1211 merges (README, config sample, etc.) that would be helpful. If so, I consider this can merge and be incorporated in the next release when ready. (I will update the description to reflect this decision.)

porcuquine:
@cryptonemo I pushed one more commit documenting usage in a comment. Please review before propagating, since it clarifies a wart: specifically, the new env var cannot be set by config file — just env var.

PR #1217, opened 2020-07-15
Refactor batch_hash for clarity.

porcuquine:
@DrPeterVanNostrand Does this help?

### DIFF
--- a/storage-proofs/porep/src/nse/vanilla/batch_hasher.rs
+++ b/storage-proofs/porep/src/nse/vanilla/batch_hasher.rs
@@ -21,31 +21,29 @@ pub fn batch_hash(
     data: &[u8],
 ) -> [u8; 32] {
     assert!(parents.len() % 2 == 0, "number of parents must be even");
+    assert!(degree % 2 == 0, "degree must be even");
     assert_eq!(parents.len(), degree * k, "invalid number of parents");
     let modulus = Fr::char();
+    let k = k as u32;
+    let zero = FrRepr::from(0);
 
-    for (i, j) in (0..degree).tuples() {
-        let k = k as u32;
+    // Read the (i, j)th parent.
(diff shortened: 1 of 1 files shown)


porcuquine:
@dignifiedquire

We don't necessarily need to merge, but @DrPeterVanNostrand needs to spec, so let's at least keep the PR open for his reference.

That said, isn't it the case that the CPU performance of NSE is poor enough that we currently assume only GPU is viable? If that's the case, even if only for now, I think we would be best off targeting clarity first. That will make spec and audits much easier.

If we come to believe CPU can actually perform acceptably, we can always optimize later.

The above points I think are worth considering even in the light of performance — so we can best get through this initial implementation period, especially with two implementations, spec, and audits to juggle.

Finally — and here's where it gets interesting — I was curious about your assertion. I assumed the compiler would be smart enough to inline these very short closures and didn't see any other reason why the refactor *must* be slower. I do take your word for it, especially since you probably spent time optimizing and generally know how to tease out good perf.

Nevertheless, I wanted to take a crack at understanding why this code would be slower, and maybe see if I could improve it, so I ran the benchmarks. Maybe I misread the results, but it looks like the refactor is actually faster (12%). Assuming that's accurate, I consider this a big win for Rust. It means common-sense hopes for a 'sufficiently smart compile' do sometimes allow prioritizing clarity without penalty. It's also a reminder that Rust's closure's are pretty sweet.

Original:
```
(base) ➜  porep git:(feat/nse) cargo bench --bench batch_hasher
batch-hash              time:   [156.69 us 157.30 us 157.86 us]
```

This PR:
```
(base) ➜  porep git:(refactor/batch-hash) cargo bench --bench batch_hasher
batch-hash              time:   [139.11 us 139.50 us 139.86 us]
                        change: [-12.846% -11.994% -11.130%] (p = 0.00 < 0.05)
                        Performance has improved.

```

Issue #1219, opened 2020-07-17
Fix over-built partial trees for cached proof generation


cryptonemo:
### Description

Reported on slack by user "chenjianye", there is a bug related to past PRs still alive in the code today.  The relevant PRs are:

https://github.com/filecoin-project/merkle_light/pull/88
and
https://github.com/filecoin-project/rust-fil-proofs/pull/1165

PR https://github.com/filecoin-project/merkle_light/pull/88 was not merged, but would have been a possible solution to this issue, and while https://github.com/filecoin-project/rust-fil-proofs/pull/1165 was merged, it missed the buggy instance in question:

This "missed" instance (i.e. a failure in grep by the author, @cryptonemo ) allows `post` to over build partial trees during cached proof generation by only using a subset of the cached data, rather than all of it.  The current code believes that the cached data had discarded 3 rows, rather than 2 rows, meaning the logic rebuilds a single row extra than it needs to due to the 'default' discrepancy.  This is because the default value is chosen not by the re-worked default logic in `rust-fil-proofs`, but by the now outdated default logic in `merkle_light`.

While there are many many checks that will not allow a level cache store to be instantiated with an incorrect value for `rows_to_discard`, in this case, the tree is already instantiated and we're later passing an incorrect value specifying what it needs to use.  Technically this is correct behaviour and it's working as intended, however, in this case, it's not at all desired or intended.

### Acceptance criteria

The preference is to no longer overbuild partial trees from cached tree data during proof generation, nor rely on defaults from merkle_light.

### Risks + pitfalls

There are some complications regarding tests, written up below.

### Where to begin

SImply changing the default in `merkle_light` as PR https://github.com/filecoin-project/merkle_light/pull/88 will solve this particular issue and tests well locally.  However, a different proposal is to use the re-worked local default for consistency in the proofs code, to bypass that default all together.  While this also solves the issue, it unfortunately causes some tests to fail, where proofs are attempted to be built with a setting that is too large for test sized tree.   In particular, `merkle_light` has checks that return failures when the value of `rows_to_discard` is smaller than (or equal to) the number of rows in the tree.  Note that this only happens during testing, where trees are small and this issue does not affect production tree widths.

The combination of correcting the default usage in `rust-fil-proofs` with the additional row checks in `merkle_light` is a proposed solution to this.  Previously, this wasn't needed because the tests passed, but apparently relied on `merkle_light`'s default value, which did not allow it to be too small for the tree.

[...]

PR #1220, opened 2020-07-17
fix: correct rows_to_discard value during post

cryptonemo:


cryptonemo:
Requires https://github.com/filecoin-project/merkle_light/pull/89

cryptonemo:
Related to https://github.com/filecoin-project/rust-fil-proofs/issues/1219

cryptonemo:
@porcuquine Can you take a look at https://github.com/filecoin-project/merkle_light/pull/89 as well?  I think the proper way to proceed is to merge that one first, cut a new release for merkle_light, then update this PR to point to that new version.

Separately, we can decide if we need to cut an unplanned proofs release today.  I'm inclined to say no since I have a handle on the issue and it's inefficient, but not breaking as-is, and I'm not entirely sure users of `filecoin-ffi` are ready to consume.

porcuquine:
Yep, I'm looking and have on question (will ask there).

I think it's fine to defer releasing this until the next natural release.

cryptonemo:
Resolves #1219

PR #1226, opened 2020-07-24
Improve feature manage

LuoZijun:
It is easier to remove `neptune-triton` dependencies when compiling machines without GPU.


storage-proofs/Cargo.toml
```diff
@@ -14,10 +14,14 @@ storage-proofs-post = { path = "./post", version = "4.0.1" }
 storage-proofs-porep = { path = "./porep", version = "4.0.1" }
 
 [features]
-default = ["gpu"]
+default = []
```
vmx, line 17: I guess making it possible to disable the gpu feature makes sense. Though I think it should still be on by default.

LuoZijun: enable default gpu feature crates: `fil-proofs-tooling`、[filecoin-ffi](https://github.com/filecoin-project/filecoin-ffi/pull/125) 、[rust-filecoin-proofs-api](https://github.com/filecoin-project/rust-filecoin-proofs-api/pull/36)

other crates (`filecoin-proofs`、`storage-proofs`、`storage-proofs-core`、`storage-proofs-porep`、`storage-proofs-post`) are disabled default gpu feature.

is this ok ?

LuoZijun: @vmx BTW,  the file ` storage-proofs/porep/src/stacked/vanilla/proof.rs` has updated, now the code of SDR Precommit2 code can work in both cases (enable GPU or disable).

LuoZijun:
GPU features enabled by default:

https://github.com/filecoin-project/filecoin-ffi/pull/125
https://github.com/filecoin-project/rust-filecoin-proofs-api/pull/36

storage-proofs/Cargo.toml
vmx: @LuoZijun Thanks for those changes. I think that `gpu` should be enabled by default. In case that breaks your use case, we should talk about that in more detail.

Before you make any more changes, I'd like to my team first, if we want to have non-GPU support in rust-fil-proofs.

vmx:
@LuoZijun I'm sorry for taking so long coming back to you.

Would it work for you if all crates of this repo have `gpu` enabled as default feature. Then `filecoin-ffi` includes `filecoin-proofs-api` with `default-features = false`, but has a `gpu` feature that is enabled by default.

dignifiedquire:
We now have a `gpu` feature in all the crates, except `filecoin-ffi` which is enabled by default but can be disabled with `--no-default-features`

dignifiedquire:
There is https://github.com/filecoin-project/rust-fil-proofs/issues/1346 to track that OpenCL is not required when that feature is disabled.

PR #1234, opened 2020-07-29
Use (varied) constant-length vector to hash column roots.

porcuquine:
This PR replaces a stopgap @dignifiedquire had implemented as a replacement for the previous unpadded MD construction. It depends on changes in neptune to add domain separation tags and support for a number of distinct hashing functions built over the Poseidon permutation: https://github.com/filecoin-project/neptune/pull/43

The new hash function allows for preimages which are shorter than the 'natural' size of the permutation — but whose actual size must be known such that the actual length is encoded in the domain tag.

We need (or would like to have) such a hash function so we can use fewer layers in test sectors/circuits but still use generic arrays with statically-typed sizes as required by the neptune API. This usage fits ideally with the new 'constant length' hash type.

To facilitate cheap creation of constants which vary only by their specified length, we use the `with_length` function provided for that purpose.

@dignifiedquire I think this is ready for review and will just need to update deps once the neptune PR works its way through.

cc: @DrPeterVanNostrand, since this change will need to land in the spec also.

cc: @nicola for interest re: domain separation tags.

### DIFF
--- a/storage-proofs/core/Cargo.toml
+++ b/storage-proofs/core/Cargo.toml
@@ -40,7 +40,7 @@ hex = "0.4.0"
 generic-array = "0.13.2"
 anyhow = "1.0.23"
 thiserror = "1.0.6"
-neptune = { version = "1.0.1", features = ["gpu"] }
+neptune = { version = "2.0.0", features = ["gpu"] }
 cpu-time = { version = "1.0", optional = true }
 gperftools = { version = "0.2", optional = true }
 num_cpus = "1.10.1"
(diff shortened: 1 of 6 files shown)


porcuquine:
@dignifiedquire @cryptonemo I have updated the dependencies to refer to neptune 2.0.0, which completes the current work.

This will need to be rebased — but that should probably wait until the target branch (`feat/nse`) is fixed. Meanwhile, my expectation is that CI here should pass (except for the two jobs which seem to always time out with NSE).

We need to get that fixed soon. As increasing effort gets put into NSE, chronically broken CI will become more problematic. @keyvank will try to fix it — but this is a general note that it needs to happen.

cryptonemo:
CI ok?

porcuquine:
It's the same as the base NSE branch, which I believe passes locally but is still being tweaked for CI (cc: @keyvank )

PR #1235, opened 2020-07-30
feat: add tool to split phase2 parameters

vmx:
The `phase2` CLI has a new command called `split-keys` which allows to
split phase2 parameter output into parameters that can be used for
rust-fil-proofs.

I'm not done yet, the path handling is not robust enough. Though it would
be good to get early review.

### DIFF
--- a/filecoin-proofs/src/bin/phase2.rs
+++ b/filecoin-proofs/src/bin/phase2.rs
@@ -31,8 +31,12 @@ use simplelog::{self, CombinedLogger, LevelFilter, TermLogger, TerminalMode, Wri
 use storage_proofs::compound_proof::{self, CompoundProof};
 use storage_proofs::hasher::Sha256Hasher;
 use storage_proofs::merkle::MerkleTreeTrait;
-use storage_proofs::porep::stacked::{StackedCircuit, StackedCompound, StackedDrg};
-use storage_proofs::post::fallback::{FallbackPoSt, FallbackPoStCircuit, FallbackPoStCompound};
+use storage_proofs::porep::stacked::{
+    PublicParams as PoRepPublicParams, StackedCompound, StackedDrg,
+};
+use storage_proofs::post::fallback::{
+    FallbackPoSt, FallbackPoStCircuit, FallbackPoStCompound, PublicParams as PoStPublicParams,
+};
 
 const CHUNK_SIZE: usize = 10_000;
 
@@ -281,9 +285,7 @@ fn parse_params_filename(path: &str) -> (Proof, Hasher, Sector, String, usize, P
     )
 }
 
-fn blank_sdr_poseidon_circuit<Tree: MerkleTreeTrait>(
-    sector_size: u64,
-) -> StackedCircuit<'static, Tree, Sha256Hasher> {
+fn blank_sdr_poseidon_params<Tree: MerkleTreeTrait>(sector_size: u64) -> PoRepPublicParams<Tree> {
     let n_partitions = *POREP_PARTITIONS.read().unwrap().get(&sector_size).unwrap();
 
     let porep_config = PoRepConfig {
@@ -308,11 +310,7 @@ fn blank_sdr_poseidon_circuit<Tree: MerkleTreeTrait>(
         _,
     >>::setup(&setup_params)
     .unwrap();
-
-    <StackedCompound<Tree, Sha256Hasher> as CompoundProof<
-        StackedDrg<Tree, Sha256Hasher>,
-        _,
-    >>::blank_circuit(&public_params.vanilla_params)
+    public_params.vanilla_params
 }
 
 /*
@@ -350,9 +348,9 @@ fn blank_porep_sha_pedersen_circuit(
 }
 */
 
-fn blank_winning_post_poseidon_circuit<Tree: 'static + MerkleTreeTrait>(
+fn blank_winning_post_poseidon_params<Tree: 'static + MerkleTreeTrait>(
     sector_size: u64,
-) -> FallbackPoStCircuit<Tree> {
+) -> PoStPublicParams {
     let post_config = PoStConfig {
         sector_size: SectorSize(sector_size),
         challenge_count: WINNING_POST_CHALLENGE_COUNT,
@@ -361,17 +359,12 @@ fn blank_winning_post_poseidon_circuit<Tree: 'static + MerkleTreeTrait>(
         priority: false,
     };
 
-    let public_params = winning_post_public_params::<Tree>(&post_config).unwrap();
-
-    <FallbackPoStCompound<Tree> as CompoundProof<
-        FallbackPoSt<Tree>,
-        FallbackPoStCircuit<Tree>,
-    >>::blank_circuit(&public_params)
+    winning_post_public_params::<Tree>(&post_config).unwrap()
 }
 
-fn blank_window_post_poseidon_circuit<Tree: 'static + MerkleTreeTrait>(
+fn blank_window_post_poseidon_params<Tree: 'static + MerkleTreeTrait>(
     sector_size: u64,
-) -> FallbackPoStCircuit<Tree> {
+) -> PoStPublicParams {
     let post_config = PoStConfig {
         sector_size: SectorSize(sector_size),
         challenge_count: WINDOW_POST_CHALLENGE_COUNT,
@@ -384,12 +377,7 @@ fn blank_window_post_poseidon_circuit<Tree: 'static + MerkleTreeTrait>(
         priority: false,
     };
 
-    let public_params = window_post_public_params::<Tree>(&post_config).unwrap();
-
-    <FallbackPoStCompound<Tree> as CompoundProof<
-        FallbackPoSt<Tree>,
-        FallbackPoStCircuit<Tree>,
-    >>::blank_circuit(&public_params)
+    window_post_public_params::<Tree>(&post_config).unwrap()
 }
 
 /// Creates the first phase2 parameters for a circuit and writes them to a file.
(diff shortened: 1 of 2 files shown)


vmx:
It already works like the other commands. It creates the .params and .vk file in the directory the program was run from. The log file is created in the directory of the source file.

I'm now looking for ideas on how to verify that it does the right thing.

filecoin-proofs/src/bin/phase2.rs
```diff
@@ -1018,6 +1022,67 @@ fn setup_logger(log_filename: &str) {
...
 
+fn parameter_identifier<Tree: 'static + MerkleTreeTrait>(sector_size: u64, proof: Proof) -> String {
+    match proof {
+        Proof::Sdr => {
+            let n_partitions = *POREP_PARTITIONS.read().unwrap().get(&sector_size).unwrap();
+
+            let porep_config = PoRepConfig {
+                sector_size: SectorSize(sector_size),
+                partitions: PoRepProofPartitions(n_partitions),
+                porep_id: [0; 32],
```
DrPeterVanNostrand, line 1033: I'm not certain that setting the porep-id to all zeroes will produce the correct identifier for mainnet params which have a nonzero porep-id. It may, but idk

filecoin-proofs/src/bin/phase2.rs
```diff
@@ -1423,6 +1496,61 @@ fn main() {
...
+                    let vk_path = verifying_key_id(&identifier);
+                    info!("writing verifying key to file: {}", vk_path);
+                    let mut vk_file = File::create(&vk_path).unwrap();
+                    vk_data.write(&mut vk_file).unwrap();
+                }
+
+                info!(
+                    "size of the verifying key is {} bytes",
+                    input_file.seek(SeekFrom::Current(0)).unwrap()
+                );
```
DrPeterVanNostrand, line 1527: I think that you could combine these:

```rust
{
    ...
    let vk_file_len = vk_data.write(&mut vk_file).unwrap();
    info!("size of the verifying key is {} bytes", vk_file_size);
}
```

filecoin-proofs/src/bin/phase2.rs
vmx: I got them from https://github.com/filecoin-project/rust-fil-proofs/blob/5e760a737ffb74267c119bc662b865b884f1aac3/filecoin-proofs/src/bin/phase2.rs#L277. I used the parameter of the initial files.

DrPeterVanNostrand: The phase2 initial parameter gen won't be affected by the porep-id, that's why it's set to all zeros. But, I didn't know whether or not the porep-id will affect the identifier

filecoin-proofs/src/bin/phase2.rs
```diff
@@ -1423,6 +1496,61 @@ fn main() {
...
+                    io::copy(&mut input_file, &mut params_file).unwrap();
+
+                    // The file still contains information about the contributions and the cs_hash,
+                    // remoe those things with truncating the file
+                    let current_pos = params_file.seek(SeekFrom::Current(0)).unwrap();
+                    // file_size - cs_hash - contributions_length -
+                    //   (num_contributions * public_key_size)
+                    let params_file_size = current_pos - 64 - 4 - (param_num as u64 * 544);
+                    info!("size of the params file is {} bytes", params_file_size);
+                    params_file.set_len(params_file_size).unwrap();
```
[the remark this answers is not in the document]
vmx: You seem to have done more file io stuff in the past. That's sooooo much better. Thanks!

filecoin-proofs/src/bin/phase2.rs
vmx: https://github.com/filecoin-project/rust-fil-proofs/blob/5e760a737ffb74267c119bc662b865b884f1aac3/storage-proofs/porep/src/stacked/vanilla/params.rs#L84-L91 is called on the public parameters.

Example output: `layered_drgporep::PublicParams{ graph: stacked_graph::StackedGraph{expansion_degree: 8 base_graph: drgraph::BucketGraph{size: 1073741824; degree: 6; hasher: poseidon_hasher} }, challenges: LayerChallenges { layers: 11, max_count: 18 }, tree: merkletree-poseidon_hasher-8-8-0 }`

So it doesn't look like the `porep_id` is part of it.

porcuquine: It's possible that there is a bug (no reason to expect it, and a scan of code suggests not), but for reference:

- The identifier used to derive the working file for the parameters should *not* depend on `porep_id`.
- There should be a 1-1 relationship between distinct groth params and file names.
- `porep_id` is intentionally decoupled from the concrete circuit (by design, it can be used to change the graphs without changing the circuit).
- This explains both why phase2 does not require `porep_id` and also why it must not affect the generated filename.
- If the generated filename *does* depend on `porep_id`, that is a bug — and we should fix it.
- Here is the code which I *think* shows that it does not (even though it 'could'): https://github.com/filecoin-project/rust-fil-proofs/blob/master/storage-proofs/porep/src/stacked/vanilla/params.rs#L84-L91

So I think it should be safe to assume an arbitrary `porep_id` can be supplied. However, it would be worth verifying that choice of `porep_id` doesn't affect the final filenames — as an independent check on this important logic.

vmx:
Even if this would work, I want to make one more change, which is extracting the contributions into yet another file (that's purely for the verification process).

porcuquine:
One quick note, bearing in mind that I haven't had a chance to review this overall yet:

Ideally, this tool would be general-purpose in the sense that it could be used for any circuits created by phase2. That can and will change in the future when we perform more trusted setups.

I *think* (based on cursory discussion) it's the case that you currently extract the circuit parameters by parsing the filenames. In a perfect world, you would get all the circuit specification in the same way the initial phase2 parameters were created. Then, as long as the commands used to create initial parameters for a circuit were saved, they could be used verbatim to also perform this final splitting. I think, if designed correctly, that will allow for the final phase to happen correctly without need of any manual intervention or decision-making, and even with new, not-yet-supported circuits. In other words, we can/should reuse the circuit-selection logic used for creating initial phase2 params. This might require more refactoring than is convenient now, though — in which case it could be added as a follow-up but prior to the next trusted setup.

vmx:
> I _think_ (based on cursory discussion) it's the case that you currently extract the circuit parameters by parsing the filenames.

No I don't, I copy and pasted the parameters from earlier in the file.

> In a perfect world, you would get all the circuit specification in the same way the initial phase2 parameters were created.

I actually thought about refactoring the code to do that. Though I hesitated as I don't know how much we want to refactor the original trusted setup code. It's a matter of e.g. taking https://github.com/filecoin-project/rust-fil-proofs/blob/5e760a737ffb74267c119bc662b865b884f1aac3/filecoin-proofs/src/bin/phase2.rs#L338-L355

and splitting it into two functions, one that generates the parameters (which will then be re-used by my code) and the `    <FallbackPoStCompound<Tree> as CompoundProof<
        FallbackPoSt<Tree>,
        FallbackPoStCircuit<Tree>,
    >>::blank_circuit(&public_params)` call.

vmx:
If I should do the refactoring mentioned above, let me know.

porcuquine:
@vmx I think what you're describing is the same thing I have believed is the 'right thing'. If you think this can be done practically in such a way as to be usable for the current trusted setup, then I think it's probably worth doing now. Unless @DrPeterVanNostrand sees a problem with the original code being refactored in what will be used in the last step of this trusted setup, I would proceed. If we decide we don't want to touch it now, we should still do it in a separate step for next time. This all makes me think it's probably better to do it now, fwiw.

filecoin-proofs/src/bin/phase2.rs
DrPeterVanNostrand: ha, I should have looked at the `write` function first

filecoin-proofs/src/bin/phase2.rs
```diff
@@ -1423,6 +1496,74 @@ fn main() {
...
+                        &mut params_file,
+                    )
+                    .unwrap();
+                }
+
+                {
+                    // Writing the contributions to disk is not needed for the final parameters,
+                    // they won't be published, they are only there for verification purpose.
+                    let contribs_path =
+                        format!("v{}-{}.contribs", parameter_cache::VERSION, &identifier);
```
porcuquine, line 1550: There is a subtlety here. If we use the current value of `VERSION` when running the tool, then we will end up with the version corresponding to the current (untrusted) parameters.

That would imply a sequence like the following:

- Bump `VERSION`.
- Run split tool to create trusted parameters.
- Publish parameters.
- Release.

Note that in this sequence, the version must be bumped before the tool is run. We *could* bump the version here/now, but then the trusted parameters would need to be included in the very next release. For that reason, something like the sequence above is probably better. It's also possible that I'm seeing something wrong, so please let me know if so or explicitly confirm below otherwise (cc: @cryptonemo @vmx ).

vmx: I would use the sequence then @cryptonemo would use when he publishes new parameters. I'd expect it to be what @porcuquine outlined above, but @cryptonemo would know best.

vmx:
@DrPeterVanNostrand I've made the refactoring that @porcuquine proposed. If you look at the commit in isolation https://github.com/filecoin-project/rust-fil-proofs/pull/1235/commits/a6b2f943eeeb8e70a99a82798e2a5146fdab4029 it shows quite nicely that is is just a basic refactoring for the existing code, but makes the new code more robust as there's less duplicated code.

filecoin-proofs/src/bin/phase2.rs
porcuquine: @cryptonemo Can you confirm that we're not missing anything here and that my sequence above is workable (or propose an alternative)?

cryptonemo: Confirmed

vmx:
@DrPeterVanNostrand could you please approve or request changes? I especially would like to know if you are OK with the refactoring that also changes other parts of the code. I'd then make any needed changes (and the rebase that's needed anyway) my tomorrow and then we can hopefully merge it.

vmx:
@DrPeterVanNostrand just confirmed through a side-channel that the refactoring is OK. So I'll go with that.

PR #1237, opened 2020-08-03
fix: update tree usage when using the gpu builder

cryptonemo:
fix: update merkletree inconsistent usage as lc trees


keyvank:
LGTM

PR #1238, opened 2020-08-04
feat: add a check parameter command that maps parameter files

cryptonemo:


cryptonemo:
Related to #1235

cryptonemo:
> I'm not sure why you use a subcommand if there is only a single one, but perhaps you've bigger plans.

I figured it could someday be used for more than mapping, maybe verifying, or dumping info about, etc.  Not urgent right now, but might be nice to have a tool that can do some more checking on parameters as needed.

cryptonemo:
I'll merge this after #1235 btw.  @porcuquine should still review/approve.

vmx:
I'd merge this right away once @porcuquine reviewed it. It's independent of #1235. I'd merge #1235 after #1242/#1244 so that @DrPeterVanNostrand doesn't have to rebase.

PR #1239, opened 2020-08-04
feat: verify non-raw to raw param transitions in phase2-cli

DrPeterVanNostrand:
Allows phase2 to verify non-raw to raw contributions.

### DIFF
--- a/filecoin-proofs/src/bin/phase2.rs
+++ b/filecoin-proofs/src/bin/phase2.rs
@@ -17,7 +17,7 @@ use filecoin_proofs::types::{
     PaddedBytesAmount, PoRepConfig, PoRepProofPartitions, PoStConfig, PoStType, SectorSize,
 };
 use filecoin_proofs::with_shape;
-use log::{error, info};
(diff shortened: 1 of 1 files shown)


filecoin-proofs/src/bin/phase2.rs
```diff
@@ -1278,68 +1271,75 @@ fn main() {
...
                     } else {
-                        panic!(
-                            "large 'before' params not found `{}`, fallback failed",
-                            path_before_large
+                        let err_msg = format!(
+                            "no 'before' params found, attempted: {}, {}, {}",
+                            small_raw, small_nonraw, large
                         );
+                        error!("{}", err_msg);
+                        panic!("{}", err_msg);
```
porcuquine, line 1310: Why both?

DrPeterVanNostrand: Participant's may not be familiar with tmux commands, if an error occurs it's better to have the error in a sendable file rather than asking them to mess with tmux scolling.

PR #1240, opened 2020-08-05
fix: phase2-cli paramgen and filename parsing

DrPeterVanNostrand:
Phase2-cli incorrectly labeled the small initial params as raw.

### DIFF
--- a/filecoin-proofs/src/bin/phase2.rs
+++ b/filecoin-proofs/src/bin/phase2.rs
@@ -209,7 +209,7 @@ fn params_filename(
 }
 
 // Parses a phase2 parameters filename into the tuple:
-// (proof, hasher, sector-size, head, param-number, param-size).
+// (proof, hasher, sector-size, head, param-number, param-size, is-raw).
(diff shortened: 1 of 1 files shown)


filecoin-proofs/src/bin/phase2.rs
```diff
@@ -452,13 +455,18 @@ fn create_initial_params<Tree: 'static + MerkleTreeTrait>(
         info!("finished writing large params to file");
     }
 
+    // TODO: add conversion from large to small params to phase2 crate, then write initial params as
+    // small-raw.
+    /*
+    let small_path = params_filename(proof, hasher, sector_size, &head, 0, ParamSize::Small, true);
```
porcuquine, line 461: I assume this is something we should do before the next trusted setup. Can you please create an issue to track this? (Fine to link to this TODO.).

DrPeterVanNostrand: Ok, I made issues in rust-fil-proofs and phase2.

PR #1242, opened 2020-08-05
feat: phase2-cli merge command

DrPeterVanNostrand:
The phase2-cli `merge` command merges large and small param files.

The final phase2 participant generates small-raw params, which are converted into non-raw using `phase2 convert`. The final small-nonraw params are then merged into the large initial params to produce the final `MPCParameters` file for a circuit.

PR #1244, opened 2020-08-05
  feat: phase2-cli merge command (vmx version)

vmx:
This PR is based on https://github.com/filecoin-project/rust-fil-proofs/pull/1242/ and simplifies things a bit. I'm not sure if that's desired or if the parsing is done for additional validation.


vmx:
There is another upcoming command that uses the values from `FileInfo` that I've removed: https://github.com/filecoin-project/rust-fil-proofs/pull/1242#issuecomment-669265027

PR #1249, opened 2020-08-06
feat: add scripts to validate phase2 output and parameter

vmx:
This commit adds two scripts. One is to validate that the trusted
setup phase2 result matches the published parameter files.

The other one is to verify that a `.params` file (and the corresponding
`.vk file`) is part of the `parameters.json`.

### DIFF
--- a/scripts/verify-parameters-json.sh
+++ b/scripts/verify-parameters-json.sh
@@ -1,0 +1,58 @@
+#!/bin/sh
+
+# This script verifies that a given `.params` file (and the corresponding
+# `.vk` file) is part of `parameters.json` as has the correct digest.
+#
+# This script runs on POSIX compatible shells. You need to have standard
+# utilities (`basename`, `head`, `grep`) as well as have `jq` and `b2sum`
+# installed.
+#
+# The input a `parameter.json` file and a `.params' file.
+
+if [ "${#}" -ne 1 ]; then
+    echo "Verify that a given .params file (and the corresponding .vk file)"
+    echo "is part of parameters.json as has the correct digest."
+    echo ""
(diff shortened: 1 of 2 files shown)


cryptonemo:
Looks good, but I'll approve after I have a chance to test

DrPeterVanNostrand:
Maybe these scripts should live in the [`phase2-attestations`](https://github.com/filecoin-project/phase2-attestations) repo? Which is where I putting the rest of the validation commands.

cryptonemo:
Hrm, not sure, but I was thinking they could live here so that I can easily run them.  Not that I can't get the other repo and test it out ...

Between both of you, what's the best way to verify and be happy with the merged files and the split v28 params that are output?

DrPeterVanNostrand:
@nemo If this way is easier, then that's fine with me. I just wanted to ask

I was going to have participant's run `split-keys` on the merged phase2 files and then verify checksums against the v28 split files. I was probably going to re-upload the split files to s3

So something like this, which is in vmx's scripts: 
```
$ ./phase2 split-keys <merged file>
$ split_digest=$(cat <.vk file> <.params file> <.contribs file> | awk '{print $1}'); echo "${split_digest}  <merged file>" | b2sum -c
```

porcuquine:
I approved pending @cryptonemo's test, if it can happen before we need this to test verification instructions.

vmx:
I've split this PR into two pieces.

 - The script to verify the parameters.json file makes sense to have in this repo, hence https://github.com/filecoin-project/rust-fil-proofs/pull/1251
 - The script to verify the trusted setup phase2 results now lives at the attestation repo: https://github.com/filecoin-project/phase2-attestations/pull/3

The scripts are byte-identical to the ones of this PR.

PR #1256, opened 2020-08-12
feat: phase2-cli verify raw g1 points

porcuquine:


porcuquine:
Yes, done — was just checking to make sure what it does agrees with our latest decisions.

PR #1260, opened 2020-08-18
Remove unwrap

cryptonemo:
This is an updated version of #1122, as a lot of code has changed since then

### DIFF
--- a/fil-proofs-tooling/src/bin/benchy/main.rs
+++ b/fil-proofs-tooling/src/bin/benchy/main.rs
@@ -1,4 +1,6 @@
-#![warn(clippy::unwrap_used)]
+//requires nightly, or later stable version
+//#![warn(clippy::unwrap_used)]
+
 use std::io::{stdin, stdout};
 
 use anyhow::Result;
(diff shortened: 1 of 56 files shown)


fil-proofs-tooling/src/bin/benchy/main.rs
```diff
@@ -1,3 +1,6 @@
+//requires nightly, or later stable version
+//#![warn(clippy::unwrap_used)]
```
vmx, line 2: This might be possible now. I haven't checked though.

vmx: That surely isn't a blocker to get this merged.

cryptonemo: I think it works fine on later stable, definitely not on the toolchain we're using though.  If/when we upgrade, we can add those back in.  I think that's a separate discussion from this PR though.

vmx:
I think we should merge this one soon, before it gets outdated again.

cryptonemo:
@vmx Agreed. Also, related to #1261 @porcuquine Please review.

PR #1263, opened 2020-08-20
Feat: pin params to the filecoin collab cluster.

hsanjuan:
This cluster is way faster to ingest data.

The `--local` option makes ingestion even faster (it will put the data
on one peer only while adding and pin everywhere at the end).

Additionally, we trigger pin requests in the main cluster (async) so that
params will be replicated by those nodes too (at some point, once they finish
to fetch them) from the collab cluster.

### DIFF
--- a/scripts/pin-params.sh
+++ b/scripts/pin-params.sh
@@ -53,19 +53,31 @@ else
   exit 1
 fi
 
-CLUSTER_HOST="/dnsaddr/cluster.ipfs.io"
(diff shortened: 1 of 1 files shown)


hsanjuan:
Not sure this script is tested by CI.

cryptonemo:
> Not sure this script is tested by CI.

No, it's not.  I run this manually when need, hopefully not for some time ;-)

cryptonemo:
@hsanjuan Once CI is green, I'll merge

Issue #854, opened 2019-08-30
Distributed PoSt support 


steven004:
### Description
This is a reminder for enabling PoSt to be distributed between servers. This could be done with 1) multiple miner workers support for one miner, and one worker will independently submitPoSt; 2) Allow multiple submitPoSt messages with different provingSet for each in one period; 3) Allow one submitPoSt message contain different proofs of spacetime with specified provingSet. 

This had been discussed with @anorth via email, and @icorderi / @michellebrous and others in Berlin.   

### Acceptance criteria
In a distributed pool, the PoSt can be done on the server node which actually save the data. No data move or share required. 

### Risks + pitfalls
While implementing multiple workers, a new layer is required, more complicated.
While allowing multiple PoSt messages in one period, more messages and pressure for the chain.

### Where to begin
Consider this when implement rational PoSt.

porcuquine:
@cryptonemo In case something here is relevant to your current design work.

cryptonemo:
@magik6k We could really use some focused information on what this would look like.  

I've been looking at this for some time, but have been quite sidetracked as other priorities have popped up.  But it's getting down to the wire now, and we believe that this is needed.  Help me get us there!  Tag in anyone else that you think may be able to help.  I'm tagging @porcuquine for tracking/input as well.

You had proposed a sketch API of something like this:

```
type Challenge struct{SectorID, stuff..}
GenerateChallenges() []Challenge
ReadChallenges([]Challenge) []PoStin // Can work on subset of challenges generated
GenerateWindowPoSt(.., challenges []Challenge)
```

While I understand this at some level, better requirements would be very useful (particularly including the flow, as that's not clear in my head).  For reference, I have not run a multi-node mining setup, but am familiar with single node mining setups.  I don't quite understand the separations across machines at the lotus process level, so bear that in mind.

Some questions I'd like clarified:

Of the machines involved, where is the (storage) data located, and where are the GPUs located?

How can we design this to minimize data transfer, but optimize parallelization/independence?

What is the primary use-case for this?  (e.g. is it making sure WindowPoSt can be offloaded so as not to interfere with WinningPoSt, or just a general performance improvement by distributing the load)?

Regarding the proposed API:

My current understanding is that we can generate challenges and combine all requested inputs for post in some Go struct, and that thing can be shipped (w/o requiring storage/persistence), so long as it's not bigger than hundreds of MiBs (i.e. trees are not involved), correct?

[...]

PR #1269, opened 2020-09-01
feat: split up window post API into separate calls

cryptonemo:
This is a draft PR that is trying to address #854.  This code affects Window PoSt only, and once that is nailed down, Winning PoSt will be updated after (using a similar split).

### DIFF
--- a/filecoin-proofs/src/types/mod.rs
+++ b/filecoin-proofs/src/types/mod.rs
@@ -1,8 +1,13 @@
+use std::collections::BTreeMap;
+
 use serde::{Deserialize, Serialize};
 use storage_proofs::hasher::Hasher;
 use storage_proofs::porep::stacked;
+use storage_proofs::post::fallback::*;
+use storage_proofs::sector::*;
 
 use crate::constants::*;
+use crate::parameters::*;
 
 mod bytes_amount;
 mod piece_info;
@@ -82,4 +87,31 @@ pub struct SealPreCommitPhase1Output<Tree: MerkleTreeTrait> {
     pub labels: Labels<Tree>,
     pub config: StoreConfig,
     pub comm_d: Commitment,
+}
+
+#[derive(Debug, Serialize, Deserialize)]
+pub struct WindowPoStChallenges {
+    pub post_config: PoStConfig,
+    pub randomness: ChallengeSeed,
+    pub prover_id: ProverId,
+    pub sector_challenges: BTreeMap<SectorId, Vec<u64>>,
+}
+
+pub type SnarkProof = Vec<u8>;
+pub type VanillaProof<Tree> = Proof<<Tree as MerkleTreeTrait>::Proof>;
+
+#[derive(Debug, Serialize, Deserialize)]
+pub struct WindowPoStVanillaProofs<Tree: MerkleTreeTrait> {
+    pub vanilla_params: WindowPostSetupParams,
+    pub partitions: Option<usize>,
+    #[serde(bound(
+        serialize = "Vec<PublicSector<<Tree::Hasher as Hasher>::Domain>>: Serialize",
+        deserialize = "Vec<PublicSector<<Tree::Hasher as Hasher>::Domain>>: Deserialize<'de>"
+    ))]
+    pub pub_sectors: Vec<PublicSector<<Tree::Hasher as Hasher>::Domain>>,
+    #[serde(bound(
+        serialize = "Vec<VanillaProof<Tree>>: Serialize",
+        deserialize = "Vec<VanillaProof<Tree>>: Deserialize<'de>"
+    ))]
+    pub vanilla_proofs: Vec<VanillaProof<Tree>>,
 }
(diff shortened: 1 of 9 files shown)


cryptonemo:
2 comments: This is draft work because 1, winning post isn't addressed at all, 2 there is work to be done on prove_vanilla, which does not accept the challenges as directly as it could yet, and 3) while not discussed, all API changes are additive, to preserve the existing call (assuming we wanted that for now).  Expect updates, but early feedback is welcome too.

storage-proofs/core/src/compound_proof.rs
```diff
@@ -70,25 +73,81 @@ where
...
+        pub_params: &PublicParams<'a, S>,
+        pub_in: &S::PublicInputs,
+        vanilla_proofs: Vec<S::Proof>,
+        groth_params: &'b groth16::MappedParameters<Bls12>,
+    ) -> Result<MultiProof<'b>> {
+        let partition_count = Self::partition_count(pub_params);
+
+        // This will always run at least once, since there cannot be zero partitions.
+        ensure!(partition_count > 0, "There must be partitions");
+
```
porcuquine, line 150: We should probably include a sanity check so we never bother trying to create circuit proofs for invalid vanilla proofs.

storage-proofs/core/src/compound_proof.rs
```diff
@@ -70,25 +73,81 @@ where
...
 
         let sanity_check =
             S::verify_all_partitions(&pub_params.vanilla_params, &pub_in, &vanilla_proofs)?;
         ensure!(sanity_check, "sanity check failed");
 
+        Ok(vanilla_proofs)
+    }
+
+    /// prove_vanilla and prove_circuit combined is equivalent to ProofScheme::prove.
+    fn prove_circuit<'b>(
```
porcuquine, line 140: I would call this something like `prove_with_vanilla`.

storage-proofs/post/src/fallback/vanilla.rs
```diff
@@ -355,6 +356,84 @@ impl<'a, Tree: 'a + MerkleTreeTrait> ProofScheme<'a> for FallbackPoSt<'a, Tree>
         Ok(partition_proofs)
     }
 
+    fn prove_all_partitions_with_challenges<'b>(
+        pub_params: &'b Self::PublicParams,
+        pub_inputs: &'b Self::PublicInputs,
+        priv_inputs: &'b Self::PrivateInputs,
+        partition_count: usize,
+        sector_challenges: &BTreeMap<SectorId, Vec<u64>>,
+    ) -> Result<Vec<Self::Proof>> {
```
porcuquine, line 365: I think maybe this is not decomposed right or fully enough. I'm not sure partitions and challenges should be mixed like this. There should be two distinct actions:

- Generate vanillla proofs for a single sector, given the challenges for that sector.
- Group vanilla proofs into partitions.

When decomposed as planned, the vanilla proofs will have been generated separately, but not grouped into partitions.  I think we will need a `partition_vanilla_proofs` (note 'partition' is a verb here) function which just takes all the vanilla proofs and creates partitions proofs from them. (It happens to be the case that both Window and Winning PoSt only use a single partition, but we don't need to and shouldn't rely on that.) This will include the padding logic from `prove_all_partitions` — for the case that fewer than the 'required' number of vanilla proofs are provided.

porcuquine (commented):
@cryptonemo I left some general notes. If you need more clarification or question my suggestions, let's chat and make sure we're on the same page. This is going in the right direction, but I think the underlying structure is not quite right yet.

filecoin-proofs/src/types/mod.rs
```diff
@@ -83,3 +88,30 @@ pub struct SealPreCommitPhase1Output<Tree: MerkleTreeTrait> {
...
+        serialize = "Vec<PublicSector<<Tree::Hasher as Hasher>::Domain>>: Serialize",
+        deserialize = "Vec<PublicSector<<Tree::Hasher as Hasher>::Domain>>: Deserialize<'de>"
+    ))]
+    pub pub_sectors: Vec<PublicSector<<Tree::Hasher as Hasher>::Domain>>,
+    #[serde(bound(
+        serialize = "Vec<VanillaProof<Tree>>: Serialize",
+        deserialize = "Vec<VanillaProof<Tree>>: Deserialize<'de>"
+    ))]
+    pub vanilla_proofs: Vec<VanillaProof<Tree>>,
+}
```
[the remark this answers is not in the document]
porcuquine: What I think may be missing are the sector-specific *private* inputs: `comm_c` and `comm_r_last`. We could technically get the latter from the inclusion proofs, but the former is needed in order to prove the relationship between `comm_r_last` and `comm_r`.

Note that `PublicReplicaInfo` above is effectively just `comm_r`. We could either include `PersistentAux` here, create a new structure just for all the needed commitments, or just individually include `comm_c`, `comm_r_last`, and `comm_r`.

cryptonemo:
Closing in favor of #1278

PR #1272, opened 2020-09-04
Make fil-blst usage in Window PoSt possible

vmx:
This PR consists of three parts:

 - fil-blst based verification to Window PoSt.
 - Logging so that you can easily verify when fil-blst is used.
 - a new CI job is added to run the tests that use the flt-blst code path with `use_fil_blst` enabled.

On the [CI logs](https://app.circleci.com/pipelines/github/filecoin-project/rust-fil-proofs/2318/workflows/2e947ebe-581c-4533-a015-0aadf4ff01d4/jobs/55260) you can see that fil_blst was used (search for `use_fil_blst=true`).

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -127,6 +127,33 @@ jobs:
             RUST_TEST_THREADS: 1
           no_output_timeout: 30m
 
+  # Running with `use_fil_blst=true` should be integrated directly into the test code. For now we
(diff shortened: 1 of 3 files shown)


porcuquine (approved):
LGTM once CI passes. @cryptonemo I think once this merges, we can create another release. That will then contain all the verification optimizations we can provide directly from `rust-fil-proofs`.

PR #1274, opened 2020-09-07
Add FaultySectors error to Fallback PoSt.

porcuquine:
Instead of just failing when trying to generate a Fallback PoSt which contains faulty sectors, instead return a new `FaultySectors` error which contains a vector of faulty `SectorId`s.

In order to make use of the information in the enhanced error, `filecoin-ffi` will need to handle:

https://github.com/filecoin-project/filecoin-ffi/blob/master/rust/src/proofs/api.rs#L606-L609
https://github.com/filecoin-project/filecoin-ffi/blob/master/rust/src/proofs/types.rs#L253-L258

This might entail a breaking change of the `filecoin-ffi` consumed by Go, but the current change to `rust-fil-proofs` is purely additive. It just adds a new, more informative, error in a case where an error would already have been returned.

The included test should give a clear example of how to extract the faulty `SectorId`s from a potentially faulty result.

### DIFF
--- a/storage-proofs/post/src/fallback/vanilla.rs
+++ b/storage-proofs/post/src/fallback/vanilla.rs
@@ -346,23 +346,24 @@ impl<'a, Tree: 'a + MerkleTreeTrait> ProofScheme<'a> for FallbackPoSt<'a, Tree> 
                             challenged_leaf_start as usize,
                             Some(rows_to_discard),
                         );
-                        let valid = if let Ok(proof) = &proof {
-                            proof.validate(challenged_leaf_start as usize)
-                                && proof.root() == priv_sector.comm_r_last
-                        } else {
-                            false
-                        };
-                        if valid {
-                            Ok(ProofOrFault::Proof(proof))
-                        } else {
-                            Ok(ProofOrFault::Fault(sector_id))
+                        match proof {
+                            Ok(proof) => {
+                                if proof.validate(challenged_leaf_start as usize)
+                                    && proof.root() == priv_sector.comm_r_last
+                                {
+                                    Ok(ProofOrFault::Proof(proof))
+                                } else {
+                                    Ok(ProofOrFault::Fault(sector_id))
+                                }
+                            }
+                            Err(_) => Ok(ProofOrFault::Fault(sector_id)),
                         }
                     })
                     .collect::<Result<Vec<_>>>()?
                 {
                     match proof_or_fault {
                         ProofOrFault::Proof(proof) => {
-                            inclusion_proofs.push(proof?);
+                            inclusion_proofs.push(proof);
                         }
                         ProofOrFault::Fault(sector_id) => {
                             faulty_sectors.insert(sector_id);
@@ -386,11 +387,11 @@ impl<'a, Tree: 'a + MerkleTreeTrait> ProofScheme<'a> for FallbackPoSt<'a, Tree> 
             partition_proofs.push(Proof { sectors: proofs });
         }
 
-        ensure!(faulty_sectors.is_empty(), {
-            Error::FaultySectors(faulty_sectors.into_iter().collect())
-        });
-
-        Ok(partition_proofs)
+        if faulty_sectors.is_empty() {
+            Ok(partition_proofs)
+        } else {
+            Err(Error::FaultySectors(faulty_sectors.into_iter().collect()).into())
+        }
     }
 
     fn verify_all_partitions(
@@ -550,13 +551,11 @@ mod tests {
 
         let mut pub_sectors = Vec::new();
         let mut priv_sectors = Vec::new();
-        let mut trees = Vec::new();
 
-        for _i in 0..total_sector_count {
-            let (_data, tree) =
-                generate_tree::<Tree, _>(rng, leaves, Some(temp_path.to_path_buf()));
-            trees.push(tree);
-        }
+        let trees = (0..total_sector_count)
+            .map(|_| generate_tree::<Tree, _>(rng, leaves, Some(temp_path.to_path_buf())).1)
+            .collect::<Vec<_>>();
+
         for (i, tree) in trees.iter().enumerate() {
             let comm_c = <Tree::Hasher as Hasher>::Domain::random(rng);
             let comm_r_last = tree.root();
@@ -695,7 +694,6 @@ mod tests {
                 Err(_) => panic!("failed to downcast to Error"),
                 Ok(Error::FaultySectors(sector_ids)) => assert_eq!(faulty_sectors, sector_ids),
                 Ok(_) => panic!("PoSt failed to return FaultySectors error."),
-                _ => (),
             },
         };
     }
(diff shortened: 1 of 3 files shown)


storage-proofs/post/src/fallback/vanilla.rs
```diff
@@ -562,102 +600,303 @@ mod tests {
...
+        let temp_path = temp_dir.path();
+
+        let mut pub_sectors = Vec::new();
+        let mut priv_sectors = Vec::new();
+
+        let mut trees = Vec::new();
+
+        let mut faulty_sectors = Vec::<SectorId>::new();
+
+        for _i in 0..total_sector_count {
```
vmx, line 634: Just FYI: I think more Rust-y would be something like:

```rust
let trees = 0..total_sector_count.map(|_| {
    generate_tree::<Tree, _>(rng, leaves, Some(temp_path.to_path_buf())).1;
})
```

storage-proofs/post/src/fallback/vanilla.rs
```diff
@@ -562,102 +600,303 @@ mod tests {
...
+                let is_valid =
+                    FallbackPoSt::<Tree>::verify_all_partitions(&pub_params, &pub_inputs, &proof)
+                        .expect("verification failed");
+                assert!(!is_valid, "PoSt returned a valid proof with invalid input");
+            }
+            Err(e) => match e.downcast::<Error>() {
+                Err(_) => panic!("failed to downcast to Error"),
+                Ok(Error::FaultySectors(sector_ids)) => assert_eq!(faulty_sectors, sector_ids),
+                Ok(_) => panic!("PoSt failed to return FaultySectors error."),
+                _ => (),
```
vmx, line 698: Shouldn't all cases be already covered by `Err(_)` and `Ok(`_`)?

storage-proofs/post/src/fallback/vanilla.rs
```diff
@@ -352,6 +386,10 @@ impl<'a, Tree: 'a + MerkleTreeTrait> ProofScheme<'a> for FallbackPoSt<'a, Tree>
             partition_proofs.push(Proof { sectors: proofs });
         }
 
+        ensure!(faulty_sectors.is_empty(), {
```
vmx, line 389: As this isn't really an *early* return, I'd find it easier to understand if it would be written as:

```rust
if (faulty_sectors.is_empty()) {
    Ok(partition_proofs)
} else {
    Err(Error::FaultySectors(faulty_sectors.into_iter().collect()))
}
```

But that's a matter of taste, hence no need to change it.

storage-proofs/post/src/fallback/vanilla.rs
```diff
@@ -332,9 +342,33 @@ impl<'a, Tree: 'a + MerkleTreeTrait> ProofScheme<'a> for FallbackPoSt<'a, Tree>
                             challenge_index,
                         )?;
 
-                        tree.gen_cached_proof(challenged_leaf_start as usize, Some(rows_to_discard))
+                        let proof = tree.gen_cached_proof(
+                            challenged_leaf_start as usize,
+                            Some(rows_to_discard),
+                        );
+                        let valid = if let Ok(proof) = &proof {
```
vmx, line 349: Only FYI (it is correct). I would write this as:

```rust
let valid = match &proof {
    Ok(proof) => {
        proof.validate(challenged_leaf_start as usize)
            && proof.root() == priv_sector.comm_r_last
    }
    Err(_) => false
}
```

I generally use `if let` only if there is no `else`, in all other cases a `match`.

Perhaps I'd even go as far and write the whole part as:

```rust
match &proof {
    Ok(proof) => {
        if proof.validate(challenged_leaf_start as usize)
            && proof.root() == priv_sector.comm_r_last {
            Ok(ProofOrFault::Proof(proof))
        } else {
            Ok(ProofOrFault::Fault(sector_id))
        }
    }
    Err(_) => Ok(ProofOrFault::Fault(sector_id))
}
```

storage-proofs/post/src/fallback/vanilla.rs
```diff
@@ -332,9 +342,33 @@ impl<'a, Tree: 'a + MerkleTreeTrait> ProofScheme<'a> for FallbackPoSt<'a, Tree>
...
+                            false
+                        };
+                        if valid {
+                            Ok(ProofOrFault::Proof(proof))
+                        } else {
+                            Ok(ProofOrFault::Fault(sector_id))
+                        }
                     })
-                    .collect::<Result<Vec<_>>>()?;
+                    .collect::<Result<Vec<_>>>()?
```
vmx, line 361: I don't think you need to `collect()` when you loop over the iterator and are not interested in having a vector in the end.

storage-proofs/post/src/fallback/vanilla.rs
```diff
@@ -332,9 +342,33 @@ impl<'a, Tree: 'a + MerkleTreeTrait> ProofScheme<'a> for FallbackPoSt<'a, Tree>
...
+                {
+                    match proof_or_fault {
+                        ProofOrFault::Proof(proof) => {
+                            inclusion_proofs.push(proof?);
+                        }
+                        ProofOrFault::Fault(sector_id) => {
+                            faulty_sectors.insert(sector_id);
+                        }
+                    }
+                }
```
vmx, line 371: That might not work well with the `into_par_iter()` (I haven't tried), but is the whole for loop and the `ProofOrFault` enum needed at all? Couldn't you just push things directly into the vectors?

```rust
match &proof {
    Ok(proof) => {
        if proof.validate(challenged_leaf_start as usize)
            && proof.root() == priv_sector.comm_r_last {
            inclusion_proofs.push(proof?);
        } else {
            faulty_sectors.insert(sector_id);
        }
    }
    Err(_) => faulty_sectors.insert(sector_id);
}
```

vmx (approved):
Things look correct to me, I've just left comments that are coding style/simplification things that I would do. That's FYI as I find it useful to see how others approach things.

cryptonemo (approved):
This looks good to me.  Up to you if you want to address @vmx's feedback

storage-proofs/post/src/fallback/vanilla.rs
porcuquine: Yes, that's vestigial, thanks.

storage-proofs/post/src/fallback/vanilla.rs
porcuquine: I actually agree. I changed it because what you wrote (and I tried first) wouldn't compile. I figured out how to fix it, though, so thanks for the nudge.

```
        if faulty_sectors.is_empty() {
            Err(Error::FaultySectors(faulty_sectors.into_iter().collect()).into())
        } else {
            Ok(partition_proofs)
        }
```

storage-proofs/post/src/fallback/vanilla.rs
porcuquine: That's what I thought too, but sadly this was the only one of the obvious alternatives I could get to compile.

storage-proofs/post/src/fallback/vanilla.rs
porcuquine: Agreed, this case (though not necessarily about always eschewing `else`).

cryptonemo (approved):

PR #1275, opened 2020-09-09
refactor: simplify the code that reports faulty sectors

vmx:
The code was using a parallel iterator, whose result was then processed
sequentially in a for loop. I'm not sure how much parallization that would
give us. Hence I decided to just use a sequntial (non-parallel) iterator
instead.

The alternative would be using a a parallel iterator with mutex around
the vectors, which seems overkill for this use case.

@porcuquine given you mentioned that you'd also like to see a simpler version, I couldn't resist.

### DIFF
--- a/storage-proofs/post/src/fallback/vanilla.rs
+++ b/storage-proofs/post/src/fallback/vanilla.rs
@@ -6,7 +6,6 @@ use byteorder::{ByteOrder, LittleEndian};
 use generic_array::typenum::Unsigned;
 use log::trace;
 use paired::bls12_381::Fr;
-use rayon::prelude::*;
 use serde::{Deserialize, Serialize};
 use sha2::{Digest, Sha256};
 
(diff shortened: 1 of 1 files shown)


porcuquine (changes requested):
I would be okay with a change like this if it preserved parallelism, but I also agree that adding a Mutex here would be overkill. It would just be an alternate way of trying to bend the code toward a desired shape — but one that adds noise related to the concurrency.

There's nothing inherently wrong with sequentially processing the results of a collection which was assembled in parallel, so I would just leave this as is since we do want/need the parallelism. (It's probably worth stopping to think about why removing it would be substantive, to lend confidence to ongoing refactoring efforts.)

vmx:
Oh right. I first thought there won't be any real parallelism as it is a sequential for loop. But it is kind of a scatter-gather thing. I'm closing it for now.

PR #1278, opened 2020-09-10
feat: split up window post API into separate calls

cryptonemo:
An alternate PR/proposal to #1269.  Winning PoSt is not working at the moment.

### DIFF
--- a/filecoin-proofs/src/api/post.rs
+++ b/filecoin-proofs/src/api/post.rs
@@ -16,6 +16,7 @@ use storage_proofs::merkle::{
 };
 use storage_proofs::multi_proof::MultiProof;
 use storage_proofs::post::fallback;
+use storage_proofs::post::fallback::SectorProof;
 use storage_proofs::proof::ProofScheme;
 use storage_proofs::sector::*;
 use storage_proofs::settings;
@@ -294,6 +295,7 @@ pub fn generate_winning_post_with_vanilla<Tree: 'static + MerkleTreeTrait>(
 
     let partitions = 1;
     let partitioned_proofs = partition_vanilla_proofs(
+        &post_config,
         &pub_params.vanilla_params,
         &pub_inputs,
         partitions,
@@ -432,9 +434,9 @@ pub fn generate_winning_post_sector_challenge<Tree: MerkleTreeTrait>(
     result
 }
 
-/// Generates the challenges per SectorId required for a Window proof-of-spacetime.
-// FIXME: rename to avoid confusion with fallback::generate_sector_challenges
-pub fn generate_sector_challenges<Tree: 'static + MerkleTreeTrait>(
+/// Generates the challenges per SectorId required for either a Window
+/// proof-of-spacetime or a Winning proof-of-spacetime.
+pub fn generate_fallback_sector_challenges<Tree: 'static + MerkleTreeTrait>(
     post_config: &PoStConfig,
     randomness: &ChallengeSeed,
     pub_sectors: &[SectorId],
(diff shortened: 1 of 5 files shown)


filecoin-proofs/src/api/post.rs
```diff
@@ -241,7 +244,76 @@ pub fn clear_caches<Tree: MerkleTreeTrait>(
     Ok(())
 }
 
-pub type SnarkProof = Vec<u8>;
+/// Generates a Window proof-of-spacetime with provided vanilla proofs.
```
cryptonemo, line 247: Typo

filecoin-proofs/src/api/post.rs
```diff
@@ -461,6 +644,199 @@ pub fn verify_winning_post<Tree: 'static + MerkleTreeTrait>(
...
+
+            for pub_sector in sectors_chunk.iter() {
+                let cur_proof = vanilla_proofs
+                    .iter()
+                    .find(|&proof| proof.sector_id == pub_sector.id)
+                    .expect("failed to locate sector proof");
+
+                // Note: Window post requires all inclusion proofs (based on the challenge
+                // count per sector) per sector proof.
+                proofs.extend(cur_proof.vanilla_proof.sectors.clone());
```
porcuquine, line 692: This is confusing. It reads like `proofs` is a vector of sectors? Something is poorly named.

porcuquine (commented):
I think this is good. The general form and structure is much better than before and basically as I would expect. There are some rough edges in terms of shape/names and also where some assumptions feel too baked in. Getting this right is tricky, and as you suggest, we may need to make another pass to improve things even after moving forward to release this feature. I think one round of effort to incorporate at least the low-hanging fruit in my comments may be worthwhile. I'd also like to take one more look at it, perhaps after just a little more cleanup, in light of having had time to absorb it. As I said, I think everything looks right, and passing tests is good. However, there *are* ways a refactor like this could lead to hard-to-detect changes of meaning, so we should be sure before merging. Just a little more scrubbing (on your part) and study/thought (on mine) should get us there.

vmx:
I don't have a enough knowledge to do a review about the correctness. I saw that there are quite some iterators which are not parallel ones. I don't know if that matter and I also know that many of them are likely hard to fix. Parallel iterators are hard to do with the current code, there are just too many side effects in the loops. In order to make it easier, the coding style would need to change to a more functional, side-effect free style.

I guess making everything parallelizable is probably out of scope for this PR, I just wanted to mention it to keep it in mind for future refactors.

filecoin-proofs/src/api/post.rs
```diff
@@ -362,6 +434,117 @@ pub fn generate_winning_post_sector_challenge<Tree: MerkleTreeTrait>(
...
+    challenges: &[u64],
+) -> Result<FallbackPoStVanillaProof<Tree>> {
+    let sector_count = 1;
+
+    info!("generate_single_vanilla_proof:start");
+    ensure!(
+        post_config.sector_count == sector_count,
+        "invalid post config sector size ({} required)",
+        sector_count
+    );
```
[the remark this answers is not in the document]
cryptonemo: The comment was incorrect and has been updated.

In this case we check to make sure it's 1 because the way the arguments are structured, there's only enough information to generate a single vanilla proof.  I've updated this and added a comment saying this, but let me know if it needs another pass.

filecoin-proofs/src/api/post.rs
```diff
@@ -460,15 +463,16 @@ pub fn generate_fallback_sector_challenges<Tree: 'static + MerkleTreeTrait>(
...
+    let partitions = match post_config.typ {
+        PoStType::Window => match get_partitions_for_window_post(pub_sectors.len(), &post_config) {
             Some(x) => x,
             None => 1,
-        }
-    } else {
-        assert_eq!(post_config.typ, PoStType::Winning);
+        },
+        PoStType::Winning => {
+            assert_eq!(post_config.typ, PoStType::Winning);
```
porcuquine, line 472: This assertion is quite redundant here. The previous line literally guarantees this will never fail, so I think it can be removed. I would instead add a comment justifying the decision. Since I don't think the number of partitions is _structurally_ guaranteed to be one, I would probably add a comment justifying this decision. That comment should make clear what change to the code base (even if it can't be conveniently checked here) would necessitate a change here.

porcuquine (approved):
Looks good.

filecoin-proofs/src/api/post.rs
cryptonemo: Agreed, that looks like a hold-over from the previous if/else branch change.  Rather than update and push, I'm opting to merge this and update in a future PR.

PR #1280, opened 2020-09-14
fix: improve sector ID logging

### LINKED ISSUE #1261: Suggest to add more sector information in case of fault.
### Description
Some time it's hard to fix the sectors fault, but it's really need in production.

For example like do the window PoSt, one sector broken could make the deadline broken, some case it's very hard to exclude which sector is  errors.

Suggest to add more about sector information in case of error to return or log which sector is in fault. i can make a fork, but i really not want to make more forks, they are really hard to managed \o/

Example:
https://github.com/filecoin-project/rust-fil-proofs/compare/master...IPFS-grandhelmsman:feat/log

image

vmx:
Sector IDs are more often included in log messages/attached to errors.

This PR is based on @IPFS-grandhelmsman work at #1261.

Closes #1261.

I've only gone over the files that were suggested in #1261, if there are more places where logging should be added, please let me know.

### DIFF
--- a/filecoin-proofs/src/api/post.rs
+++ b/filecoin-proofs/src/api/post.rs
@@ -127,12 +127,12 @@ impl<Tree: 'static + MerkleTreeTrait> PrivateReplicaInfo<Tree> {
         as_safe_commitment(&self.comm_r, "comm_r")
     }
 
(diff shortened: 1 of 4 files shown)


cryptonemo (approved):
Looks good, thanks!

PR #1284, opened 2020-09-21
feat: proof upgrade testing

cryptonemo:
This PR is NOT intended to be merged, as it contains incompatible proofs behaviour.  At the moment, this branch is for testing.

### DIFF
--- a/fil-proofs-tooling/src/bin/benchy/prodbench.rs
+++ b/fil-proofs-tooling/src/bin/benchy/prodbench.rs
@@ -15,6 +15,7 @@ use log::info;
 use rand::SeedableRng;
 use rand_xorshift::XorShiftRng;
 use serde::{Deserialize, Serialize};
+use storage_proofs::api_version::APIVersion;
 use storage_proofs::compound_proof::CompoundProof;
 use storage_proofs::hasher::Sha256Hasher;
 #[cfg(feature = "measurements")]
@@ -23,6 +24,8 @@ use storage_proofs::measurements::Operation;
 use storage_proofs::measurements::OP_MEASUREMENTS;
 use storage_proofs::parameter_cache::CacheableParameters;
 use storage_proofs::proof::ProofScheme;
+
+const FIXED_API_VERSION: APIVersion = APIVersion::V1_0;
 
 const SEED: [u8; 16] = [
     0x59, 0x62, 0xbe, 0x5d, 0x76, 0x3d, 0x31, 0x8d, 0x17, 0xdb, 0x37, 0x32, 0x54, 0x06, 0xbc, 0xe5,
@@ -188,6 +191,7 @@ pub fn run(
         inputs.num_sectors as usize,
         only_add_piece,
         arbitrary_porep_id,
+        FIXED_API_VERSION,
     );
 
     if only_add_piece || only_replicate {
@@ -313,6 +317,7 @@ fn generate_params(i: &ProdbenchInputs) {
         sector_size,
         partitions,
         porep_id: dummy_porep_id,
+        api_version: FIXED_API_VERSION,
     });
 }
 
@@ -320,11 +325,10 @@ fn cache_porep_params(porep_config: PoRepConfig) {
     use filecoin_proofs::parameters::public_params;
     use storage_proofs::porep::stacked::{StackedCompound, StackedDrg};
 
-    let dummy_porep_id = [0; 32];
     let public_params = public_params(
         PaddedBytesAmount::from(porep_config),
         usize::from(PoRepProofPartitions::from(porep_config)),
-        dummy_porep_id,
+        porep_config.porep_id,
     )
     .expect("failed to get public_params");
 
--- a/fil-proofs-tooling/src/bin/benchy/prodbench.rs
+++ b/fil-proofs-tooling/src/bin/benchy/prodbench.rs
@@ -17,7 +17,7 @@ use log::info;
 use rand::SeedableRng;
 use rand_xorshift::XorShiftRng;
 use serde::{Deserialize, Serialize};
-use storage_proofs::api_version::APIVersion;
+use storage_proofs::api_version::ApiVersion;
 use storage_proofs::compound_proof::CompoundProof;
 use storage_proofs::hasher::Sha256Hasher;
 #[cfg(feature = "measurements")]
@@ -26,8 +26,6 @@ use storage_proofs::measurements::Operation;
 use storage_proofs::measurements::OP_MEASUREMENTS;
 use storage_proofs::parameter_cache::CacheableParameters;
 use storage_proofs::proof::ProofScheme;
-
-const FIXED_API_VERSION: APIVersion = APIVersion::V1_0;
 
 const SEED: [u8; 16] = [
     0x59, 0x62, 0xbe, 0x5d, 0x76, 0x3d, 0x31, 0x8d, 0x17, 0xdb, 0x37, 0x32, 0x54, 0x06, 0xbc, 0xe5,
@@ -59,8 +57,8 @@ impl ProdbenchInputs {
     pub fn sector_size_bytes(&self) -> u64 {
         bytefmt::parse(&self.sector_size).expect("failed to parse sector size")
     }
-    pub fn api_version(&self) -> APIVersion {
-        APIVersion::from_str(&self.api_version).expect("failed to parse api version")
+    pub fn api_version(&self) -> ApiVersion {
+        ApiVersion::from_str(&self.api_version).expect("failed to parse api version")
     }
 }
 
@@ -323,7 +321,7 @@ fn generate_params(i: &ProdbenchInputs) {
         sector_size,
         partitions,
         porep_id: dummy_porep_id,
-        api_version: FIXED_API_VERSION,
+        api_version: i.api_version(),
     });
 }
 
--- a/fil-proofs-tooling/src/bin/benchy/prodbench.rs
+++ b/fil-proofs-tooling/src/bin/benchy/prodbench.rs
@@ -1,3 +1,5 @@
+use std::str::FromStr;
+
 use bellperson::bls::Bls12;
 use bellperson::util_cs::bench_cs::BenchCS;
 use bellperson::Circuit;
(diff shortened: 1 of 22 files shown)


filecoin-proofs/src/api/post.rs
```diff
@@ -604,18 +595,18 @@ pub fn verify_winning_post<Tree: 'static + MerkleTreeTrait>(
...
-            pub_sectors.push(fallback::PublicSector {
+            Ok(fallback::PublicSector {
                 id: *sector_id,
                 comm_r,
-            });
-        }
-    }
+            })
+        })
+        .collect::<Result<_>>()?;
```
porcuquine: I think the problem with this change is that it assumed each sector could only appear once in the set of challenged sectors. But in the case of Winning PoSt, there is just one sector, yet it is challenged repeatedly (each time with only a single challenge).

Yes, this is strange and confusing. Yes, the pervasive naming makes this almost impossible to understand.

filecoin-proofs/src/api/post.rs
```diff
@@ -452,6 +452,7 @@ pub fn generate_fallback_sector_challenges<Tree: 'static + MerkleTreeTrait>(
     randomness: &ChallengeSeed,
     pub_sectors: &[SectorId],
     _prover_id: ProverId,
+    is_winning: bool,
```
porcuquine, line 455: If we want to avoid completely splitting Window and Winning PoSt, we need to be able to signal which we are using. We could try to 'detect' somehow based on the actual values, but that would be very brittle and a bad path to follow.

filecoin-proofs/src/api/post.rs
```diff
@@ -595,18 +598,18 @@ pub fn verify_winning_post<Tree: 'static + MerkleTreeTrait>(
...
-            Ok(fallback::PublicSector {
+            pub_sectors.push(fallback::PublicSector {
                 id: *sector_id,
                 comm_r,
-            })
-        })
-        .collect::<Result<_>>()?;
+            });
+        }
+    }
```
porcuquine: We need to roll this back. Winning PoSt will not actually change its behavior. We need the challenges of Winning PoSt to be sensitive to order, because otherwise they would all be the same. This is because a single challenge appears repeatedly at different positions within the list of challenged sectors. Each instance needs fresh challenges.

storage-proofs/post/src/fallback/compound.rs
```diff
@@ -68,7 +68,11 @@ impl<'a, Tree: 'static + MerkleTreeTrait>
                     &pub_params,
                     pub_inputs.randomness,
                     sector.id.into(),
-                    challenge_index as u64,
+                    if pub_params.is_winning {
+                        i
+                    } else {
+                        challenge_index
+                    } as u64,
```
porcuquine, line 75: This works, but it's a hack which now depends on each Winning PoSt sector being challenged only once. To be more correct, the `true` branch of this conditional should fully restore previous behavior. In practice, this won't matter with current parameterization, but we shouldn't depend on that. Certainly we would need at least an assertion if we intended to. That might actually be preferable. We have an enormous amount of complexity in place to support a generality I don't think we need. If we can close off some possibilities for the sake of simpler, clearer, code that might be worth doing.

storage-proofs/post/src/fallback/compound.rs
```diff
@@ -234,6 +238,7 @@ mod tests {
                 sector_size: sector_size as u64,
                 challenge_count,
                 sector_count,
+                is_winning: false,
```
porcuquine, line 241: For the sake of getting this working quickly, I ended up specializing all the tests to Window PoSt (`is_winning: false`). To be thorough, they should probably be extended to exercise both code paths.

filecoin-proofs/tests/api.rs
```diff
@@ -375,8 +373,11 @@ fn window_post<Tree: 'static + MerkleTreeTrait>(
     /////////////////////////////////////////////
     // The following methods of proof generation are functionally equivalent:
     // 1)
-    //let proof = generate_window_post::<Tree>(&config, &randomness, &priv_replicas, prover_id)?;
-    //
+    let proof = generate_window_post::<Tree>(&config, &randomness, &priv_replicas, prover_id)?;
```
porcuquine, line 376: Commit description is misleading (mentioning for anyone else trying to make sense of it). It's Window PoSt that is affected.

porcuquine: No need.

storage-proofs/post/src/fallback/vanilla.rs
```diff
@@ -380,16 +383,28 @@ impl<'a, Tree: 'a + MerkleTreeTrait> ProofScheme<'a> for FallbackPoSt<'a, Tree>
                     Tree::Arity::to_usize(),
                 );
 
+                let num_challenges = if pub_params.is_winning {
+                    pub_params.sector_count
```
porcuquine, line 387: I think this depends on `pub_params.challenge_count` being equal to 1. Since that's not structurally guaranteed (but will be the case in extant WindowPoSt), I think it would be good to also assert it here. I'd rather be safe than sorry, at least until we can clean this up more broadly.

storage-proofs/post/src/fallback/compound.rs
```diff
@@ -65,10 +64,16 @@ impl<'a, Tree: 'static + MerkleTreeTrait>
...
             for n in 0..pub_params.challenge_count {
-                let challenge_index = ((partition_index * pub_params.sector_count + i)
-                    * pub_params.challenge_count
-                    + n) as u64;
-                let challenged_leaf_start = fallback::generate_leaf_challenge(
+                let challenge_index = if pub_params.is_winning {
+                    // Note that this generality is perhaps over-complicated and unnecessary
+                    // with the current parameterization.  To avoid complexity, the
+                    // challenge_index could be set to 'i' here.
+                    (partition_index * pub_params.sector_count + i) * pub_params.challenge_count + n
```
porcuquine, line 71: Can we guard the simplifying assumption with an assertion then revert to the simple expression? That works for me.

storage-proofs/post/src/fallback/vanilla.rs
```diff
@@ -427,11 +434,25 @@ impl<'a, Tree: 'a + MerkleTreeTrait> ProofScheme<'a> for FallbackPoSt<'a, Tree>
...
                 }
 
-                proofs.push(SectorProof {
-                    inclusion_proofs,
-                    comm_c: priv_sector.comm_c,
-                    comm_r_last: priv_sector.comm_r_last,
-                });
+                // Winning PoSt and Window PoSt have different proof shapes regarding inclusion
+                // and sector proofs, so we are careful to partition them appropriately.
+                if pub_params.is_winning {
```
porcuquine, line 439: My experience of reading this whole section now is that it feels obfuscated and is very hard to follow. I am starting to suspect it would be better to just select on type (`is_winning`) then clearly and concisely perform the appropriate work in each branch. Treating these as 'same but different' with all the conditionals, hidden assumptions, and gymnastics doesn't seem worth it. If nothing else, it makes the code hard to follow — and it also makes it hard to reason about well enough to gain confidence. I think that goal might be more achievable with the alternate structure I just proposed. What do you think?

cryptonemo: It's a good idea for sure -- but the reason I'd hold off is because we already do something like this here (arguably the 'improved' version, or at least inspired by the idea): https://github.com/filecoin-project/rust-fil-proofs/blob/master/filecoin-proofs/src/api/post.rs#L672 and ideally, as part of a larger re-factor, we can share code across as needed better (rather than duplicating).

porcuquine: Do you want to do that now then? My concern is trying to avoid accidentally releasing broken code. I'm not saying I believe this code is necessarily broken, but if we let the mandate of an upgrade complicate the code base, then we're at risk of making it very hard to continue maintaining in case further change becomes required.

I would personally prefer to maximize simplicity, even at the cost of some 'duplication' which can be refactored away later. Or do the refactor now, if not too difficult. The only reason the code isn't duplicated now is because we have two *different* methods, one of which is very hard to comprehend.

porcuquine:
@cryptonemo I think my previous commit is ready for review. This is about all the refactoring that feels needed for now (apart from whatever is logically implied that I might have overlooked).

fil-proofs-tooling/src/bin/benchy/prodbench.rs
```diff
@@ -313,18 +317,18 @@ fn generate_params(i: &ProdbenchInputs) {
         sector_size,
         partitions,
         porep_id: dummy_porep_id,
+        api_version: FIXED_API_VERSION,
```
dignifiedquire, line 320: this should probably a parameter of the tooling

fil-proofs-tooling/src/bin/benchy/window_post.rs
```diff
@@ -514,6 +516,7 @@ pub fn run_window_post_bench<Tree: 'static + MerkleTreeTrait>(
             .expect("unknown sector size"),
         typ: PoStType::Window,
         priority: true,
+        api_version: APIVersion::V1_1,
```
dignifiedquire, line 519: should probably a parameter

storage-proofs/core/src/api_version.rs
```diff
@@ -1,5 +1,21 @@
...
+
 #[derive(Debug, Copy, Clone)]
 pub enum APIVersion {
     V1_0,
     V1_1,
 }
+
+impl FromStr for APIVersion {
+    type Err = anyhow::Error;
+    fn from_str(api_version: &str) -> Result<Self> {
```
dignifiedquire, line 12: Lets associate a `Semver` version with each of the enum and use that for parsing using: https://docs.rs/semver/

storage-proofs/core/src/api_version.rs
```diff
@@ -1,5 +1,21 @@
+use anyhow::{self, Result};
+use std::str::FromStr;
+
 #[derive(Debug, Copy, Clone)]
 pub enum APIVersion {
```
dignifiedquire, line 5: Lets call this `ApiVersion`, it is a little bit more consistent with naming conventions in rust

storage-proofs/core/src/api_version.rs
cryptonemo: For this, do we want `ApiVersion::{V1_0_0, V1_1_0}` for the constants also?

dignifiedquire: good question, I am not sure tbh

PR #1289, opened 2020-09-25
Optimize Phase1.

porcuquine:
This implements Sealing optimizations but needs a little more cleanup.

@dignifiedquire An early step in this was simplification of the previous sealing (an intentional regression to facilitate moving forward). Now that we have decided to retain previous behavior and make the optimizations (initially) opt-in, some work is required to restore the previous behavior. I started in on this, but it's incomplete, as you will see.

If you have some time and the will to do it, you could reenable the previous behavior. The setting-based selection is implemented, but the `generate_labels` code path now calls the simplified, intermediate code. Instead, it should call the previous code — which is deleted in this PR's first commit. Once that is restored, this work will be almost done.

- [x] Restore previous sealing behavior as default.
- [x] Document settings and how to invoke new sealing in order to maximize performance.


fil-proofs-tooling/src/bin/settings/main.rs
```diff
@@ -3,49 +3,6 @@ use anyhow::Result;
...
-        window_post_synthesis_num_cpus
-    );
-    println!(
-        "pedersen_hash_exp_window_size: {}",
-        pedersen_hash_exp_window_size
-    );
-
-    println!("use_fil_blst: {}", use_fil_blst);
-
+    println!("{:#?}", *SETTINGS.lock().unwrap());
```
porcuquine, line 6: I would really like to do this. Having to manually maintain the previous is a pain, and if all we want is to print all the settings, this should be fine.

storage-proofs/porep/src/stacked/vanilla/proof.rs
```diff
@@ -1005,7 +936,7 @@ impl<'a, Tree: 'static + MerkleTreeTrait, G: 'static + Hasher> StackedDrg<'a, Tr
             other_arity_valid
         );
         assert!(binary_arity_valid);
-        assert!(other_arity_valid);
+        // assert!(other_arity_valid);
```
cryptonemo, line 939: Guessing this could be added back in?

storage-proofs/porep/src/stacked/vanilla/create_label/multi.rs
```diff
@@ -0,0 +1,681 @@
...
+        hasher.finalize()
+    } else {
+        let cache_parents = &cache.read(node as u32)?;
+
+        // prefetch previous node, which is always a parent
+        let prev = &layer_labels[(node - 1) * NODE_SIZE..node * NODE_SIZE];
+        prefetch!(prev.as_ptr() as *const i8);
+
+        if first_layer {
+            let mut parents = [0u8; BASE_DEGREE * NODE_SIZE]; // FIXME: Don't hardcode 14.
```
cryptonemo, line 529: FIXME

storage-proofs/porep/src/stacked/vanilla/create_label/multi.rs
```diff
@@ -0,0 +1,681 @@
...
+            fetch(&cache_parents[0..BASE_DEGREE], layer_labels, &mut parents);
+
+            hash_fixed_size(&mut hasher, 37, &parents);
+
+            hasher.finalize()
+
+        // copy_parents_data_exp(node as u32, &*layer_labels, exp_labels, hasher, cache)?
+        } else {
+            let mut parents = [0u8; 14 * NODE_SIZE]; // FIXME: Don't hardcode 14.
```
cryptonemo, line 538: FIXME

storage-proofs/porep/src/stacked/vanilla/create_label/multi.rs
porcuquine: Thanks. It was fixed, just needs comment removed. However, this is a good reminder that this is all dead code now anyway. Will remove in next push.

storage-proofs/porep/src/stacked/vanilla/proof.rs
porcuquine: done

storage-proofs/porep/src/stacked/vanilla/memory_handling.rs
```diff
@@ -0,0 +1,255 @@
...
+        unsafe { &std::slice::from_raw_parts((*self.bufs.get()).as_ptr(), 2) }
+    }
+
+    #[inline]
+    #[allow(clippy::mut_from_ref)]
+    unsafe fn get_mut_bufs(&self) -> &mut [Mmap] {
+        std::slice::from_raw_parts_mut((*self.bufs.get()).as_mut_ptr(), 2)
+    }
+
+    // TODO: is this actually needed?
```
porcuquine, line 75: It's not currently used. It was split into the start and finish portions and split across layer-transition boundaries. I think it remains useful documentation of intent. It could probably be refactored into a call to start then finish, which would serve that purpose even better.

dignifiedquire (approved):

PR #1292, opened 2020-09-25
feat(porep): skip existing layers if they already exist

dignifiedquire:


storage-proofs/porep/src/stacked/vanilla/create_label/multi.rs
```diff
@@ -399,13 +400,101 @@ fn create_layer_labels(
...
     config: StoreConfig,
-) -> Result<(LabelsCache<Tree>, Labels<Tree>)> {
+) -> Result<(Labels<Tree>, Vec<LayerState>)> {
+    info!("create labels");
+
+    let layer_states = super::prepare_layers::<Tree>(graph, &config, layers);
+
+    let sector_size = graph.size() * NODE_SIZE;
+    let node_count = graph.size() as u64;
+    let cache_window_nodes = settings::SETTINGS
```
cryptonemo, line 416: Var name and setting name mismatch?

dignifiedquire: not entirely sure what is correct here, @porcuquine ?

cryptonemo (approved):
@porcuquine Please take a look

storage-proofs/porep/src/stacked/vanilla/create_label/multi.rs
cryptonemo: The convention I've been going with is using the var name as the setting name

porcuquine: Looking now. Pro tip: put your comment after all the context, otherwise it gets cut off.

porcuquine: I think a general policy of naming variables after settings is fine. I also think it's fine to have local assignments be more specific. The setting needs to communicate context which may already be present at the point of use. In this case, I think `cache_window_nodes` is a good example of providing more specific information relevant to usage. Locally, we need to know that the unit is nodes. At the level of global settings, it's more important to know this value relates to the SDR parent cache's size — but the units are less important, since we don't advertise this as being in bytes, for example. If we did want to rectify, the right answer might be to change the setting to include `nodes`, but I don't think we should change published settings lightly. This is all a long way of saying, I think it's fine in this case, and also fine to prefer matching setting and variable names when all else is equal.

porcuquine (approved):

PR #1294, opened 2020-09-28
feat(storage-proofs-porep): implement prefetch macro for aarch64

dgbo:
Co-authored-by: Wang Maozhang <wangmaozhang@huawei.com>

This implement prefetch marco for aarch64 in `storage-proofs/porep/src/stacked/vanilla/macros.rs`, with prefetch hints merged into core_arch library [rust-lang/stdarch-PR918](https://github.com/rust-lang/stdarch/pull/918) and [rust-lang/rust-PR77259](https://github.com/rust-lang/rust/pull/77259).

Patch delivers about 23% performance improvement in `(lotus-bench seal: preCommit phase 1)` with `--sector-size=512MB/32GiB` on Kunpeng920 aarch64 server.

### DIFF
--- a/storage-proofs/porep/src/lib.rs
+++ b/storage-proofs/porep/src/lib.rs
@@ -1,6 +1,8 @@
 //requires nightly, or later stable version
 //#![warn(clippy::unwrap_used)]
 
+#![cfg_attr(target_arch = "aarch64", feature(stdsimd))]
+
 pub mod drg;
 pub mod stacked;
 
--- a/storage-proofs/porep/src/stacked/vanilla/macros.rs
+++ b/storage-proofs/porep/src/stacked/vanilla/macros.rs
@@ -55,6 +55,11 @@ macro_rules! prefetch {
 
             _mm_prefetch($val, _MM_HINT_T0);
         }
+        #[cfg(all(target_arch = "aarch64"))]
+        unsafe {
+            use std::arch::aarch64::*;
+            _prefetch($val, _PREFETCH_READ, _PREFETCH_LOCALITY3);
+        }
     };
 }
 


cryptonemo:
Thanks! Will see if we can find a tester for this.

dgbo:
> Thanks! Will see if we can find a tester for this.

The result on our server are very good. 
Looking forward to see the results on other aarch64 based platforms.

Thanks.

storage-proofs/porep/src/lib.rs
```diff
@@ -1,6 +1,8 @@
 //requires nightly, or later stable version
 //#![warn(clippy::unwrap_used)]
 
+#![cfg_attr(target_arch = "aarch64", feature(stdsimd))]
```
dignifiedquire, line 4: this means that aarch64 now requires a nightly compiler, which is fine, but should be documented

dignifiedquire (approved):
thanks

storage-proofs/porep/src/lib.rs
dgbo: Thanks for the quick review.

Seems the nightly compiler requirement has been documented in README.md already:
275 ## Building for Arm64
276
277 In order to build for arm64 the current requirements are
278
279 - nightly rust compiler
280
281 Example for building `filecoin-proofs`
282
283 ```
284 $ rustup +nightly target add aarch64-unknown-linux-gnu
285 $ cargo +nightly build -p filecoin-proofs --release --target aarch64-unknown-linux-gnu
286 ```

Do we need extra comments in `storage-proofs/porep/src/lib.rs` before the cfg_attr line or elsewhere? Maybe something like `//requires nightly, or later stable version to build for aarch64`.

Best Regards.

dignifiedquire: ha, I forgot about that, perfect thank you!

PR #1295, opened 2020-09-28
refactor(storage-proofs-porep): improve unsafe code structure

dignifiedquire:


storage-proofs/porep/src/stacked/vanilla/memory_handling.rs
```diff
@@ -197,26 +194,32 @@ impl<T: FromByteSlice> CacheReader<T> {
...
             self.window_size as usize,
             &self.file,
         )
         .unwrap();
 
-        unsafe {
-            self.get_mut_bufs()[replace_idx] = new_buf;
-        }
+        let buf = if new_window % 2 == 0 {
+            &mut *self.buf0.get()
```
cryptonemo, line 217: I think the buf should be alternating with this conditional, no?

porcuquine:
This seems to mainly replace `%` operations with manual if-else statements, adding verbosity. I think the original code actually does exactly what we want and generalizes to more than two buffers. Not that we need more than two in practice, but that's an example of why I don't think we inherently need to special-case this. The general handling accurately and concisely does exactly what is needed.

dignifiedquire:
not merging for now

PR #1305, opened 2020-10-06
bind replication threads to specific cores

porcuquine:
In order to maximize replication (precommit phase 1) performance, bind all replication threads to a single core (distinct from other replication threads). Consumer and producer threads for each replication task should be on cores which share cache. If multiple such cores exist, keep threads for a single replication task together (on cores sharing a cache). Otherwise, group cores based on the number of threads required. We create groups even when they do not correspond to a shared cache in order to make use of the core binding facility used in the preferred case.

If a core cannot be bound, a message will be logged in a debug mode, but operation is not interrupted. The thread is simply scheduled without constraint as though no attempt to bind a core had been made.

NOTE: This implementation does not support currently support core binding on macos. 

- [x] Document in README.
- [x] Identify and remove minimal cause of perf regression (vs best seen in this work, not vs previous).

### PERF UPDATES

- I benchmarked replication of a 32GiB sector with the final code at 2:03:05 (on 3970x) — so regression is conquered.
- I benchmarked the currently-latest version without mlocked window pages (and not run as root) and saw a noticeably faster result: 2:02:23.
- 2 sectors, with window pages locked: 2:06:27
- 2 sectors, without window pages locked 2:06:17

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -314,6 +314,16 @@ jobs:
       - configure_environment_variables
       - checkout
       - run:
+          name: Install hwloc 1.11.9
+          command: |
+            cd /tmp
+            curl https://www.open-mpi.org/software/hwloc/v1.11/downloads/hwloc-1.11.9.tar.gz --location --output /tmp/hwloc-1.11.9.tar.gz
+            tar xzvf hwloc-1.11.9.tar.gz
+            cd hwloc-1.11.9
+            ./configure
+            make
+            sudo make install
+      - run:
           name: Install Rust
           command: |
             curl https://sh.rustup.rs -sSf | sh -s -- -y
--- a/storage-proofs/porep/Cargo.toml
+++ b/storage-proofs/porep/Cargo.toml
@@ -37,6 +37,8 @@ bincode = "1.1.2"
 byteorder = "1.3.4"
 lazy_static = "1.2"
 byte-slice-cast = "0.3.5"
+hwloc = "0.3.0"
+libc = "0.2"
 
 [dev-dependencies]
 tempfile = "3"
--- a/storage-proofs/porep/src/stacked/vanilla/cores.rs
+++ b/storage-proofs/porep/src/stacked/vanilla/cores.rs
@@ -1,0 +1,219 @@
+use log::*;
+use std::sync::{Mutex, MutexGuard};
+
+use anyhow::Result;
+use hwloc::{ObjectType, Topology, TopologyObject, CPUBIND_THREAD};
+use lazy_static::lazy_static;
+
+use storage_proofs_core::settings;
+
+type CoreGroup = Vec<CoreIndex>;
+lazy_static! {
+    pub static ref TOPOLOGY: Mutex<Topology> = Mutex::new(Topology::new());
+    pub static ref CORE_GROUPS: Option<Vec<Mutex<CoreGroup>>> = {
+        let settings = settings::SETTINGS.lock().expect("settings lock failure");
+        let num_producers = settings.multicore_sdr_producers;
+        let cores_per_unit = num_producers + 1;
+
+        core_groups(cores_per_unit)
+    };
+}
+
+#[derive(Clone, Copy, Debug, PartialEq)]
+/// `CoreIndex` is a simple wrapper type for indexes into the set of vixible cores. A `CoreIndex` should only ever be
+/// created with a value known to be less than the number of visible cores.
+pub struct CoreIndex(usize);
+
+pub fn checkout_core_group() -> Option<MutexGuard<'static, CoreGroup>> {
+    match &*CORE_GROUPS {
+        Some(groups) => {
+            for (i, group) in groups.iter().enumerate() {
+                match group.try_lock() {
+                    Ok(guard) => {
+                        debug!("checked out core group {}", i);
+                        return Some(guard);
+                    }
+                    Err(_) => debug!("core group {} locked, could not checkout", i),
+                }
+            }
+            None
+        }
+        None => None,
+    }
+}
+
+#[cfg(not(target_os = "windows"))]
+pub type ThreadId = libc::pthread_t;
+
+#[cfg(target_os = "windows")]
+pub type ThreadId = winapi::winnt::HANDLE;
+
+/// Helper method to get the thread id through libc, with current rust stable (1.5.0) its not
+/// possible otherwise I think.
+#[cfg(not(target_os = "windows"))]
+fn get_thread_id() -> ThreadId {
+    unsafe { libc::pthread_self() }
+}
+
+#[cfg(target_os = "windows")]
+fn get_thread_id() -> ThreadId {
+    unsafe { kernel32::GetCurrentThread() }
+}
+
+pub struct Cleanup {
+    tid: ThreadId,
+    prior_state: Option<hwloc::Bitmap>,
+}
+
+impl Drop for Cleanup {
+    fn drop(&mut self) {
+        match self.prior_state.take() {
+            Some(prior) => {
+                let child_topo = &TOPOLOGY;
+                let mut locked_topo = child_topo.lock().unwrap();
+                let _ = locked_topo.set_cpubind_for_thread(self.tid, prior, CPUBIND_THREAD);
+            }
+            None => (),
+        }
+    }
+}
+
+pub fn bind_core(core_index: CoreIndex) -> Result<Cleanup> {
+    let child_topo = &TOPOLOGY;
+    let tid = get_thread_id();
+    let mut locked_topo = child_topo.lock().unwrap();
+    let core = get_core_by_index(&locked_topo, core_index).map_err(|err| {
+        anyhow::format_err!("failed to get core at index {}: {:?}", core_index.0, err)
+    })?;
+
+    let cpuset = core.allowed_cpuset().ok_or_else(|| {
+        anyhow::format_err!("no allowed cpuset for core at index {}", core_index.0,)
+    })?;
+    debug!("allowed cpuset: {:?}", cpuset);
+    let mut bind_to = cpuset;
+
+    // Get only one logical processor (in case the core is SMT/hyper-threaded).
+    bind_to.singlify();
+
+    // Thread binding before explicit set.
+    let before = locked_topo.get_cpubind_for_thread(tid, CPUBIND_THREAD);
+
+    debug!("binding to {:?}", bind_to);
+    // Set the binding.
+    let result = locked_topo
+        .set_cpubind_for_thread(tid, bind_to, CPUBIND_THREAD)
+        .map_err(|err| anyhow::format_err!("failed to bind CPU: {:?}", err));
+
+    if result.is_err() {
+        warn!("error in bind_core, {:?}", result);
+    }
+
+    Ok(Cleanup {
+        tid,
+        prior_state: before,
+    })
+}
+
+fn get_core_by_index<'a>(topo: &'a Topology, index: CoreIndex) -> Result<&'a TopologyObject> {
+    let idx = index.0;
+
+    match topo.objects_with_type(&ObjectType::Core) {
+        Ok(all_cores) if idx < all_cores.len() => Ok(all_cores[idx]),
+        Ok(all_cores) => Err(anyhow::format_err!(
+            "idx ({}) out of range for {} cores",
+            idx,
+            all_cores.len()
+        )),
+        _e => Err(anyhow::format_err!("failed to get core by index {}", idx,)),
+    }
+}
+
+fn core_groups(cores_per_unit: usize) -> Option<Vec<Mutex<Vec<CoreIndex>>>> {
+    let topo = TOPOLOGY.lock().unwrap();
+
+    let core_depth = match topo.depth_or_below_for_type(&ObjectType::Core) {
+        Ok(depth) => depth,
+        Err(_) => return None,
+    };
+    let all_cores = topo.objects_with_type(&ObjectType::Core).unwrap();
+    let core_count = all_cores.len();
+
+    let mut cache_depth = core_depth;
+    let mut cache_count = 0;
+
+    while cache_depth > 0 {
+        let objs = topo.objects_at_depth(cache_depth);
+        let obj_count = objs.len();
+        if obj_count < core_count {
+            cache_count = obj_count;
+            break;
+        }
+
+        cache_depth -= 1;
+    }
+
+    assert_eq!(0, core_count % cache_count);
+    let mut group_size = core_count / cache_count;
+    let mut group_count = cache_count;
+
+    if cache_count <= 1 {
+        // If there are not more than one shared caches, there is no benefit in trying to group cores by cache.
+        // In that case, prefer more groups so we can still bind cores and also get some parallelism.
+        // Create as many full groups as possible. The last group may not be full.
+        group_count = core_count / cores_per_unit;
+        group_size = cores_per_unit;
+
+        info!(
+            "found only {} shared cache(s), heuristically grouping cores into {} groups",
+            cache_count, group_count
+        );
+    } else {
+        debug!(
+            "Cores: {}, Shared Caches: {}, cores per cache (group_size): {}",
+            core_count, cache_count, group_size
+        );
+    }
+
+    let core_groups = (0..group_count)
+        .map(|i| {
+            (0..group_size)
+                .map(|j| {
+                    let core_index = i * group_size + j;
+                    assert!(core_index < core_count);
+                    CoreIndex(core_index)
+                })
+                .collect::<Vec<_>>()
+        })
+        .collect::<Vec<_>>();
+
+    Some(
+        core_groups
+            .iter()
+            .map(|group| Mutex::new(group.clone()))
+            .collect::<Vec<_>>(),
+    )
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_cores() {
+        core_groups(2);
+    }
+
+    #[test]
+    fn test_checkout_cores() {
+        let checkout1 = checkout_core_group();
+        dbg!(&checkout1);
+        let checkout2 = checkout_core_group();
+        dbg!(&checkout2);
+
+        // This test might fail if run on a machine with fewer than four cores.
+        match (checkout1, checkout2) {
+            (Some(c1), Some(c2)) => assert!(*c1 != *c2),
+            _ => panic!("failed to get two checkouts"),
+        }
+    }
+}
--- a/storage-proofs/porep/src/stacked/vanilla/create_label/multi.rs
+++ b/storage-proofs/porep/src/stacked/vanilla/create_label/multi.rs
@@ -2,6 +2,7 @@ use std::convert::TryInto;
 use std::marker::PhantomData;
 use std::mem::size_of;
 use std::sync::atomic::{AtomicU64, Ordering::SeqCst};
+use std::sync::{Arc, MutexGuard};
 
 use anyhow::{Context, Result};
 use byte_slice_cast::*;
@@ -24,6 +25,7 @@ use storage_proofs_core::{
 
 use super::super::{
     cache::ParentCache,
+    cores::{bind_core, checkout_core_group, CoreIndex},
     graph::{StackedBucketGraph, DEGREE, EXP_DEGREE},
     memory_handling::{setup_create_label_memory, CacheReader},
     params::{Labels, LabelsCache},
@@ -31,7 +33,10 @@ use super::super::{
     utils::*,
 };
 
+const MIN_BASE_PARENT_NODE: u64 = 2000;
+
 const NODE_WORDS: usize = NODE_SIZE / size_of::<u32>();
+const SHA_BLOCK_SIZE: usize = 64;
 
 const SHA256_INITIAL_DIGEST: [u32; 8] = [
     0x6a09_e667,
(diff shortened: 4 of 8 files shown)


.circleci/config.yml
```diff
@@ -34,6 +34,8 @@ jobs:
       - run: cargo update
       - run: cargo fetch
       - run: rustc +$(cat rust-toolchain) --version
+      - run: apt-get update
+      - run: apt-get install -y hwloc libhwloc-dev
```
dignifiedquire, line 38: can likely be removed now, `libhwloc-dev` is no in the main image

storage-proofs/porep/Cargo.toml
```diff
@@ -9,34 +9,36 @@ repository = "https://github.com/filecoin-project/rust-fil-proofs"
 readme = "README.md"
 
 [dependencies]
+anyhow = "1.0.23"
```
dignifiedquire, line 12: can you please not reorder the deps, this is gonna be painful when I have to rebase some PRs onto this

storage-proofs/porep/src/stacked/vanilla/cores.rs
```diff
@@ -0,0 +1,189 @@
...
+        let settings = settings::SETTINGS.lock().expect("settings lock failure");
+        let num_producers = settings.multicore_sdr_producers;
+        let cores_per_unit = num_producers + 1;
+
+        core_groups(cores_per_unit)
+    };
+}
+
+#[derive(Clone, Copy, Debug, PartialEq)]
+pub struct CoreIndex(usize);
```
dignifiedquire, line 23: I can haz docstring?

storage-proofs/porep/src/stacked/vanilla/cores.rs
```diff
@@ -0,0 +1,189 @@
...
+
+use anyhow::Result;
+use hwloc::{ObjectType, Topology, TopologyObject, CPUBIND_THREAD};
+use lazy_static::lazy_static;
+
+use storage_proofs_core::settings;
+
+type CoreGroup = Vec<CoreIndex>;
+lazy_static! {
+    pub static ref TOPOLOGY: Arc<Mutex<Topology>> = Arc::new(Mutex::new(Topology::new()));
```
dignifiedquire, line 12: why is this `Arc`, `Mutex`? it seems to me this is never changed, so just `Topology` should be enough

storage-proofs/porep/src/stacked/vanilla/cores.rs
```diff
@@ -0,0 +1,189 @@
...
+            "idx ({}) out of range for {} cores",
+            idx,
+            all_cores.len()
+        )),
+        _e => Err(anyhow::format_err!("failed to get core by index {}", idx,)),
+    }
+}
+
+/// Helper method to get the thread id through libc, with current rust stable (1.5.0) its not
+/// possible otherwise I think.
```
dignifiedquire, line 88: it should be

```rust
let current_thread = std::thread::current();
let id = current_thread.id();
```

storage-proofs/porep/src/stacked/vanilla/cores.rs
```diff
@@ -0,0 +1,189 @@
...
+    let topo = TOPOLOGY.lock().unwrap();
+
+    let core_depth = match topo.depth_or_below_for_type(&ObjectType::Core) {
+        Ok(depth) => depth,
+        Err(_) => return None,
+    };
+    let all_cores = topo.objects_with_type(&ObjectType::Core).unwrap();
+    let core_count = all_cores.len();
+
+    all_cores[0].allowed_cpuset().unwrap();
```
dignifiedquire, line 109: I am not sure, but this might happen in the wild that `all_cores` is empty?

.circleci/config.yml
vmx: Yes, those lines should be removed again.

vmx: I removed those as I touched the config to make macOS work anyway.

vmx:
@porcuquine The CI is setup properly, hwloc is installed on Linux as well as macOS. The current CI failure seem to be actual failures.

storage-proofs/porep/src/stacked/vanilla/memory_handling.rs
```diff
@@ -8,21 +8,76 @@ use anyhow::Result;
...
+    }
+    fn store(&self, val: usize) {
+        self.cur.store(val, SeqCst);
+        self.cur_safe.store(val, SeqCst);
+    }
+    fn compare_and_swap(&self, before: usize, after: usize) {
+        self.cur.compare_and_swap(before, after, SeqCst);
+        self.cur_safe.compare_and_swap(before, after, SeqCst);
+    }
+    fn increment(&self, target: usize, wait_fn: &dyn Fn() -> bool, advance_fn: &dyn Fn()) {
```
dignifiedquire, line 49: if change the signature to 

```rust
fn increment<F: Fn(), G: Fn()>(&self, target: usize, wait_fn: F, advance_fn: G)
```

you avoid dynamic dispatch + boxing

storage-proofs/porep/Cargo.toml
porcuquine: Fine, but let's reorder them at some point in a single focused PR, then keep them that way.

storage-proofs/porep/src/stacked/vanilla/cores.rs
porcuquine: The `set_cpubind_for_thread` call requires it to be mutable. I can get rid of the `Arc`, though. It was needed at one point along the way but not was things ended up.

storage-proofs/porep/src/stacked/vanilla/cores.rs
porcuquine: FYI, that comments comes from the example code I borrowed. Your snippet doesn't work (wrong) type.

storage-proofs/porep/src/stacked/vanilla/cores.rs
porcuquine: Fair point. This would violate my expectation of how things work, but you're right that it's possible from a types/code perspective.

porcuquine: Anyway, this was just a stray line to test callability of`allowed_cpuset()`, so will remove. Good eye.

storage-proofs/porep/src/stacked/vanilla/cores.rs
porcuquine: Haz now.

storage-proofs/porep/src/stacked/vanilla/create_label/multi.rs
```diff
@@ -205,17 +211,21 @@ fn create_layer_labels(
...
+
+    // NOTE: Stride must not exceed `sdr_parents_cache_window_nodes`.
+    // If it does, the process will deadlock with producers and consumers
+    // waiting for each other.
+    let cache_window_nodes = parents_cache.size() / (4 * DEGREE);
+    let producer_stride = settings
+        .multicore_sdr_producer_stride
+        .min(cache_window_nodes as u64);
 
     const BYTES_PER_NODE: usize = (NODE_SIZE * DEGREE) + 64;
```
cryptonemo, line 230: Comment to describe what this `64` and the above `4` are representing?

storage-proofs/porep/src/stacked/vanilla/memory_handling.rs
```diff
@@ -77,7 +148,8 @@ impl<T: FromByteSlice> CacheReader<T> {
     // This allows splitting the reset in order to avoid a pause.
     pub fn reset(&self) -> Result<()> {
         self.start_reset()?;
-        self.finish_reset()
+        self.finish_reset()?;
+        Ok(())
```
cryptonemo, line 152: Just curious why you split this?

porcuquine: It lets the first remapping happen while the previous layer is written to disk. In theory, it can happen sooner (while the previous layer is still being finished). That requires a bit more bookkeeping to get right, though — so the current implementation is enough for now.

cryptonemo: I just meant the return of the finish result, compared to swallowing the result and explicitly returning Ok (the before/after seem comparable), just more verbose now.  Maybe I'm missing something.

storage-proofs/porep/Cargo.toml
dignifiedquire: fair enough

storage-proofs/porep/src/stacked/vanilla/cores.rs
dignifiedquire: it likely is the case that hwloc isn't able to deal with rusts `ThreadId` :/

storage-proofs/porep/src/stacked/vanilla/memory_handling.rs
porcuquine: Oh, good point. I didn't actually do that on purpose. It's probably a casualty of the interactive-rebase/fix-errors slaughterhouse. I'll change it back.

dignifiedquire (dismissed):

Dizans:
Just curious, how much can this optimize accelerate the PreCommitPhase1 on AMD 3970X.

dignifiedquire (dismissed):

porcuquine:
@Dizans I put the latest benchmark in the description. It is almost down to two hours on 3970x.

storage-proofs/porep/src/stacked/vanilla/cores.rs
```diff
@@ -42,7 +42,25 @@ pub fn checkout_core_group() -> Option<MutexGuard<'static, CoreGroup>> {
...
+pub struct Cleanup {
+    prior_state: Option<hwloc::Bitmap>,
+}
+
+impl Drop for Cleanup {
+    fn drop(&mut self) {
+        match self.prior_state.take() {
+            Some(prior) => {
+                let child_topo = &TOPOLOGY;
+                let tid = get_thread_id();
```
dignifiedquire, line 54: maybe store the thread id too, to be sure this does it for the correct thread?

storage-proofs/porep/src/stacked/vanilla/create_label/multi.rs
```diff
@@ -439,14 +439,13 @@ pub fn create_labels_for_encoding<Tree: 'static + MerkleTreeTrait, T: AsRef<[u8]
 
     let core_group = Arc::new(checkout_core_group());
 
-    if let Some(group) = &*core_group {
+    // When `_cleanup_handle` is dropped, the previous binding of thread will be restored.
+    let _cleanup_handle = (*core_group).as_ref().map(|group| {
```
dignifiedquire, line 443: you need this in the `decoding` one too

porcuquine: Thanks, will add. It would be really nice if we could unify those paths, exactly for reasons like this. I'll look into it later.

storage-proofs/porep/src/stacked/vanilla/cores.rs
porcuquine: Done, thanks.

storage-proofs/porep/src/stacked/vanilla/cores.rs
```diff
@@ -43,6 +43,7 @@ pub fn checkout_core_group() -> Option<MutexGuard<'static, CoreGroup>> {
 }
 
 pub struct Cleanup {
+    tid: libc::pthread_t,
```
dignifiedquire, line 46: this will fail on windows

porcuquine: What's your point? Just kidding.

storage-proofs/porep/src/stacked/vanilla/create_label/multi.rs
porcuquine: I converted all the magic numbers to self-describing constants.

storage-proofs/porep/src/stacked/vanilla/cores.rs
porcuquine: Fixed.

dignifiedquire (dismissed):

dignifiedquire (approved):

dignifiedquire (approved):

cryptonemo (approved):

xueyangl:
@porcuquine i am trying to run lotus-bench 1.5.2 on Intel server platform with 2 Icelake cpu , i run 14 task in parallel, at the very begin of replication, your changes can work well, there are 7 tasks bind into cpu0 and 7 tasks bind into cpu1, however after running some time but still in replication, i saw the binding rule is changed,  some tasks(maybe in layer 6 or layer 7) bind to cpu0 originally  are re-bind to cpu1, do you have any idea about this symptom ?

porcuquine:
I don't have any insight into why that would be, but I'll be curious to hear as you or others learn more. I think @dignifiedquire may soon have hardware on which he could experiment.

xueyangl:
@porcuquine  so glad to hear your feedback, just want to know why you didn't consider HWLOC_CPUBIND_STRICT when trying to bind CPU instead of using CPUBIND_THREAD . actually , we are trying  CPUBIND_STRICT on our symptom to see if there is any improvement.

porcuquine:
I don't remember thinking about it either way. The current implementation worked well on my test hardware, which is most of what I was considering at the time.

xueyangl:
So glad to hear your feedback, actually I am working for intel , and we have Icelake products right now which can support SHA256.

We test both intel platform and AMD platform, both have this kind of issues, I am little confused what’s different between us,

my platform configuration :

Platform ：Intel Whitley Server Platform
CPU : Intel Icelake CPU   x 2
Memory :2T, each CPU has 1T


Regarding to CPU binding code, I still have some questions, could you please help clarify ?
Let’s have an example of one task with multi_core_sdr enabled.

Consumer thread binding to CPU core at the early beginning  --  my understanding is consumer thread should be fixed on one cpu core during whole progress, should not migrate to other core .
Producer thread :  on each layer, producer thread will do cpu core binding again , so it means for different layer, producer thread may be bind onto different CPU core.

Is my understanding correct ?

If my understanding is correct, so how can we guarantee consumer thread and producer threads are in same socket ? is it possible consumer thread bind onto CPU0 core and related producer threads bind onto CPU1 cores?

Thanks
Derek

porcuquine:
Out of curiosity, how many cores share an L2 cache on this Ice Lake CPU? You will need to make sure you don't try to use more processes than this number (set appropriate environment variables) for sealing, since shared cache is important to this optimization.

>Consumer thread binding to CPU core at the early beginning  --  my understanding is consumer thread should be fixed on one cpu core during whole progress, should not migrate to other core .

That should be true.

> Producer thread :  on each layer, producer thread will do cpu core binding again , so it means for different layer, producer thread may be bind onto different CPU core.

Although the binding is released between layers, the same one should be reacquired for the next layer. Maybe some other thread is getting scheduled there in between, and this reacquisition is failing. The code notes:

```
                // This could fail, but we will ignore the error if so.
                // It will be logged as a warning by `bind_core`.
```

Check your logs for evidence (or lack thereof) that this may be happening. Make sure to set log level appropriately.

>If my understanding is correct, so how can we guarantee consumer thread and producer threads are in same socket ? is it possible consumer thread bind onto CPU0 core and related producer threads bind onto CPU1 cores?

You might want to also log at DEBUG level and check the output from `cores.rs`. This might help you understand what is happening. If you compare that output against what you know about the topology, you might be able to confirm or deny that things are being detected and chosen as expected.

xueyangl:
Here is the hwloc information I got from my dual sockets system,  on Intel ICX CPU(should be all Intel CPU), L2 is specific to core , not shared, each core has specific 1.28M L2,
L3 is shared between all cores. I am getting the log with debug level= Trace and will do further analysis later.
Appreciate for your help, please share me any insights if you have .

*** Objects at level 0
0: Machine ()
*** Objects at level 1
0: Package ()
1: Package ()
*** Objects at level 2
0: L3 (48MB)
1: L3 (48MB)
*** Objects at level 3
0: L2 (1280KB)
1: L2 (1280KB)
2: L2 (1280KB)
3: L2 (1280KB)
4: L2 (1280KB)
5: L2 (1280KB)
6: L2 (1280KB)
7: L2 (1280KB)
8: L2 (1280KB)
9: L2 (1280KB)
10: L2 (1280KB)
11: L2 (1280KB)
12: L2 (1280KB)
13: L2 (1280KB)
14: L2 (1280KB)
15: L2 (1280KB)
16: L2 (1280KB)
17: L2 (1280KB)
18: L2 (1280KB)
19: L2 (1280KB)
20: L2 (1280KB)
21: L2 (1280KB)
22: L2 (1280KB)
23: L2 (1280KB)
24: L2 (1280KB)
25: L2 (1280KB)
26: L2 (1280KB)
27: L2 (1280KB)
28: L2 (1280KB)
29: L2 (1280KB)
30: L2 (1280KB)
31: L2 (1280KB)
32: L2 (1280KB)
33: L2 (1280KB)
34: L2 (1280KB)
35: L2 (1280KB)
36: L2 (1280KB)
37: L2 (1280KB)
38: L2 (1280KB)
39: L2 (1280KB)
40: L2 (1280KB)
41: L2 (1280KB)
42: L2 (1280KB)
43: L2 (1280KB)
44: L2 (1280KB)
45: L2 (1280KB)
46: L2 (1280KB)
47: L2 (1280KB)
48: L2 (1280KB)
49: L2 (1280KB)
50: L2 (1280KB)
51: L2 (1280KB)
52: L2 (1280KB)
53: L2 (1280KB)
54: L2 (1280KB)
55: L2 (1280KB)
56: L2 (1280KB)
57: L2 (1280KB)
58: L2 (1280KB)
59: L2 (1280KB)
60: L2 (1280KB)
61: L2 (1280KB)
62: L2 (1280KB)
63: L2 (1280KB)

porcuquine:
Right, sorry, I should have said L3 (fixed in my original message now).

If the L3 cache is shared by 32 cores, then it seems quite likely this optimization will not help much — or not nearly as much as is ideal. It might help some, so it's still trying to make sure you can get the producer threads stably pinned to a single core. But I don't think you will be able to get the very fast inter-core communication (via cache) which makes this particular consumer-producer pattern fast enough to let the producers keep up with the hashing of the consumer.

The idea is that we want the hashing to be the bottleneck, which means we need to be able to feed the cache fast enough. For this to work, the cache needs to be fast — and it needs to stay undisturbed by other work loads. The buffer shared between the cooperating cores (running the producers and consumer threads) needs to stay in that fast cache. This works out well when we can ensure no other work happens on the cores sharing an L3 cache. I am not sure it will work out well with your topology.

xueyangl:
We properly have fixed this NUMA issue on intel cpu, and I am wondering if it can be workaround commited into community to make sure end user can benefit it.

I am trying to find some way to identify CPU vendor info , and add some code to check if it is intel cpu this workaround will be executed,  make sure no impact to your previous code .  can hwloc retrieve cpu information ? vendor or CPUID .

porcuquine:
I'm not sure hwloc can, but you might try [cpuid](https://docs.rs/cpuid/0.1.1/cpuid/).

xueyangl:
Yes, appreciate for your great help , we are doing performance measurement now , will sync up with you later .

xueyangl:
Sorry for trouble you again, I have another problem that we frequently saw some PRODUCER NOT READY message when I enable Trace log,

Just like below , do you have any idea how to fix it ? or any insight to guide us to improve it ?

2021-04-17T13:20:50.443 DEBUG storage_proofs_porep::stacked::vanilla::create_label::multi > PRODUCER NOT READY! 978743553 2021-04-17T13:20:50.444 DEBUG storage_proofs_porep::stacked::vanilla::create_label::multi > PRODUCER NOT READY! 978744449 2021-04-17T13:20:50.445 DEBUG storage_proofs_porep::stacked::vanilla::create_label::multi > PRODUCER NOT READY! 978745217 2021-04-17T13:20:50.446 DEBUG storage_proofs_porep::stacked::vanilla::create_label::multi > PRODUCER NOT READY! 978746369 2021-04-17T13:20:50.447 DEBUG storage_proofs_porep::stacked::vanilla::create_label::multi > PRODUCER NOT READY! 978747265 2021-04-17T13:20:50.448 DEBUG storage_proofs_porep::stacked::vanilla::create_label::multi > PRODUCER NOT READY! 978748161 2021-04-17T13:20:50.449 DEBUG storage_proofs_porep::stacked::vanilla::create_label::multi > PRODUCER NOT READY! 978748929


From: Lun, Derek
Sent: Friday, April 16, 2021 3:57 PM
To: filecoin-project/rust-fil-proofs ***@***.***>; filecoin-project/rust-fil-proofs ***@***.***>
Cc: Comment ***@***.***>
Subject: RE: [filecoin-project/rust-fil-proofs] bind replication threads to specific cores (#1305)

Yes, appreciate for your great help , we are doing performance measurement now , will sync up with you later .

porcuquine:
This means that the main thread is waiting for data from a producer. If it doesn't happen 'too much', it's probably fine and is (I think) pretty normal. But if you are always waiting then that is a bottleneck. The obvious thing to try is to increase the number of producers, which you can set with `FIL_PROOFS_MULTICORE_SDR_PRODUCERS`. For this to be useful, though, the producers and the consumer need to share a cache (and no other threads should be disrupting that cache during the sealing process).

If you can't increase number of producers, you need to make sure the producers are producing fast enough. They read from disk, so I guess making sure you are using SSD would help. There are many _possible_ reasons, some of which might not be fixable, and it might not matter… Hope that helps.

PR #1309, opened 2020-10-07
 test: add test for resumable sealing

vmx:
There is a short running version, which doesn't run the proofs and
a longer running one, which does run the proofs.

Currently tests fail when the proofs are running. I don't know why,
perhaps I'm removing files that are needed/not keeping files around
that should be kept.

It can easily be reproduced locally via:

    RUST_LOG=debug cargo test test_resumable_seal -- --ignored --nocapture

@cryptonemo Can you please have a look at the test itself, if it does what
you expect and then please also check why it is failing?

### DIFF
--- a/filecoin-proofs/tests/api.rs
+++ b/filecoin-proofs/tests/api.rs
@@ -410,14 +410,7 @@ fn window_post<Tree: 'static + MerkleTreeTrait>(
     Ok(())
 }
 
-fn create_seal<R: Rng, Tree: 'static + MerkleTreeTrait>(
-    rng: &mut R,
-    sector_size: u64,
-    prover_id: ProverId,
-    skip_proof: bool,
-) -> Result<(SectorId, NamedTempFile, Commitment, tempfile::TempDir)> {
-    init_logger();
-
+fn generate_piece_file(sector_size: u64) -> Result<(NamedTempFile, Vec<u8>)> {
     let number_of_bytes_in_piece = UnpaddedBytesAmount::from(PaddedBytesAmount(sector_size));
 
     let piece_bytes: Vec<u8> = (0..number_of_bytes_in_piece.0)
@@ -428,6 +421,35 @@ fn create_seal<R: Rng, Tree: 'static + MerkleTreeTrait>(
     piece_file.write_all(&piece_bytes)?;
     piece_file.as_file_mut().sync_all()?;
     piece_file.as_file_mut().seek(SeekFrom::Start(0))?;
+
+    Ok((piece_file, piece_bytes))
+}
+
+fn porep_config(sector_size: u64, porep_id: [u8; 32]) -> PoRepConfig {
+    PoRepConfig {
+        sector_size: SectorSize(sector_size),
+        partitions: PoRepProofPartitions(
+            *POREP_PARTITIONS
+                .read()
+                .expect("POREP_PARTITIONS poisoned")
+                .get(&sector_size)
+                .expect("unknown sector size"),
+        ),
+        porep_id,
+    }
+}
+
+fn run_seal_pre_commit_phase1<Tree: 'static + MerkleTreeTrait>(
+    config: PoRepConfig,
+    prover_id: ProverId,
+    sector_id: SectorId,
+    ticket: [u8; 32],
+    cache_dir: &tempfile::TempDir,
+    mut piece_file: &mut NamedTempFile,
+    sealed_sector_file: &NamedTempFile,
+) -> Result<(Vec<PieceInfo>, SealPreCommitPhase1Output<Tree>)> {
+    let number_of_bytes_in_piece =
+        UnpaddedBytesAmount::from(PaddedBytesAmount(config.sector_size.into()));
 
     let piece_info = generate_piece_commitment(piece_file.as_file_mut(), number_of_bytes_in_piece)?;
     piece_file.as_file_mut().seek(SeekFrom::Start(0))?;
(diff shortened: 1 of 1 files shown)


vmx:
Ready for review. It's easiest to be reviewed as individual commits, as most part of it is a pure refactor without functional changes.

filecoin-proofs/tests/api.rs
```diff
@@ -479,6 +481,38 @@ fn create_seal<R: Rng, Tree: 'static + MerkleTreeTrait>(
...
+    skip_proof: bool,
+) -> Result<(SectorId, NamedTempFile, Commitment, tempfile::TempDir)> {
+    init_logger();
+
+    let (mut piece_file, piece_bytes) = generate_piece_file(sector_size)?;
+    let sealed_sector_file = NamedTempFile::new()?;
+    let mut unseal_file = NamedTempFile::new()?;
+    let cache_dir = tempfile::tempdir().expect("failed to create temp dir");
+
+    let arbitrary_porep_id = rng.gen();
```
porcuquine, line 727: If you randomly generate this, then you won't be able to reuse the parents cache. So the price of regenerating the cache will be paid on every run, and more and more disk will be consumed. I think this should be a hardcoded constant for that reason.

filecoin-proofs/tests/api.rs
```diff
@@ -484,6 +485,82 @@ fn run_seal_pre_commit_phase1<Tree: 'static + MerkleTreeTrait>(
     Ok((piece_infos, phase1_output))
 }
 
+fn run_proof<Tree: 'static + MerkleTreeTrait>(
+    config: PoRepConfig,
```
porcuquine, line 640: As a refactor just to decompose the work more, this seems fine, but the name `run_proof` is misleading, since this is also bundling unsealing, which is definitely not part of the proof.

porcuquine: It would probably be better to pull the unsealing part out into its own function.

filecoin-proofs/tests/api.rs
```diff
@@ -79,6 +80,156 @@ fn seal_lifecycle<Tree: 'static + MerkleTreeTrait>(sector_size: u64) -> Result<(
...
+    )
+    .expect("failed to run seal pre commit phase1");
+    let layers = get_layer_file_paths(&cache_dir);
+    assert_eq!(layers.len(), 2, "not all expected layers were created");
+
+    // Delete one layer, keep the other
+    clear_cache_dir_keep_data_layer(&cache_dir);
+    std::fs::remove_file(&layers[layer_to_delete]).expect("failed to remove layer");
+    let layers_removed = get_layer_file_paths(&cache_dir);
+    assert_eq!(layers_removed.len(), 1, "found more than one layer");
```
porcuquine, line 171: This name is somewhat confusing. As written, I assume `layers_removed` would contain the layers which were removed, not the layers *not* removed (i.e. those remaining).

The assertion doesn't clarify it for me, either, since with only two layers hardcoded as the test size, either the removed or remaining layers should have length 1.

Also, I'm skeptical of the assertion, since it could also fail if no layers were found, for whatever reason.

vmx: > Also, I'm skeptical of the assertion, since it could also fail if no layers were found, for whatever reason.

I think it should fail if no layers are found as we are testing the resuming with existing layers. I'll reword the assertion message though.

porcuquine (commented):
I think this works, but a less hyper-specific test (assuming exactly two total layers, using if to select from the only two options) would be a bit more versatile when considered against the range of behaviors the resumable seal feature might have. That said, we can update the tests as needed if behavior ever changes. I do think this does exercise what was implemented, though.

What I'm a little more concerned about is that this does not actually verify that resumable sealing does less work than the initial sealing, just that it doesn't fail or give wrong results. In that way, I'm not sure this test distinguishes between the observable behavior before and after implementation of the resumable sealing feature. Ideally, that's what this test would verify, though.

I think there may be some internal return values that track which layers were actually generated. Maybe those can be exploited, even if it means a slight change to the internal functions to make them more effectively testable.

vmx:
@porcuquine it sounds like you're after a unit-test like thing along the lines of https://github.com/filecoin-project/rust-fil-proofs/blob/master/storage-proofs/porep/src/stacked/vanilla/proof.rs#L1567-L1692

porcuquine:
@vmx I don't think it needs to be so thorough — just somehow capturing the central point of the feature (that resumable sealing is actually 'resuming') seems important. I'm not sure what mechanism would allow that minimally, but I think it can be 'minimal' if you see a way to make it be.

vmx:
> just somehow capturing the central point of the feature (that resumable sealing is actually 'resuming') seems important. I'm not sure what mechanism would allow that minimally, but I think it can be 'minimal' if you see a way to make it be.

To me the most important thing for this test is that resuming doesn't break anything.

You can look at the logs if it does actually resume. I understand the point of having this automated, but I don't think it's worth the time/making the code more complicated. In case the resume doesn't work (it pretends to resume, but actually does a full rebuild) it would be bad performance wise, but things would still work. If we break this accidentally I expect that someone would report. Please also note that these tests only work on 2 layers as this is what the smaller sized parameters are doing.

I would rather have performance tests on the actual parameters that run on a daily basis to catch regressions on the overall performance. One of those tests could also be one that resumes the sealing.

To conclude: I would just merge this PR and don't bother too much about the resumable sealing.

porcuquine (approved):
Alright, @vmx, you convinced me. Approved.

porcuquine:
Whoops, @vmx, looks like this need a rebase now. Ping me when done so we can merge.

vmx:
@porcuquine rebased as tests pass.

porcuquine (approved):

Issue #1320, opened 2020-10-18
Fast PC1 branch seems to go belly up for me


karalabe:
I.e. https://app.slack.com/client/TEHTVS1L6/C0179RNEMU4/thread/C0179RNEMU4-1603007653.135600

My hardware setup is: EPYC 7402P, 512GB RAM, PCIe x4 NVMe SSD, Nvidia 1070. Ubuntu 20.04

I've switched to `feat/proofs-5.2.3` in lotus to try out the faster PC1 but it failed in a lot of different ways for me on PC2.

Originally I ran with `FIL_PROOFS_USE_MULTICORE_SDR=1` but seen super weird behavior so I dropped this env var altogether to keep the change surface a bit smaller. My worker configs were:

```
Environment=FIL_PROOFS_MAXIMIZE_CACHING=1
Environment=FIL_PROOFS_USE_GPU_COLUMN_BUILDER=1
Environment=FIL_PROOFS_USE_GPU_TREE_BUILDER=1
Environment=BELLMAN_NO_GPU=true
Environment=BELLMAN_CUSTOM_GPU="GeForce GTX 1070:1920"
Environment=FIL_PROOFS_MAX_GPU_COLUMN_BATCH_SIZE=360000
Environment=FIL_PROOFS_SDR_PARENTS_CACHE_SIZE=1073741824
Environment=TRUST_PARAMS=1
Environment=RUST_LOG=info
Environment=RUST_BACKTRACE=1
```

These configs seem to be stable on the 0.10.2 release branch and never fail.

---

One immediate error that appeared newly is `"msg":"Cannot get device list for platform: Clover!"`

The failure that choked all my PC2s were:

```
Oct 18 09:04:58 filecoin lotus-worker[6450]: thread '<unnamed>' panicked at 'failed to add columns: GPUError("FutharkError(FutharkError { error: \"./lib/a.c:21158: OpenCL call\\n  opencl_alloc(&ctx->opencl, size, desc, &block->mem)\\nfailed with error code -4 (Memory object allocation failure)\\n\" })")', /root/.cargo/registry/src/github.com-1ecc6299db9ec823/storage-proofs-porep-5.2.3/src/stacked/vanilla/proof.rs:548:71
Oct 18 09:04:58 filecoin lotus-worker[6450]: stack backtrace:
Oct 18 09:04:58 filecoin lotus-worker[6450]:    0: rust_begin_unwind
[...]

PR #1322, opened 2020-10-20
Downgrade neptune to v1.1.x.

porcuquine:
Some problems reported in #1320 appear to be caused by neptune v1.2.0 — which has not actually been incorporated in a `rust-fil-proofs` release yet. We should downgrade to `1.1.2` for now. I'm only specifying `1.1` here so we will get any further patches needed before moving to `2.x`. We may need new API/FFI releases or branches in order to get a revised version of https://github.com/filecoin-project/lotus/tree/feat/proofs-5.2.3 working. We should do that, though, so wider testing of multicore SDR can continue.


porcuquine:
I think this is defunct and will be resolved differently in the next release.

PR #1323, opened 2020-10-21
fix: nse window-post challenge derivation

DrPeterVanNostrand:
Changes window-post challenge derivation to be agnostic of sector ordering.

### DIFF
--- a/storage-proofs/post/src/nse_window/compound.rs
+++ b/storage-proofs/post/src/nse_window/compound.rs
@@ -59,25 +59,23 @@ impl<'a, Tree: 'static + MerkleTreeTrait>
             .nth(partition_index)
             .ok_or_else(|| anyhow!("invalid number of sectors/partition index"))?;
 
-        for (i, sector) in sectors.iter().enumerate() {
+        for sector in sectors.iter() {
             // 1. Inputs for verifying comm_r = H(comm_layer_0 || ..)
             inputs.push(sector.comm_r.into());
             inputs.extend(sector.comm_layers.iter().copied().map(Into::into));
 
             inputs.push(sector.comm_replica.into());
 
+            let mut challenge_index_in_sector: u64 = 0;
+
             // 2. Inputs for verifying inclusion paths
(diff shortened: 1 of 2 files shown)


porcuquine (approved):
This looks like a correct fix to me. I agree that challenges must not depend on the ordering of sectors. It would be worth @dignifiedquire taking a look just to confirm there's not something intended subtlety we're missing.

porcuquine:
I reran CI, but one job still seems to be failing. Looks like a timeout.

DrPeterVanNostrand:
@porcuquine I think the test timeout was inherited from from the parent branch

dignifiedquire (approved):

storage-proofs/post/src/nse_window/compound.rs
```diff
@@ -59,25 +59,23 @@ impl<'a, Tree: 'static + MerkleTreeTrait>
...
+            let mut challenge_index_in_sector: u64 = 0;
+
             // 2. Inputs for verifying inclusion paths
-            for window_index in 0..pub_params.num_windows() {
-                for n in 0..pub_params.window_challenge_count {
-                    let challenge_index = ((partition_index * num_sectors_per_chunk + i)
-                        * pub_params.window_challenge_count
-                        + n) as u64;
+            for _ in 0..pub_params.num_windows() {
+                for _ in 0..pub_params.window_challenge_count {
```
cryptonemo, line 73: Instead of `for _ and for _`, why not use `for window_index and for index` and use something like `let challenge_index = window_index * num_windows + index` that instead of manually incrementing?

DrPeterVanNostrand: Personal preference, I'd rather do one addition per inner-loop than a multiplication+addition. But, I can see how the multiplication+addition would be easier to read.

If you think that the nested loops are harder to read now, I'm fine with changing it.

cryptonemo: No problem,  don't block on style

PR #1328, opened 2020-10-26
fix: install hwloc to fix metrics capture on CI

cryptonemo:


cryptonemo:
Previous CI run passed on metrics-capture.  I put it back to only run on the master branch after that, so when CI passes, please approve @porcuquine or @dignifiedquire

dignifiedquire (approved):

Issue #1327, opened 2020-10-24
Aborted PC2 will always fail with bad descriptor until PC1 regenerated


karalabe:
I've seen this enough times to decide it's a bug.

Every now and again I need to restart my miner/worker (config changes, lotus update, etc). Since I don't want to wait for 12+ hours for all the seal tasks to finish, I usually just tell it to shut down gracefully.

When I start the workers back up, it picks up where it left off (obviously restarting the PC1 or PC2 or C2, can't continue). However, whilst restarted PC1 or C2 will succeed, restarted PC2 will always fail with "Bad File Descriptor"

```
Oct 24 11:00:38 filecoin lotus-worker[20925]: thread '<unnamed>' panicked at 'failed to flatten and write store: Bad file descriptor (os error 9)', /root/.cargo/registry/src/github.com>
Oct 24 11:00:38 filecoin lotus-worker[20925]: stack backtrace:
Oct 24 11:00:38 filecoin lotus-worker[20925]:    0: backtrace::backtrace::libunwind::trace
Oct 24 11:00:38 filecoin lotus-worker[20925]:              at ./cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-0.3.44/src/backtrace/libunwind.rs:86
Oct 24 11:00:38 filecoin lotus-worker[20925]:    1: backtrace::backtrace::trace_unsynchronized
Oct 24 11:00:38 filecoin lotus-worker[20925]:              at ./cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-0.3.44/src/backtrace/mod.rs:66
Oct 24 11:00:38 filecoin lotus-worker[20925]:    2: std::sys_common::backtrace::_print_fmt
Oct 24 11:00:38 filecoin lotus-worker[20925]:              at src/libstd/sys_common/backtrace.rs:78
Oct 24 11:00:38 filecoin lotus-worker[20925]:    3: <std::sys_common::backtrace::_print::DisplayBacktrace as core::fmt::Display>::fmt
Oct 24 11:00:38 filecoin lotus-worker[20925]:              at src/libstd/sys_common/backtrace.rs:59
Oct 24 11:00:38 filecoin lotus-worker[20925]:    4: core::fmt::write
Oct 24 11:00:38 filecoin lotus-worker[20925]:              at src/libcore/fmt/mod.rs:1063
Oct 24 11:00:38 filecoin lotus-worker[20925]:    5: std::io::Write::write_fmt
[...]

PR #1338, opened 2020-11-02
feat: raise soft fdlimit to max at runtime (os x/linux)

cryptonemo:
feat: add tool that shows max available open files

### DIFF
--- a/storage-proofs/porep/src/stacked/vanilla/proof.rs
+++ b/storage-proofs/porep/src/stacked/vanilla/proof.rs
@@ -997,8 +997,7 @@ impl<'a, Tree: 'static + MerkleTreeTrait, G: 'static + Hasher> StackedDrg<'a, Tr
             LabelsCache::<Tree>::new(&label_configs).context("failed to create labels cache")?;
         let configs = split_config(tree_c_config.clone(), tree_count)?;
 
-        let res = fdlimit::raise_fd_limit()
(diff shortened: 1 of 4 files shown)


cryptonemo:
I believe this is related to https://github.com/filecoin-project/rust-fil-proofs/issues/1327.  In that issue, an open file descriptor limit is hit write persisting the tree data in parallel batches (or at least that's the case that has been reproduced resulting in the same error reported).  This PR does 2 things: 1) It logs what the max available fd limit is, which if often mis-configured on users' systems, and 2) It explicitly raises the fd descriptor limit from the soft limit to the max.

storage-proofs/porep/src/stacked/vanilla/proof.rs
```diff
@@ -997,6 +997,9 @@ impl<'a, Tree: 'static + MerkleTreeTrait, G: 'static + Hasher> StackedDrg<'a, Tr
             LabelsCache::<Tree>::new(&label_configs).context("failed to create labels cache")?;
         let configs = split_config(tree_c_config.clone(), tree_count)?;
 
+        let res = fdlimit::raise_fd_limit().expect("Failed to raise the fd limit");
```
dignifiedquire, line 1000: This should not panic, but simply print a warning

dignifiedquire (approved):

PR #1358, opened 2020-11-10
ci: parallelize test runs accross packages

### LINKED ISSUE #1354: Improve CI speed
We can improve the CI speed by splitting up the test per crate and as such have more jobs be executed in parallel. 

Eg

```
job1: cargo test -p storage-proofs-core
job2: cargo test -p storage-proofs-porep
job3: cargo test -p storage-proofs-post
```

dignifiedquire:
Closes #1354

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,24 +1,36 @@
 version: 2.1
 
-jobs:
-  ensure_groth_parameters_and_keys_linux:
+parameters:
+  nightly-toolchain:
+    type: string
+    default: "nightly-2020-11-09"
+
+executors:
+  default:
     docker:
       - image: filecoin/rust:latest
     working_directory: /mnt/crate
     resource_class: 2xlarge+
+
+setup-env: &setup-env
+  FIL_PROOFS_PARAMETER_CACHE: "/root/filecoin-proof-parameters/"
+  RUST_LOG: info
+
+
+jobs:
+  ensure_groth_parameters_and_keys_linux:
+    executor: default
+    environment: *setup-env
     steps:
-      - configure_environment_variables
       - checkout
       - restore_parameter_cache
       - ensure_filecoin_parameters
       - save_parameter_cache
+
   cargo_fetch:
-    docker:
-      - image: filecoin/rust:latest
-    working_directory: /mnt/crate
-    resource_class: 2xlarge+
+    executor: default
+    environment: *setup-env
     steps:
-      - configure_environment_variables
       - checkout
       - run:
           name: Calculate dependencies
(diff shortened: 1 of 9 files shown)


.circleci/config.yml
```diff
@@ -132,26 +124,24 @@ jobs:
...
+            - cargo-v28-b-{{ checksum "rust-toolchain" }}-{{ checksum "Cargo.toml" }}-{{ checksum "Cargo.lock" }}-{{ arch }}
       - restore_parameter_cache
       - run:
           name: Test with use_multicore_sdr pairing enabled
           command: |
             ulimit -n 20000
             ulimit -u 20000
             ulimit -n 20000
-            cargo +nightly -Zpackage-features test --all --verbose --release --test api -- --ignored lifecycle
+            cargo +nightly-2020-11-09 -Zpackage-features test --all --verbose --release lifecycle -- --ignored --nocapture
```
vmx, line 143: Can be a followup PR: making the exactly nightly version (`nightly-2020-11-09`) a variable would be great. This would make updating to never versions easier (I already see myself wanting a newer nightly and then forgot to change it at all places).

storage-proofs/porep/src/stacked/vanilla/cache.rs
```diff
@@ -157,9 +157,21 @@ impl ParentCache {
     {
         let path = cache_path(cache_entries, graph);
         if path.exists() {
-            Self::open(len, cache_entries, graph, path)
+            Self::open(len, cache_entries, graph, &path)
         } else {
-            Self::generate(len, cache_entries, graph, path)
+            match Self::generate(len, cache_entries, graph, &path) {
+                Ok(c) => Ok(c),
```
vmx, line 163: I think `.or_else()` could be used. Though I'm not sure if that makes things better.

storage-proofs/post/Cargo.toml
```diff
@@ -34,6 +34,7 @@ tempfile = "3"
 pretty_assertions = "0.6.1"
 rand_xorshift = "0.2.0"
 filecoin-hashers = { path = "../../filecoin-hashers", version = "1.0.0", default-features = false, features = ["poseidon", "sha256", "blake2s"]}
+pretty_env_logger = "0.4.0"
```
vmx, line 37: I'd use `fil_logger` instead for consistency with other crates.

storage-proofs/core/src/parameter_cache.rs
```diff
@@ -305,12 +313,14 @@ where
 
         // generate (or load) verifying key
         let cache_path = ensure_ancestor_dirs_exist(parameter_cache_verifying_key_path(&id))?;
-        read_cached_verifying_key(&cache_path)
-            .or_else(|_| write_cached_verifying_key(&cache_path, generate()?))
+        match read_cached_verifying_key(&cache_path) {
```
vmx, line 316: I guess `or_else()` could be kept here? Any reasons to change, other than a preference for pattern matching?

storage-proofs/porep/src/stacked/vanilla/cache.rs
dignifiedquire: it made the code even harder to read

.circleci/config.yml
dignifiedquire: if you tell me how I can use a variable in the config, I would love to use one

storage-proofs/core/src/parameter_cache.rs
dignifiedquire: the types got really messy, because of two different Error types, and it got really hard to read

.circleci/config.yml
vmx: https://github.com/filecoin-project/bellperson/blob/929337fc19be417fc16d1aa3d3139d1e21fb6bdb/.circleci/config.yml#L3-L6 :)

storage-proofs/post/Cargo.toml
vmx: I forgot to add. If `fil_logger` doesn't cut it for some reason, I'd rather fix fil_logger. Having the same logger across a larger system really is beneficial (I've been there in the past, with different log structures from different sub-subsystems running within the same daemon).

dignifiedquire: Yeah it doesn't work for tests, because `fil_logger::init()` panics if it is called multiple times

vmx: Looks like I need to check what pretty_env_logger is doing there :)

vmx (approved):

vmx (approved):

Issue #1357, opened 2020-11-09
Re-factor graph fix / upgrade


cryptonemo:
Using a versioned API enum similar to https://github.com/filecoin-project/rust-fil-proofs/pull/1284, re-factor the existing switching logic.

cryptonemo:
Closed by https://github.com/filecoin-project/rust-fil-proofs/pull/1362

PR #1362, opened 2020-11-11
feat: add api version enum for determining runtime behaviour

cryptonemo:
feat: add tests paths for both v1_0 and v1_1 versions
feat: allow api_version to be used as cli argument
fix: renamed ApiVersion and use semver string parsing


cryptonemo:
Resolves #1357

storage-proofs/core/src/api_version.rs
```diff
@@ -0,0 +1,29 @@
+use std::str::FromStr;
+
+use anyhow::{self, Result};
+use semver::Version;
+
+#[derive(Debug, Copy, Clone, Eq, PartialEq)]
+pub enum ApiVersion {
```
dignifiedquire, line 7: lets add a manual implementation for `Debug` and `Display` that prints them as their semver version aka `1.1.0`

storage-proofs/core/src/api_version.rs
```diff
@@ -0,0 +1,29 @@
...
+    V1_1,
+}
+
+impl FromStr for ApiVersion {
+    type Err = anyhow::Error;
+    fn from_str(api_version_str: &str) -> Result<Self> {
+        let api_version = Version::parse(api_version_str)?;
+        match api_version.major {
+            1 => match api_version.minor {
+                0 => Ok(ApiVersion::V1_0),
```
dignifiedquire, line 18: this should check that `patch` is `0`

storage-proofs/core/src/lib.rs
```diff
@@ -37,19 +38,3 @@ pub(crate) const TEST_SEED: [u8; 16] = [
 pub const MAX_LEGACY_POREP_REGISTERED_PROOF_ID: u64 = 4;
 
 pub type PoRepID = [u8; 32];
-pub fn is_legacy_porep_id(porep_id: PoRepID) -> bool {
```
dignifiedquire, line 40: we should keep this around and validate that we have the correct porep id for now

storage-proofs/core/src/api_version.rs
cryptonemo: Ok, in that case, I think we should also more strongly consider full versioning (e.g. `ApiVersion::V1_0_0` instead of `ApiVersion::V1_0`)

storage-proofs/core/src/lib.rs
cryptonemo: I can add this back, but I don't think asserting this around too often is the best idea, as the behavior is fully decoupled from the porep_id at the moment, and that would explicitly tie it back to it.

storage-proofs/core/src/api_version.rs
dignifiedquire: fair enough

storage-proofs/core/src/lib.rs
dignifiedquire: I know, but we should have this check for these two API versions at the high level at least, maybe in filecoin-proofs, just to make sure we maintain the same invariants that we have right now.

storage-proofs/post/src/fallback/compound.rs
```diff
@@ -63,21 +64,25 @@ impl<'a, Tree: 'static + MerkleTreeTrait>
             // 1. Inputs for verifying comm_r = H(comm_c || comm_r_last)
             inputs.push(sector.comm_r.into());
 
+            // avoid rehashing fixed inputs
+            let mut challenge_hasher = Sha256::new();
+            challenge_hasher.update(AsRef::<[u8]>::as_ref(&pub_inputs.randomness));
+            challenge_hasher.update(&u64::from(sector.id).to_le_bytes()[..]);
+
```
porcuquine, line 71: What is this? Is there supposed to be some changed behavior? It would really make current and historical review of this 45-file changeset simpler if it were restricted to *only* changes associated with introduction of this versioning scheme. If this is a simple refactor, it would be much better to put it in a distinct PR.

storage-proofs/post/src/fallback/vanilla.rs
```diff
@@ -364,10 +365,9 @@ impl<'a, Tree: 'a + MerkleTreeTrait> ProofScheme<'a> for FallbackPoSt<'a, Tree>
             num_sectors_per_chunk,
         );
 
-        let mut partition_proofs = Vec::new();
-
         // Use `BTreeSet` so failure result will be canonically ordered (sorted).
         let mut faulty_sectors = BTreeSet::new();
+        let mut partition_proofs = Vec::new();
```
porcuquine, line 370: Similar comment to above — is there a significance to this reordering?

storage-proofs/core/src/drgraph.rs
```diff
@@ -157,10 +159,9 @@ impl<H: Hasher> Graph<H> for BucketGraph<H> {
                 let metagraph_node = node as u64 * m_prime as u64;
                 let n_buckets = (metagraph_node as f64).log2().ceil() as u64;
 
-                let (predecessor_index, other_drg_parents) = if self.is_legacy {
-                    (m_prime, &mut parents[..])
-                } else {
-                    (0, &mut parents[1..])
+                let (predecessor_index, other_drg_parents) = match self.api_version {
+                    ApiVersion::V1_0_0 => (m_prime, &mut parents[..]),
+                    ApiVersion::V1_1_0 => (0, &mut parents[1..]),
```
porcuquine, line 164: Here's an idea. What if instead of doing this kind of ad hoc inline matching, we *only* enact conditional code  by calling methods of the `ApiVersion`? The downside: loss of local context in the implementations. The upside: it would be very easy to maintain and inspect an exhaustive set of differences between versions. This example would be something like (untested):
```
trait DRGParentOrdering {
    …
}

impl DRGParentOrdering for ApiVersion::1_0_0 {
    fn parent_ordering(&self, m_prime: usize, parents: &mut[T]) -> (usize, &mut[T]) {
        (m_prime, parents)
    }

impl DRGParentOrdering for ApiVersion::1_1_0 {
    fn parent_ordering(&self, m_prime: usize, parents: &mut[T]) -> (usize, &mut[T]) {
        (0, parents[1..])
    }
}
```

… and then:
```
let (predecessor_index, other_drg_parents) = self.api_version.parent_ordering(m_prime, parents);
```

Add appropriate comments, and this becomes a very clear and self-documenting record of exactly what has changed between versions, as well as for each atomic patch's dependencies.

In addition to the self-documenting property, this would help enforce a discipline to combat the almost inevitable spaghetti-like nature of changes likely to accumulate. By establishing and enforcing such a pattern, you would disable and disarm the temptation to just make the implementation a little more complicated, a little more clever — a complexity which might accumulate in the name of simplicity and be difficult to untangle later. Such accumulations and potential multiplicative complexity are one of the things that worries me most about this path (although I understand why its _general_ shape is probably the right one now).

This is one reason I think the right discipline and a strong set of rules (even if they have to be amended in accordance with need over time) will go a long way toward protection. If something is too difficult to do within the trait-patch method, it might be a *good idea* to rethink it. Admittedly, this proposal has the disadvantage of moving the actual logic out of line. I think this could be solved with a little tooling. One idea: a macro which inlines the values for the *current — i.e. correct* version and defers to to trait implementations for the others.

```
let (predecessor_index, other_drg_parents) = with_version_or!(DRGParentOrdering, (0, &mut [parents[..]));
```

This assumes there *is* a concept of a 'current version', and that it is tracked in the code, but I think that may be a good/safe assumption.

Again, I realize this is more work, and introduces a certain amount of awkwardness. At the same time, I think the rigidity can be protective by:

- Making it *harder* to do ad hoc things or to introduce unexpected variations or multiplicative complexity into the code.
- Isolating and forcing centralized documentation of all version deltas.
- Forcing a higher level of discipline around the form of allowable version-based changes. Nothing prevents gradually extending the complexity of what is allowed, but the *need* to do that or else view the result of shortcutting it as at least technical debt, might force more deliberate attention around each such change introduced and how they interact.

I realize I this all comes across as very grave, but as we have discussed before, I think the approach we are taking here (however necessary) skirts a dangerous territory — and it's well worth extra forethought and extra discipline to ensure we stay on the right side of the line. Avoiding the problems is easy *now*, at the first very simple change. It's later, as they stack up, grow in complexity, and need to survive refactors, etc. that it might become hard to deal with. Whatever policies we can put in place to prevent that will be valuable. Enshrining those in code makes it just that much harder to break or change them in the heat of the moment.

porcuquine (commented):
Looks good. I made one detailed suggestion about how to potentially structure (with an argument for why something like it might be valuable).

I also noted a few places where what appear to have been unrelated code changes and refactors were introduced. It would really be better not to slip those in amidst enormous procedural refactoring. They raise the burden of review and make it harder to evaluate the changeset later.

storage-proofs/post/src/fallback/compound.rs
cryptonemo: Fair, but this brings these parts of the code up to date with the 'new normal' way of how these are done.

storage-proofs/core/src/drgraph.rs
cryptonemo: It's nice for simple cases (like the above), but won't work when parameters or return values change.  Each behavior change will have to be considered differently at some point.

storage-proofs/core/src/drgraph.rs
```diff
@@ -284,15 +286,15 @@ mod tests {
         let legacy_porep_id = porep_id(0);
         let new_porep_id = porep_id(5);
 
-        graph_bucket_aux::<H>(legacy_porep_id);
-        graph_bucket_aux::<H>(new_porep_id);
+        graph_bucket_aux::<H>(legacy_porep_id, ApiVersion::V1_0_0);
+        graph_bucket_aux::<H>(new_porep_id, ApiVersion::V1_1_0);
```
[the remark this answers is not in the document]
cryptonemo: There are a number of tests that have 'new' or 'legacy' porep that can be simplified by removing that distinction completely (they just need to be different).  We can revise those further, but I didn't see it as necessary on the first cut through.  I agree that 'new' and 'legacy' porep_ids is just confusing, so could see doing that now.

storage-proofs/core/src/api_version.rs
```diff
@@ -0,0 +1,58 @@
...
+                    )),
+                },
+                _ => Err(anyhow::format_err!(
+                    "Could not parse API Version from string (minor)"
+                )),
+            },
+            _ => Err(anyhow::format_err!(
+                "Could not parse API Version from string (major)"
+            )),
+        }
```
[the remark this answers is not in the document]
cryptonemo: For the simple 2 case distinction, this makes sense.  I don't think this simplifies much for the general cases though (i.e. when we're actually using a range of versions across semver), in which case it needs to be broken out further into something like what's already there.

storage-proofs/post/src/fallback/vanilla.rs
cryptonemo: I'll look at this one -- it's likely a rebase artifact.

storage-proofs/core/src/drgraph.rs
dignifiedquire: I like the idea, but I fear it is not quite practical, as we end up with circular imports. `ApiVersion` is defined in `storage-proofs-core`, but if we were to push all these details onto it, we would have to use types in core from `storage-proofs-porep` and `storage-proofs-post`, resulting in a circle.

One idea to help reduce the random places with api versions matching, is to at least always put the two different versions into a function call, named with a suffix of `_v_1_0_0` or simlar, for example

```rust
impl Bar {
  fn foo(&self, version: ApiVersion) {
    match version {
      ApiVersion::V1_0_0 => self.foo_v1_0_0(),
      ApiVersion::V1_1_0 => self.foo_v1_1_0(),
    }
  }

  fn foo_v1_0_0(&self) {
    todo!()
  }
  fn foo_v1_1_0(&self) {
    todo!()
  }
}

Comment (cryptonemo) [reply storage-proofs/core/src/drgraph.rs:164]: @dignifiedquire This is also a nice re-factor for the simple case, but breaks down when args or return types change (at least the dispatch `foo` above.  It could work if there was no dispatch and just separate versioned foo methods -- but that pushes the dispatch (i.e. match logic) up further if the return type changes, which is exactly what's trying to be avoided, no?).  The upshot of these discussions are that they are re-factors, which can be done at any time (and probably will no matter what we choose here), independent of the fact that we've settled on a logic switching mechanism for upgrades.

For this first cut, imposing a formality around something that's known to not work in all cases seems arbitrary, but if it's the will of the team, we can.

File: storage-proofs/post/src/fallback/compound.rs (hunk 4 above)
Comment (cryptonemo) [reply storage-proofs/post/src/fallback/compound.rs:71]: @dignifiedquire Thoughts here? I can back this out of this PR to make a new PR for just those updates, but it seemed clear enough to me on this pass.  Either way, it's pretty straight-forward.

File: storage-proofs/post/src/fallback/vanilla.rs (hunk 5 above)
Comment (cryptonemo) [reply storage-proofs/post/src/fallback/vanilla.rs]: Updated

File: storage-proofs/core/src/api_version.rs (hunk 8 above)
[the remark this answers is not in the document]
Comment (cryptonemo) [reply storage-proofs/core/src/api_version.rs:47]: Updated (minus the syntax error)

File: storage-proofs/core/src/drgraph.rs (hunk 6 above)
Comment (dignifiedquire) [reply storage-proofs/core/src/drgraph.rs:164]: I agree we don’t need to figure this out before merging

File: storage-proofs/post/src/fallback/compound.rs (hunk 4 above)
Comment (dignifiedquire) [reply storage-proofs/post/src/fallback/compound.rs:71]: can you put it into its own commit, fine in this PR

Comment (cryptonemo) [reply storage-proofs/post/src/fallback/compound.rs:71]: For review, yes, I'll push shortly.  When we squash/merge it makes little difference in that case :-)

Comment (dignifiedquire) [reply storage-proofs/post/src/fallback/compound.rs:71]: I was thinking we squash the rest before merge and merge regularly so we end up with two commits on master

Comment (cryptonemo) [reply storage-proofs/post/src/fallback/compound.rs:71]: Ok, updated into 2 commits, although I much prefer the squash/merge to the 'regular' merge.

Comment (dignifiedquire) [reply storage-proofs/post/src/fallback/compound.rs:71]: next time again

File: storage-proofs/core/src/api_version.rs (hunk 8 above)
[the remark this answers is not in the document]
Comment (porcuquine) [reply storage-proofs/core/src/api_version.rs:47]: You can match ranges in patterns too. I think if you order the clauses right, you *might* be able to use that fact to keep it relatively simple and legible. Maybe it will break down sometimes too, but it might be worth at least trying to isolate the complexity so remains mostly clear.

https://doc.rust-lang.org/book/ch18-03-pattern-syntax.html#matching-ranges-of-values-with-

File: storage-proofs/core/src/drgraph.rs (hunk 7 above)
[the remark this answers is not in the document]
Comment (porcuquine) [reply storage-proofs/core/src/drgraph.rs:290]: I think it's fine to wait — was just checking to make sure I understand the new world.

File: storage-proofs/core/src/drgraph.rs (hunk 6 above)
Comment (porcuquine) [reply storage-proofs/core/src/drgraph.rs:164]: Don't let it hold you up. I more wanted to inject some ideas into the mix for as the ongoing refactors @cryptonemo predicts happen. Changes to the API signature are obviously a larger problem that will need to be considered (perhaps soon). That may be somewhat orthogonal.

This is all just continued brainstorming and not meant to be blocking — just as feed for ongoing design and consideration:

@dignifiedquire Just continuing to brainstorm: you could implement the traits in the most local scope they are needed. That would be less centralized, but still easy to inspect and would capture information of itself.

File: storage-proofs/core/src/api_version.rs (hunk 8 above)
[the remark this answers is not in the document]
Comment (porcuquine) [reply storage-proofs/core/src/api_version.rs:52]: Just some more thoughts — nothing concrete to do now. Are you planning to use the semver matching facilities? In that case, the more complicated ranges could be encoded directly.

Example:
```
match api_version {
    v if VersionReq::parse("^1.2.3").matches(v) =>  todo!(),
    v if VersionReq::parse("^2.0").matches(v) => todo!(),
    v if VersionReq::parse("^1.0.0").matches(v) => todo!(),
}
```

I didn't think through the actual range matching in detail; this was just to indicate the form. Depending on what kinds of things come up, it might make sense to replace these explicit `matches` calls with custom predicates that can live in one place and capture the intended partitioning of versions as it evolves. If the same partitions are used in multiple decisions and become at all complicated, having a single source of truth to audit, update, and keep in sync between usages may be useful.

dignifiedquire (approved):

PR #1363, opened 2020-11-11
fix: update paramcache run for metrics capture CI job

### LINKED ISSUE #1361: Metrics capture is broken after source re-org
Update CI job to point to new location

cryptonemo:
Aims to resolve #1361 

Enabled metrics-capture to test CI -- will remove if successful before merging.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -249,7 +249,7 @@ jobs:
       - run:
           name: Ensure existence of Groth parameters and keys on remote host
           command: |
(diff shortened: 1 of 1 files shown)


dignifiedquire (commented):
is this the only ci job that uses a parambin?

cryptonemo:
> is this the only ci job that uses a parambin?

Looks like it.

dignifiedquire (approved):

Issue #1349, opened 2020-11-06
tests: move slow/complex tests into `tests` folders


dignifiedquire:
Only small unit tests should remain in the implementation files.

cryptonemo:
Also addressed by the in progress PR: https://github.com/filecoin-project/rust-fil-proofs/pull/1364

cryptonemo:
Closed with #1364

PR #1364, opened 2020-11-12
feat: move some tests into separate test directory

cryptonemo:
Not sure this is complete yet, but it's a start.


cryptonemo:
> Not sure this is complete yet, but it's a start.

Definitely not complete ...

storage-proofs/porep/src/stacked/vanilla/graph.rs
```diff
@@ -438,241 +438,3 @@ where
     G: Graph<H>,
 {
 }
-
```
vmx, line 441: I'm not sure if these tests should be moved. From my limited understanding the tests in this file are more like unit tests, which I think belong to the file.

As I currently work on generating coverage reports, I think tests with in files should lead to good coverage, tests in the `tests` directory are integration tests that test the bigger picture.

cryptonemo: I'm ok with whatever makes sense.  This PR is a bit of a mess in that, it might help if we had a list of tests we want moved to sort of narrow down the scope.

cryptonemo: @vmx While I'm out, feel free to take this over as well :-)

cryptonemo:
The goal is to resolve https://github.com/filecoin-project/rust-fil-proofs/issues/1349

dignifiedquire:
@DrPeterVanNostrand what's the status on this?

DrPeterVanNostrand:
This PR moves non-unit tests in `strorage-proofs` into one of the folders `{core, porep, post}/tests/`.

The non-refactoring changes are:
- removes duplicate `porep::TEST_SEED` and `post::TEST_SEED`, uses `core::TEST_SEED` everywhere
- adds constructor `PoRCircuit::new()`
- removes trait method `core::proof::MerkleProofTrait::break_me()`

DrPeterVanNostrand:
@dignifiedquire I pushed an update to this PR, it should be good to go

cryptonemo:
@DrPeterVanNostrand This looks about right to me.  Note that I can't approve because github won't let me since I issued it.

dignifiedquire (commented):
I am unable to find the equivalent of `make_bogus_proof` in the new tests, did you remove breaking the proof and checking if it invalidates?

cryptonemo:
> I am unable to find the equivalent of `make_bogus_proof` in the new tests, did you remove breaking the proof and checking if it invalidates?

I'll let @DrPeterVanNostrand address this, but it seems to be just done differently, e.g.: https://github.com/filecoin-project/rust-fil-proofs/blob/split-tests/storage-proofs/core/tests/por_vanilla.rs#L161

dignifiedquire (dismissed):

dignifiedquire:
@cryptonemo thanks, that's what I didn't find

cryptonemo:
Sorry @DrPeterVanNostrand, I merged a different PR first and this needs a rebase.  Considering this is the more difficult to rebase, I probably should have waited.  Let me know if any issue, but I think this should be good to go when ready.

DrPeterVanNostrand:
[I added one to the leaf label](https://github.com/filecoin-project/rust-fil-proofs/pull/1364/files#diff-47e49c4f75cbf1688277fd0b8f511a4dbeea69021d278322792119995127a5bbR162) to invalidate the por.

DrPeterVanNostrand:
@cryptonemo np, I'll rebase

PR #1367, opened 2020-11-15
Read trees from GPU within scope of building threads.

porcuquine:
This PR fixes ensure the channels providing the finished subtrees for writing to disk are read within the scope of the threads building the trees. Otherwise, the `rayon::scope` will never complete and writing will hang. This is a regression (vs development of that feature) introduced during the final rebase of #1335. This fix should be merged before #1335 is released, and we should make sure to run lifecycle tests using GPU tree building to exercise it.


porcuquine:
Looks like this 'fix' now hangs when using CPU – as shown by CI failures. I'll look at it again and push something that succeeds in both cases.

porcuquine:
I don't see this problem locally, rerunning CI to see if it reproduces there.

porcuquine:
If I SSH into the test box and run the test which hangs by itself, it gets through tree building fine (the errors out because of missing parameters).

dignifiedquire:
Manually merged into #1358, which I will merge, as that has passing CI

PR #1375, opened 2020-11-24
fix tree-c and tree-r-last generation in GPU mode

qy3u:
It will only 'add_columns' once for each tree and then jump to next tree in current implemention.


porcuquine (approved):
You're absolutely right. Thank you for catching this.

I verified the problem and that this fixes.

@dignifiedquire When we get get GPU tests set up in CI, we should make sure the batch sizes are set small enough to ensure multiple batches are required on the small sectors.

Issue #1378, opened 2020-12-01
Improve checkout cores test (fails by default)


cryptonemo:
### Description

The `checkout cores` test always fails when the tests are run, by default, in parallel.  Some options to fix it are to require tests be run in a single threaded manner (not generally practical due to extended wait times during development cycles), or to use a feature such as "single-threaded = []" and making the test conditionally compiled for when it's specified.  We could add a CI job for this single test configuration, but also have the option to run it locally if required.

### Acceptance criteria

All tests run need to pass.

### Where to begin

Proposed diff:

```
diff --git a/storage-proofs/porep/Cargo.toml b/storage-proofs/porep/Cargo.toml
index d0957a6e..a8ff857d 100644
--- a/storage-proofs/porep/Cargo.toml
+++ b/storage-proofs/porep/Cargo.toml
@@ -53,6 +53,7 @@ default = ["pairing", "gpu"]
 gpu = ["storage-proofs-core/gpu", "filecoin-hashers/gpu"]
 pairing = ["storage-proofs-core/pairing", "bellperson/pairing", "neptune/pairing", "filecoin-hashers/pairing"]
 blst = ["storage-proofs-core/blst", "bellperson/blst", "neptune/blst", "filecoin-hashers/blst"]
+single-threaded = []
 
 [[bench]]
 name = "encode"
diff --git a/storage-proofs/porep/src/stacked/vanilla/cores.rs b/storage-proofs/porep/src/stacked/vanilla/cores.rs
index 2aed3d47..cc7abcfa 100644
--- a/storage-proofs/porep/src/stacked/vanilla/cores.rs
+++ b/storage-proofs/porep/src/stacked/vanilla/cores.rs
@@ -201,6 +201,7 @@ mod tests {
     }
 
     #[test]
+    #[cfg(feature="single-threaded")]
     fn test_checkout_cores() {
         let checkout1 = checkout_core_group();
         dbg!(&checkout1);
```

cryptonemo:
Resolved with #1383

PR #1383, opened 2020-12-02
fix: put checkout_cores test behind a single-threaded feature

cryptonemo:
feat: add checkout_cores test to CI, run single threaded
doc: fix typo

A possible way to address #1378

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -147,6 +147,7 @@ jobs:
             ulimit -u 20000
             ulimit -n 20000
             cargo +<< pipeline.parameters.nightly-toolchain >> -Zpackage-features test --all --verbose --release lifecycle -- --ignored --nocapture
+            cargo +<< pipeline.parameters.nightly-toolchain >> -Zpackage-features test --features single-threaded --release checkout_cores -- --test-threads=1
           no_output_timeout: 30m
           environment:
             RUST_TEST_THREADS: 1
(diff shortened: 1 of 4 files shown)


.circleci/config.yml
```diff
@@ -147,6 +147,7 @@ jobs:
             ulimit -u 20000
             ulimit -n 20000
             cargo +<< pipeline.parameters.nightly-toolchain >> -Zpackage-features test --all --verbose --release lifecycle -- --ignored --nocapture
+            cargo +<< pipeline.parameters.nightly-toolchain >> -Zpackage-features test --features single-threaded --release checkout_cores -- --test-threads=1
```
dignifiedquire, line 150: this is missing the `--all` package specificer

cryptonemo: Correct, that's the purpose of this PR.  The all is performed, followed by the single checkout_cores test that is isolated.

dignifiedquire: but it still needs to select a package?

cryptonemo: The CI log shows what I see locally, which is the single test running (after the all): https://app.circleci.com/pipelines/github/filecoin-project/rust-fil-proofs/2838/workflows/c44d3b1c-e69e-43cc-8c06-6c4e533ccf92/jobs/62773

cryptonemo: Ok, adding the package is more direct, I'll change that (just tried locally).

cryptonemo:
Currently, the sdr checkout test is done on the pairing side and not mirrored on the blst side.  I don't think there's any need to run it more than once.  Should it be moved out to a separate job entirely? I only really want to know if it's failing for some reason on CI.

dignifiedquire (approved):

PR #1384, opened 2020-12-03
perf(storage-proofs-post): increase parallelism in fallback-post

dignifiedquire:
This improves the parallelism during both vanilla `prove` and vanilla `verify`.

### DIFF
--- a/storage-proofs/post/src/fallback/vanilla.rs
+++ b/storage-proofs/post/src/fallback/vanilla.rs
@@ -248,11 +248,6 @@ pub fn generate_leaf_challenge_inner<T: Domain>(
     leaf_challenge % (pub_params.sector_size / NODE_SIZE as u64)
 }
 
-enum ProofOrFault<T> {
-    Proof(T),
-    Fault(SectorId),
-}
-
 // Generates a single vanilla proof, given the private inputs and sector challenges.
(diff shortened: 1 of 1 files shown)


cryptonemo (approved):
+1 for the enum removal

dignifiedquire:
thanks to @robyhugeman666 for reminding me that we can do more here

robyhugeman666:
Very honored to be involved.


------------------------------------------------------------------

storage-proofs/post/src/fallback/vanilla.rs
```diff
@@ -507,82 +525,86 @@ impl<'a, Tree: 'a + MerkleTreeTrait> ProofScheme<'a> for FallbackPoSt<'a, Tree>
...
+                        .collect::<Result<Vec<bool>>>()?;
+
+                    Ok(is_valid_list.into_iter().all(|v| v))
+                })
+                .reduce(
+                    || Ok(true),
+                    |all_valid, is_valid| Ok(all_valid? && is_valid?),
+                )?;
+            if !is_valid {
+                return Ok(false);
```
free1139, line 607: @dignifiedquire we also need continue the check and record the fault sectors here. I had lost the storage power cause some sectors failed in here, but i forgot the detail.

PR #1386, opened 2020-12-04
fix: ensure that builds without gpu work

### LINKED ISSUE #1346: ensure all code compiles without OpenCL if the `gpu` flag is not enabled

dignifiedquire:
Closes #1346

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -165,6 +165,29 @@ jobs:
             RUST_TEST_THREADS: 1
             FIL_PROOFS_USE_MULTICORE_SDR: true
 
+  test_no_gpu:
+    executor: default
+    environment: *setup-env
+    steps:
+      - checkout
+      - attach_workspace:
+          at: "."
+      - restore_cache:
+          keys:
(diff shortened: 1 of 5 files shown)


cryptonemo:
Is there a CI config *for* GPU tests?

dignifiedquire:
The other builds all enable the `gpu` feature, as it is enabled by default. We don't have proper gpu tests yet on CI, which is a todo for another day.

cryptonemo:
I can't recall the specifics, but when this was first in development, I believe the option to compile out the GPU support was explicitly rejected.  This PR is adding that back, so just wanted to mention it's worth thinking back on why that was, or if it's still relevant.

dignifiedquire:
interesting, I don’t remember that, maybe @porcuquine  can shed some light?

porcuquine:
As I vaguely recall, the first version of GPU support in neptune only exposed the batch hasher API when the GPU feature was enabled. So in order to keep all `rust-fil-proofs` code paths building without even more conditional code, we opted to always build neptune with GPU enabled. If that is no longer needed, I don't think there's a remain reason to require the flag.

cryptonemo (approved):

PR #1395, opened 2021-01-15
Proof aggregation api

cryptonemo:


cryptonemo:
Not ready yet, but somewhat presentable at this point.

cryptonemo:
TODO: When API is out of draft status, update FIP details https://github.com/filecoin-project/FIPs/pull/72

ozhtdong:
I guess it won't work for 32Gb/64Gb sectors (yet), since the number of proofs can't be both multiples of 10 and power of 2.

cryptonemo:
> I guess it won't work for 32Gb/64Gb sectors (yet), since the number of proofs can't be both multiples of 10 and power of 2.

There are ways to address that -- the simplest of which is to pad by duplicating some proof a number of times to meet the pow2 requirement (which the tests do for now).  Other options are being investigated.

cryptonemo:
SRS Publication to IPFS can now be done as follows:

```
cargo run --release --bin srspublish -- --ipfs-bin `which ipfs`
```

SRS file download can now be done like this:

```
cargo run --release --bin paramfetch -- -j srs-inner-product.json
```

filecoin-proofs/src/api/seal.rs
```diff
@@ -538,6 +541,345 @@ pub fn seal_commit_phase2<Tree: 'static + MerkleTreeTrait>(
...
+            proofs.last().is_some(),
+            "invalid last proof for duplication"
+        );
+        proofs.push(
+            proofs
+                .last()
+                .expect("failed to access last proof for duplication")
+                .clone(),
+        );
+    }
```
nikkolasg, line 764: Could we do like in the following ?
```
let diff = target - proofs.len();
let last = proofs.last().expect(..).clone();
let mut padding = (0..diff).map(|_| last.clone());
proofs.append(&mut padding);
```
Looks more succinct ?

filecoin-proofs/src/api/seal.rs
```diff
@@ -538,6 +541,345 @@ pub fn seal_commit_phase2<Tree: 'static + MerkleTreeTrait>(
...
+    info!("verify_aggregate_seal_commit_proofs:start");
+
+    ensure!(!commit_inputs.is_empty(), "cannot verify with empty inputs");
+    ensure!(
+        commit_inputs.len() % aggregated_proofs_len == 0,
+        "invalid number of inputs provided"
+    );
+
+    let num_inputs = commit_inputs.len();
+    let num_inputs_per_proof = num_inputs / aggregated_proofs_len;
```
nikkolasg, line 814: isn't this supposed to be fixed ? i.e. shouldn't we read that info from .. porep config or stg ? 
What we want to do is twofold:
1. ensure there are always X public inputs **per proof** -> this is different than checking there are X * Y public inputs (where Y is the number of proofs) in total
2. ensure the number of proofs is a power of two, if it's not, we duplicate X public inputs as much as necessary

It feels 1. and 2. are mixed here, and 1. is checking the total number of public inputs rather than the number of public inputs per proofs.

filecoin-proofs/tests/api.rs
```diff
@@ -196,9 +196,167 @@ fn seal_lifecycle<Tree: 'static + MerkleTreeTrait>(
...
+        proofs_to_aggregate,
+        aggregate_proof,
+        commit_inputs,
+    )?;
+    assert!(verified);
+
+    Ok(())
+}
+
+fn aggregate_proofs<Tree: 'static + MerkleTreeTrait>(
```
nikkolasg, line 322: I'd suggest we put all of the test functions and helpers in a 
```
#[cfg(test)]
mod tests {
}
```
so all tests related functions like this one are only compiled in tests ?

cryptonemo: I thought this wasn't needed since we already separated the code into a test module?

cryptonemo:
In memory caching of srs prover and verifier keys (per aggregated proof count) is now separate (instead of always keeping the pair together).

cryptonemo:
Rebased against master

Note: sha2raw tests are not building

fil-proofs-param/src/bin/srspublish.rs
```diff
@@ -0,0 +1,184 @@
...
+            && !has_extension(filename, VERIFYING_KEY_EXT)
+            && !has_extension(filename, PARAMETER_METADATA_EXT)
+        {
+            warn!("file has invalid extension: {}, ignoring file", filename);
+        }
+        return false;
+    }
+    let version = filename.split('-').nth(0).unwrap();
+    let version_is_valid =
+        version.get(0..1).unwrap() == "v" && version[1..].chars().all(|c| c.is_digit(10));
```
dignifiedquire, line 42: `version[1..]` will panic if the name is only character long

fil-proofs-param/src/bin/srspublish.rs
```diff
@@ -0,0 +1,184 @@
...
+    }
+    true
+}
+
+fn get_filenames_in_cache_dir() -> Vec<String> {
+    let path = parameter_cache_dir();
+    if !path.exists() {
+        warn!("param cache dir does not exist (no files to publish), exiting");
+        exit(0);
+    }
```
dignifiedquire, line 58: shouldn't this be `exit(1)` it's not really success if nothing was done

filecoin-proofs/src/api/seal.rs
```diff
@@ -537,6 +543,346 @@ pub fn seal_commit_phase2<Tree: 'static + MerkleTreeTrait>(
     Ok(out)
 }
 
+pub fn get_seal_inputs<Tree: 'static + MerkleTreeTrait>(
```
dignifiedquire, line 546: can we get some doc comments on this?

filecoin-proofs/src/api/seal.rs
```diff
@@ -537,6 +543,346 @@ pub fn seal_commit_phase2<Tree: 'static + MerkleTreeTrait>(
...
+                &compound_public_params.vanilla_params,
+                Some(k),
+            )
+        })
+        .collect::<Result<_>>()?;
+
+    Ok(inputs)
+}
+
+// Proposed method (untested), perhaps not useful.  Remove if not needed
```
dignifiedquire, line 615: what's the state of this?

filecoin-proofs/src/api/seal.rs
```diff
@@ -537,6 +543,346 @@ pub fn seal_commit_phase2<Tree: 'static + MerkleTreeTrait>(
...
+                inputs_added += 1;
+            }
+        }
+    }
+
+    Ok(inputs)
+}
+*/
+
+pub fn aggregate_seal_commit_proofs<Tree: 'static + MerkleTreeTrait>(
```
dignifiedquire, line 718: doc comments please

filecoin-proofs/src/api/seal.rs
```diff
@@ -537,6 +543,346 @@ pub fn seal_commit_phase2<Tree: 'static + MerkleTreeTrait>(
...
+    let mut proofs: Vec<_> = commit_outputs
+        .iter()
+        .fold(Vec::new(), |mut acc, commit_output| {
+            acc.extend(
+                MultiProof::new_from_reader(
+                    Some(partitions),
+                    &commit_output.proof[..],
+                    &verifying_key,
+                )
+                .expect("failed to construct multi proof from bytes")
```
dignifiedquire, line 735: this should not panic

filecoin-proofs/src/api/seal.rs
```diff
@@ -537,6 +543,346 @@ pub fn seal_commit_phase2<Tree: 'static + MerkleTreeTrait>(
...
+    porep_config: PoRepConfig,
+    commit_outputs: &[SealCommitOutput],
+) -> Result<AggregateSnarkProof> {
+    info!("aggregate_seal_commit_proofs:start");
+
+    let partitions = usize::from(PoRepProofPartitions::from(porep_config));
+    let verifying_key = get_stacked_verifying_key::<Tree>(porep_config)?;
+    let mut proofs: Vec<_> = commit_outputs
+        .iter()
+        .fold(Vec::new(), |mut acc, commit_output| {
```
dignifiedquire, line 728: you can use [`flat_map`](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.flat_map), makes it a little easier to read

filecoin-proofs/src/api/seal.rs
```diff
@@ -537,6 +543,346 @@ pub fn seal_commit_phase2<Tree: 'static + MerkleTreeTrait>(
...
+            );
+
+            acc
+        });
+    trace!(
+        "aggregate_seal_commit_proofs called with {} commit_outputs",
+        commit_outputs.len()
+    );
+
+    let target = if commit_outputs.len() == 1 {
```
dignifiedquire, line 746: maybe `target_len`?

dignifiedquire: should check somewhere that `commit_outputs` is not empty

filecoin-proofs/tests/api.rs
dignifiedquire: > I thought this wasn't needed since we already separated the code into a test module?

correct, this is only needed when writing tests inside a source file, but not when in the `tests` module

dignifiedquire, line 322: I can haz doc comment?

storage-proofs-core/src/compound_proof.rs
```diff
@@ -272,6 +277,27 @@ where
             .collect()
     }
 
+    fn aggregate_proofs(
```
dignifiedquire, line 280: docs?

storage-proofs-core/src/compound_proof.rs
```diff
@@ -272,6 +277,27 @@ where
...
+        prover_srs: &ProverSRS<Bls12>,
+        proofs: &[groth16::Proof<Bls12>],
+    ) -> Result<AggregateProof<Bls12>> {
+        Ok(aggregate_proofs::<Bls12>(prover_srs, proofs)?)
+    }
+
+    fn verify_aggregate_proofs(
```
dignifiedquire, line 287: docs?

PR #1401, opened 2021-01-25
fix: avoid one in memory copy for unsealing

dignifiedquire:
Refactors extraction to overwrite the provided data, rather than allocating another one.

### DIFF
--- a/filecoin-proofs/src/api/mod.rs
+++ b/filecoin-proofs/src/api/mod.rs
@@ -227,9 +227,9 @@ where
         .read(true)
         .write(true)
         .open(&sealed_path)?;
-    let mut data = unsafe { MmapOptions::new().map_mut(&mapped_file)? };
+    let mut data = unsafe { MmapOptions::new().map_copy(&mapped_file)? };
 
-    unseal_range_inner::<_, _, Tree>(
+    let result = unseal_range_inner::<_, _, Tree>(
         porep_config,
         cache_path,
         &mut data,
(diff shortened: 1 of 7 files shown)


cryptonemo:
This is a memory visualization of a full 32GiB seal lifecycle test using the code from this PR, with `unseal` starting around 10pm.

unseal-improvements-01-26-2021


For comparison, this is a partial memory visualization of a 32GiB seal lifecycle test on the `master` branch, with `unseal` starting around 1:28pm.  Note that a large part of the test visualization at the start was unfortunately lost -- and that this run is not yet complete.

Screenshot(120)

cryptonemo:
To most directly take advantage of this, we also need the exposed `unseal_range` method in rust-fil-proofs-api to be re-factored in terms of `get_unsealed_range`.

https://github.com/filecoin-project/rust-filecoin-proofs-api/blob/master/src/seal.rs#L843

The `unseal_range` API call does not directly benefit from the memory reduction via mmap.  It's possible that lotus does not use `unseal_range` in preference to `get_unsealed_range`, but it should be updated in any case since it can be done transparently to the API caller and the memory performance is strictly reduced.

cryptonemo:
This is the latest run, with the updated/correct code.  This uses the `blst,gpu2` feature, so you can see that tree building is much more efficient in terms of time and memory.  Unsealing starts around `4:41am`

api-1956d861dd69f895_418389

cryptonemo:
And here is the latest run of `master`, comparable to the improved image above.

api-1956d861dd69f895_422637

dignifiedquire (commented):
lgtm (can't approve my own PR)

cryptonemo (approved):

PR #1420, opened 2021-02-25
Optimize opencl and make it default gpu feature.

porcuquine:
#1397 Added support for neptune's `opencl` feature, but did not demonstrate the expected performance gain. (Expectation was ~2x. See: https://github.com/filecoin-project/neptune/pull/78.)

Apparently the bottleneck was the `read_range` provided by `merkletree::store::disk`, which ended up calling `PoseidonDomain::from_slice` on every `Fr` element when reading from disk. The cleanest way to fix this is probably to optimize `read_range` to perform better. However, I was not able to find a simple way to make this change in the face of the layers of generic types.

Instead, I took advantage of definite knowledge that the underlying data is `Fr` and use an unsafe transmute from bytes.

As previously discussed, this PR makes `neptune/opencl` the default feature when the `gpu` feature is active — and removes the `gpu2` feature. Whether this should be merged immediately or wait until the `gpu2` flag has been released and widely tested depends on whether we already have confidence in it. As far as I know, there is no reason to believe `neptune/opencl` is problematic: it's dramatically simpler and performs better — but I will let @cryptonemo and/or @dignifiedquire make the decision.

Shown below, now column tree building takes about 71 seconds, and regular tree-building takes about 11 seconds. This is compared with 75 and 10 seconds on the isolated neptune benchmark (gbench). Total time for building both trees is now just over 11 minutes, which is as originally expected.

```
2021-02-25T01:43:13.825 INFO storage_proofs_porep::stacked::vanilla::proof > generating tree c using the GPU
2021-02-25T01:43:13.825 INFO storage_proofs_porep::stacked::vanilla::proof > Building column hashes
2021-02-25T01:43:13.940 INFO neptune::proteus::gpu > device: Device { brand: Nvidia, name: "GeForce RTX 2080 Ti", memory: 11551440896, bus_id: Some(33), platform: Platform(PlatformId(0x7f914809c1c0)), device: Device(DeviceId(0x7f91480b60e0)) }
2021-02-25T01:44:26.966 INFO storage_proofs_porep::stacked::vanilla::proof > persisting base tree_c 1/8 of length 153391689
2021-02-25T01:45:38.013 INFO storage_proofs_porep::stacked::vanilla::proof > persisting base tree_c 2/8 of length 153391689
2021-02-25T01:46:49.541 INFO storage_proofs_porep::stacked::vanilla::proof > persisting base tree_c 3/8 of length 153391689
[...]

PR #1422, opened 2021-03-02
Optimize tree-building for neptune/opencl (gpu2 flag).

porcuquine:
This PR supersedes #1420 — since we have decided to keep `neptune/opencl` behind the `gpu2` feature flag for now.

@vmx Can you rebase your CI work on this branch and continue here? I made one modification to the line that actually runs the lifecycle tests. The previous version didn't set the flags, but the new one should correctly specify `gpu2`.

### DIFF
--- a/storage-proofs-porep/src/stacked/vanilla/proof.rs
+++ b/storage-proofs-porep/src/stacked/vanilla/proof.rs
@@ -441,13 +441,11 @@ impl<'a, Tree: 'static + MerkleTreeTrait, G: 'static + Hasher> StackedDrg<'a, Tr
         TreeArity: PoseidonArity,
     {
         use std::cmp::min;
-        use std::ops::Range;
         use std::sync::{mpsc::sync_channel, Arc, RwLock};
 
         use bellperson::bls::Fr;
-        use ff::Field;
         use fr32::fr_into_bytes;
-        use generic_array::{sequence::GenericSequence, GenericArray};
+        use generic_array::GenericArray;
         use merkletree::store::DiskStore;
         use neptune::{
             batch_hasher::BatcherType,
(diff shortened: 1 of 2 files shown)


porcuquine:
Okay, I moved @vmx's CI work over to this branch and adapted it to run lifecycle tests for both `gpu` and `gpu2` features. These are now passing, and I've also run small passing lifecycle tests locally for both.

I've also verified that both the `gpu` and `gpu2` code paths still perform as expected based on current `master` and the equivalent (but differently factored) code in #1420.

I think this is ready to go once the rest of CI passes.

dignifiedquire (approved):

storage-proofs-porep/src/stacked/vanilla/proof.rs
```diff
@@ -493,43 +491,107 @@ impl<'a, Tree: 'static + MerkleTreeTrait, G: 'static + Hasher> StackedDrg<'a, Tr
...
+                                        layers
+                                    ];
+
+                                rayon::scope(|s| {
+                                    // capture a shadowed version of layer_data.
+                                    let layer_data: &mut Vec<_> = &mut layer_data;
+
+                                    // gather all layer data in parallel.
+                                    s.spawn(move |_| {
+                                        for (layer_index, mut layer_bytes) in
```
qy3u, line 565: I think you may want to spawn in the iteration.The current implemention seems not paralleled actually.

storage-proofs-porep/src/stacked/vanilla/proof.rs
```diff
@@ -493,43 +491,107 @@ impl<'a, Tree: 'static + MerkleTreeTrait, G: 'static + Hasher> StackedDrg<'a, Tr
...
+                                // Allocate layer data array and insert a placeholder for each layer.
+                                let mut layer_data: Vec<Vec<Fr>> =
+                                    vec![Vec::with_capacity(chunked_nodes_count); layers];
+
+                                rayon::scope(|s| {
+                                    // capture a shadowed version of layer_data.
+                                    let layer_data: &mut Vec<_> = &mut layer_data;
+
+                                    // gather all layer data in parallel.
+                                    s.spawn(move |_| {
```
qy3u, line 519: I think you may want to spawn in the iteration.The current implemention seems not paralleled actually.

storage-proofs-porep/src/stacked/vanilla/proof.rs
```diff
@@ -493,43 +491,107 @@ impl<'a, Tree: 'static + MerkleTreeTrait, G: 'static + Hasher> StackedDrg<'a, Tr
...
+                                        vec![0u8; chunked_nodes_count * std::mem::size_of::<Fr>()];
+                                        layers
+                                    ];
+
+                                rayon::scope(|s| {
+                                    // capture a shadowed version of layer_data.
+                                    let layer_data: &mut Vec<_> = &mut layer_data;
+
+                                    // gather all layer data in parallel.
+                                    s.spawn(move |_| {
```
qy3u, line 564: Ditto.

storage-proofs-porep/src/stacked/vanilla/proof.rs
porcuquine: You're right that this `spawn` seems unnecessary — not sure if it was always like that or got introduced in a refactor at some point. The change you suggest doesn't seem to help though (in my benchmark at least): the bottleneck is in the conversion that follows. Since the goal now is to release with minimal changes, and since all options (status quo, actually spawning in loop, just removing the whole scope/spawn) seem to perform equivalently, I think we should just leave as-is for now — to minimize change and risk, given that this has already been tested/benchmarked/etc quite a bit. It can be removed completely in some future cleanup that aims to improve further.

storage-proofs-porep/src/stacked/vanilla/proof.rs
dignifiedquire: @porcuquine lets remove the unneeded `spawn` at least from the new code path

cryptonemo: I have this in testing now with a new issue prepped for the various improvements noted here.  If my testing goes ok, I'll merge this and open that issue.

cryptonemo:
Would be interesting to see data on larger/actual
gpu-vs-gpu2-32kib-lifecycle
 tests, but here's a GPU comparison of a small test (32KiB seal lifecycle using first `gpu` and then `gpu2`).

PR #1427, opened 2021-03-05
feat: split up pc1/pc2 in the window post bench

### LINKED ISSUE #1399: add benchmark for tree building
### Description

I can benchmark building column trees directly from neptune using the `gbench` CLI program. However, performance when actually building column trees (`comm_c`) and regular trees (`comm_r_last`) during precommit 2 is different because the data is read from disk and generally flows through the code paths in `stacked/vanilla/proof.rs`.

I can run that and get timing information from logs by either running a 32 GiB lifecycle test, or by using the `benchy` CLI tool. However, both of these options only generate trees after first replicating an entire sector. For 32 GiB sectors, which is what I want to measure, this takes like two hours — which is not very practical as part of a development cycle when trying to optimize tree building throughput.

Ideally, we would have a way — whether through `benchy` or some other standalone utility — to generate either column trees or regular trees for any sector size (but at least supporting 32 GiB) without first replicating. It's fine to use random data (which must consist of valid `Fr`s), or even all zeroes as input — but the data does have to be on disk and in the expected format.

### Acceptance criteria

I can run a program from the command line and cause it to build either column trees or regular trees (depending on the option provided) for any supported sector size (or at least 32GiB), without a significant delay. This needs to exercise the normal code paths and take input from replica layers written to disk. (If writing the layers to disk is time-consuming, there could be a separate setup command which writes to a specified location once but need not be run every time.)

### Where to begin

https://github.com/filecoin-project/rust-fil-proofs/blob/master/storage-proofs-porep/src/stacked/vanilla/proof.rs#L371-L399
https://github.com/filecoin-project/rust-fil-proofs/blob/master/storage-proofs-porep/src/stacked/vanilla/proof.rs#L724-L754
https://github.com/filecoin-project/rust-fil-proofs/blob/master/fil-proofs-tooling/src/bin/benchy/prodbench.rs#L236

cryptonemo:
docs: add usage notes in the readme

Resolves #1399

### DIFF
--- a/README.md
+++ b/README.md
@@ -80,6 +80,37 @@ The main benchmarking tool is called `benchy`.  `benchy` has several subcommands
 > cargo run --release --bin benchy -- prodbench
 ```
 
+### Window PoSt Bench usages
+
+The Window PoSt bench can be used a number of ways, some of which are detailed here.
+
+First, you can run the benchmark and preserve the working directory like this:
+```
+cargo run --release --bin benchy -- window-post --size 2KiB --cache window-post-2KiB-dir --preserve-cache
+```
+
+Then if you want to run the benchmark again to test commit-phase2, you can quickly run it like this:
+```
+cargo run --release --bin benchy -- window-post --size 2KiB --skip-precommit-phase1 --skip-precommit-phase2 --skip-commit-phase1 --cache window-post-2KiB-dir
+```
+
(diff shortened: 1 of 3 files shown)


porcuquine (approved):
LGTM. Thank you.

dignifiedquire (approved):

Issue #1426, opened 2021-03-05
Improve RAM performance and organization of GPU2 work


cryptonemo:
### Description

There are improvements noted in the PR https://github.com/filecoin-project/rust-fil-proofs/pull/1422

Also, 32GiB benchmarking is required to determine the performance difference of `gpu` and `gpu2` before moving further.  See small GPU example graph in PR mentioned above.  RAM usage and wall clock time considerations are also important to measure and optimize.

### Acceptance criteria

Investigate if RAM usage can be improved with mmap or unnecessary threading can be simplified as noted (i.e. removal of unnecessary spawn).

### Where to begin

https://github.com/filecoin-project/rust-fil-proofs/pull/1422

cryptonemo:
The first part of this has been resolved by https://github.com/filecoin-project/rust-fil-proofs/pull/1427

Next up is experimenting with potentially improved performance.

cryptonemo:
Closing due to #1436 for the initial investigation

PR #1436, opened 2021-03-18
Remove deprecated code warnings and minor perf improvements

### LINKED ISSUE #1435: Clean up deprecation warnings and measure performance in memory_handling code
### Description

Our current master branch warns about deprecated code in the memory_handling code.  This needs to be replaced primarily.

I'm combining this issue with measuring the effectiveness of that code.

### Acceptance criteria

Update the code.  Get some measurements.

cryptonemo:
Aims to resolve #1435 and investigate #1426

### DIFF
--- a/storage-proofs-porep/src/stacked/vanilla/proof.rs
+++ b/storage-proofs-porep/src/stacked/vanilla/proof.rs
@@ -106,8 +106,8 @@ impl<'a, Tree: 'static + MerkleTreeTrait, G: 'static + Hasher> StackedDrg<'a, Tr
             columns.extend(
                 parents
                     .into_par_iter()
-                    .map(|parent| t_aux.column(parent).expect("failed to get parent column"))
-                    .collect::<Vec<Column<Tree::Hasher>>>(),
+                    .map(|parent| t_aux.column(parent))
+                    .collect::<Result<Vec<Column<Tree::Hasher>>>>()?
             );
 
             debug_assert!(columns.len() == base_degree);
--- a/storage-proofs-porep/src/stacked/vanilla/proof.rs
+++ b/storage-proofs-porep/src/stacked/vanilla/proof.rs
@@ -103,9 +103,12 @@ impl<'a, Tree: 'static + MerkleTreeTrait, G: 'static + Hasher> StackedDrg<'a, Tr
             let mut parents = vec![0; base_degree];
             graph.base_parents(x, &mut parents)?;
 
-            for parent in &parents {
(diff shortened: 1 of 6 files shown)


storage-proofs-porep/src/stacked/vanilla/proof.rs
```diff
@@ -103,9 +103,12 @@ impl<'a, Tree: 'static + MerkleTreeTrait, G: 'static + Hasher> StackedDrg<'a, Tr
...
             graph.base_parents(x, &mut parents)?;
 
-            for parent in &parents {
-                columns.push(t_aux.column(*parent)?);
-            }
+            columns.extend(
+                parents
+                    .into_par_iter()
+                    .map(|parent| t_aux.column(parent).expect("failed to get parent column"))
+                    .collect::<Vec<Column<Tree::Hasher>>>(),
```
dignifiedquire, line 110: you should be able to collect into `Result<Vec>>` allowing you to acutally handle the error, instead of `expect`ing

cryptonemo: Maybe I'm missing something, but it looks like not what I want.  The return type is `Result<Vec<Column<Tree::Hasher>>>`, but w/o the `expect`, we'd be returning `Vec<Result<Column<Tree::Hasher>>>`.

cryptonemo:
Rebased against master

storage-proofs-porep/src/stacked/vanilla/proof.rs
dignifiedquire:
```rust
parents
  .into_par_iter()
  .map(|parent| t_aux.column(parent))
  .collect::<Result<Vec<Column<Tree::Hasher>>, _>>()?
```

is what I have in mind, to avoid  the `expect`

cryptonemo: Hrm, that makes sense, I think I had simply noted the difference of return type if the expect was removed, not changing the collect type.  Is there a difference between:

```
.collect::<Result<Vec<Column<Tree::Hasher>>, _>>()?
```

and

```
.collect::<Result<Vec<Column<Tree::Hasher>>>()?
```

They both appear to work.

cryptonemo: > They both appear to work.

Nevermind, just seems to be the type of error, which is implied either way (explicitly or implicitly).

dignifiedquire (approved):

Issue #1442, opened 2021-04-01
Unable to compile sha2raw:2.1.0 on ARM platform


dbshch:
### Description

The commit #1394 introduced a problem in the sha2raw crate:
On ARM platform, the mod sha256_intrinsics can't be imported.

error[E0432]: unresolved import crate::sha256_intrinsics
--> sha2raw/src/platform.rs:1:13
|
1 | use crate::{sha256_intrinsics, sha256_utils};
| ^^^^^^^^^^^^^^^^^ no sha256_intrinsics in the root

cryptonemo:
Build issue resolved by https://github.com/filecoin-project/rust-fil-proofs/pull/1444

cryptonemo:
Resolved by #1457

Issue #1353, opened 2020-11-09
Setup CI for aarch64


dignifiedquire:
Unfortunately getting native arm is hard in the current CI systems, but we can use qemu to at least figure some things out

PR #1444, opened 2021-04-07
feat: allow M1 building by removing asm sha on aarch64

cryptonemo:
For people that need this now -- here are pretty minimal changes to get proofs building on the Apple M1 platform.  This isn't ideal because it removes asm usage on aarch64, but works for now.  Some tests fail, so patching those up is next.

Build as follows:

```
LIBRARY_PATH=/opt/homebrew/lib cargo +nightly build --release --all
```

### DIFF
--- a/fr32/src/reader.rs
+++ b/fr32/src/reader.rs
@@ -19,6 +19,7 @@ const NUM_U128S_PER_BLOCK: usize = NUM_BYTES_OUT_BLOCK / size_of::<u128>();
 const MASK_SKIP_HIGH_2: u128 = 0b0011_1111_1111_1111_1111_1111_1111_1111_1111_1111_1111_1111_1111_1111_1111_1111_1111_1111_1111_1111_1111_1111_1111_1111_1111_1111_1111_1111_1111_1111_1111_1111;
 
 /// An `io::Reader` that converts unpadded input into valid `Fr32` padded output.
+#[repr(align(16))]
 pub struct Fr32Reader<R> {
     /// The source being padded.
     source: R,
@@ -65,7 +66,14 @@ impl<R: Read> Fr32Reader<R> {
 
     /// Processes a single block in in_buffer, writing the result to out_buffer.
     fn process_block(&mut self) {
-        let in_buffer: &[u128] = self.in_buffer.as_slice_of::<u128>().unwrap();
+        let in_buffer: &[u128] = {
+            #[cfg(target_arch = "aarch64")]
+            unsafe {
+                &mut (*(&self.in_buffer as *const [u8] as *mut [u128]))
+            }
+            #[cfg(not(target_arch = "aarch64"))]
+            self.in_buffer.as_slice_of::<u128>().unwrap()
+        };
         let out = &mut self.out_buffer;
 
         // 0..254
--- a/fr32/src/reader.rs
+++ b/fr32/src/reader.rs
@@ -18,15 +18,17 @@ const NUM_U128S_PER_BLOCK: usize = NUM_BYTES_OUT_BLOCK / size_of::<u128>();
 
 const MASK_SKIP_HIGH_2: u128 = 0b0011_1111_1111_1111_1111_1111_1111_1111_1111_1111_1111_1111_1111_1111_1111_1111_1111_1111_1111_1111_1111_1111_1111_1111_1111_1111_1111_1111_1111_1111_1111_1111;
 
-/// An `io::Reader` that converts unpadded input into valid `Fr32` padded output.
 #[repr(align(16))]
+struct AlignedBuffer([u8; NUM_BYTES_IN_BLOCK + 1]);
+
+/// An `io::Reader` that converts unpadded input into valid `Fr32` padded output.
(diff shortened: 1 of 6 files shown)


cryptonemo:
Builds have not yet been tested through `api` and `ffi`.

cryptonemo:
That reminds me, it's about time to setup the aarch CI for the builds, to perhaps see the failures.

Related to https://github.com/filecoin-project/rust-fil-proofs/issues/1353

cryptonemo:
Resolves #1442

dignifiedquire:
ideally instead of disabling it, we would port the working code from the official sha2 impl for aarch64 over

cryptonemo:
All tests are passing on the M1 locally ... albeit ... very ... slow ... ly ...

cryptonemo:
Good news: CI has been enabled for aarch64.

Bad news: Testing `--all` with blst takes ~1 hour 15 minutes, and with pairing takes ~1 hour 45 minutes (both using the `arm.large` instances).

While I'd like to know that tests pass on aarch64 on a regular basis, we may have to consider constraining the tests run to a much smaller subset (and then perhaps drop back to the `arm.medium` instances).  If anyone has thoughts, feel free to chime in.  Worst case, we could start with what we have now and scale back over time as needed.

cryptonemo:
Hah, nevermind my previous comment.  So both configurations pass in ~20 minutes after enabling `--release`.  The last thing I'd like to try is find if the `arm.medium` instances are good enough for that.

cryptonemo:
~20 minutes on `arm.large` and ~40 minute test runs on `arm.medium` shows a linear speed-up with increasing core counts.  For now, ~40 minutes on the presumably cheaper `arm.medium` instances is acceptable.

cryptonemo:
Resolves #1353

fr32/src/reader.rs
```diff
@@ -65,7 +66,14 @@ impl<R: Read> Fr32Reader<R> {
 
     /// Processes a single block in in_buffer, writing the result to out_buffer.
     fn process_block(&mut self) {
-        let in_buffer: &[u128] = self.in_buffer.as_slice_of::<u128>().unwrap();
+        let in_buffer: &[u128] = {
+            #[cfg(target_arch = "aarch64")]
+            unsafe {
```
dignifiedquire, line 71: should have a comment why this is safe to do

ribasushi:
>  This isn't ideal because it removes asm usage on aarch64

I run some production infra on ARM in AWS, would be great if you can give me a heads up if this ends up shipping in a stable release, so I can migrate to x86.

sha2raw/Cargo.toml
```diff
@@ -30,8 +30,12 @@ sha2 = "0.9.1"
 rand = "0.7.3"
 rand_xorshift = "0.2.0"
 
-[features]
+[target."cfg(target_arch = \"aarch64\")".features]
+default = []
+[target."cfg(not(target_arch = \"aarch64\"))".features]
```
dignifiedquire, line 35: still needs restriction to `target_os = "macos"`

fr32/src/reader.rs
```diff
@@ -19,6 +19,7 @@ const NUM_U128S_PER_BLOCK: usize = NUM_BYTES_OUT_BLOCK / size_of::<u128>();
 const MASK_SKIP_HIGH_2: u128 = 0b0011_1111_1111_1111_1111_1111_1111_1111_1111_1111_1111_1111_1111_1111_1111_1111_1111_1111_1111_1111_1111_1111_1111_1111_1111_1111_1111_1111_1111_1111_1111_1111;
 
 /// An `io::Reader` that converts unpadded input into valid `Fr32` padded output.
+#[repr(align(16))]
```
dignifiedquire, line 22: Blocking: I believe this only aligns the struct, so to be able to rely on this, the `in_buffer` must be the first field in the struct and we need to use `repr(C)`  to avoid any reordering of the fields. 
Easier solution is to make sth like

```rust
#[repr(align(16))]
struct AlignedBuffer([u8; NUM_BYTES_IN_BLOCK + 1 ]);
```

and use that

cryptonemo: Ah, nice, thanks, will look at the options there.

cryptonemo:
NOTE: This PR is on hold because it disables `asm` for `aarch64` across the board and not for just Apple M1 machines where it isn't supported yet.  The issue is that the `asm` feature of `sha2` is used throughout our code as well as dependencies and properly disabling it for this particular arch/os appears troublesome.  Unfortunately, until we can find a reasonable solution for this, this can't be merged.

cryptonemo:
Related work: https://github.com/RustCrypto/hashes/pull/261

cryptonemo:
Darwin CI failure related to https://github.com/filecoin-project/rust-fil-proofs/pull/1454 and #1450 (and unrelated to this work)

cryptonemo:
Closing in favor of filecoin-project/rust-fil-proofs#1455 and filecoin-project/rust-fil-proofs#1457

PR #1445, opened 2021-04-08
z-6.1.0

long568:
Free allocation of CPU packets


cryptonemo:
This appears to break/remove the core selection analysis code.  It seems to instead select arbitrary cores above some specified environment variable numbered core value.

I would think it could be useful to have something like this for systems where the topology isn't matching up with what's selected, but more likely we'd benefit from a manual way to specify entire core groups for those cases (where the user knows what makes sense for their hardware and can override failed detection).

storojs72:
It seems that this change is a bit stale. Also CI tests are not passed. I'm not sure what is the adopted policy is followed regarding stale PRs but usually it is better to close PRs with abandoned work (I see that last activity was in April 2021)

PR #1452, opened 2021-04-29
fix: disable multicore sdr on os x where it's not supported

cryptonemo:


cryptonemo:
Is anyone able to verify that this is the case?  Has multicore sdr been tested on OS X?

cryptonemo:
> Is anyone able to verify that this is the case? Has multicore sdr been tested on OS X?

I do see a panic on our darwin CI machine when enabled, but am curious if this was ever intended to work on OS X (I realize it's not the ideal runtime target, just curious).

Interestingly, `M1` appears to work ok as-is, so I may update this PR to leave that enabled.

dignifiedquire:
SDR multicore should work on macs, I don' think we should generally disable it. There might be issues with the specific hardware on CI

cryptonemo:
> SDR multicore should work on macs, I don' think we should generally disable it. There might be issues with the specific hardware on CI

Alternatively, we could remove the CI darwin run as required as it's been more or less consistently failing lately (it used to be hit or miss).  It may actually have to do with what else is running on CI at the time, not sure.

cryptonemo:
Closed in favor of https://github.com/filecoin-project/rust-fil-proofs/pull/1454

Issue #1450, opened 2021-04-28
CI failures for darwin


cryptonemo:
### Description

I've been seeing this intermittently in proofs CI and it should be resolved.

```
     Running `/Users/distiller/crate/target/release/deps/storage_proofs_porep-3eae6b0d3e017065 --nocapture`

running 14 tests
test stacked::vanilla::challenges::test::test_calculate_fixed_challenges ... ok
thread 'stacked::vanilla::cores::tests::test_cores' panicked at 'attempt to calculate the remainder with a divisor of zero', storage-proofs-porep/src/stacked/vanilla/cores.rs:149:19
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
test stacked::vanilla::cores::tests::test_cores ... FAILED
thread 'stacked::vanilla::create_label::multi::tests::test_create_labels' panicked at 'poisoned lock: "PoisonError { inner: .. }"', storage-proofs-porep/src/stacked/vanilla/cores.rs:126:32
test stacked::vanilla::create_label::multi::tests::test_create_labels ... FAILED
test stacked::vanilla::cache::tests::test_read_full_range ... ok
[storage-proofs-porep/src/stacked/vanilla/graph.rs:552] &porep_id = [
    3,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
]
[storage-proofs-porep/src/stacked/vanilla/graph.rs:552] &nodes = 1073741824
[storage-proofs-porep/src/stacked/vanilla/graph.rs:552] &expect_pathological = true
test stacked::vanilla::cache::tests::test_read_partial_range ... ok
test stacked::vanilla::graph::tests::test_high_parent_bits ... ok
[storage-proofs-porep/src/stacked/vanilla/graph.rs:552] &porep_id = [
    4,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
]
[storage-proofs-porep/src/stacked/vanilla/graph.rs:552] &nodes = 2147483648
[...]

PR #1454, opened 2021-05-04
fix: update initial value for cache_count

cryptonemo:


cryptonemo:
The goal is to help resolve the OS X CI failures that have a topology that doesn't work under the assumptions of this code.

cryptonemo:
Resolves #1450

dignifiedquire (approved):
looks working

PR #1457, opened 2021-05-06
feat: update sha2-asm dependency version

cryptonemo:
fix: update call based on updated version

### DIFF
--- a/sha2raw/src/platform.rs
+++ b/sha2raw/src/platform.rs
@@ -81,7 +81,7 @@ impl Implementation {
                 for block in blocks.chunks(2) {
                     buffer[..32].copy_from_slice(&block[0]);
                     buffer[32..].copy_from_slice(&block[1]);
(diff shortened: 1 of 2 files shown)


sha2raw/src/platform.rs
```diff
@@ -81,7 +81,7 @@ impl Implementation {
                 for block in blocks.chunks(2) {
                     buffer[..32].copy_from_slice(&block[0]);
                     buffer[32..].copy_from_slice(&block[1]);
-                    sha2_asm::compress256(state, &buffer);
+                    sha2_asm::compress256(state, &vec![buffer]);
```
cryptonemo, line 84: @dignifiedquire What's the better non-allocating way to do this?

cryptonemo: Hah! Nevermind, clippy for the win

cryptonemo: Also, I thought that's what I first tried, but apparently not ...

PR #1460, opened 2021-05-13
WIP: scheduler integration

neithanmo:
## This PR is a work-in-progress and must not be used in production

Integrates the scheduler into the current code trying to keep untouched as much code as possible:
some test fails, meaning something is wrong, so this PR is just for getting an insight about the problem source, do not use in production yet.

### DIFF
--- a/storage-proofs-core/src/compound_proof.rs
+++ b/storage-proofs-core/src/compound_proof.rs
@@ -5,7 +5,7 @@ use bellperson::{
         aggregate::{
             aggregate_proofs, verify_aggregate_proof, AggregateProof, ProverSRS, VerifierSRS,
         },
-        create_random_proof_batch, create_random_proof_batch_in_priority, verify_proofs_batch,
+        create_random_proof_batch_with_type, verify_proofs_batch, BellTaskType,
         PreparedVerifyingKey,
     },
     Circuit,
@@ -77,12 +77,15 @@ where
         }
     }
 
-    /// prove is equivalent to ProofScheme::prove.
-    fn prove<'b>(
+    // helper function it is almost the same as prove
+    // but with a taskType parameter which is used in bellperson to define
+    // the priority of the task and handling it accordingly.
+    fn prove_with_type<'b>(
         pub_params: &PublicParams<'a, S>,
         pub_in: &S::PublicInputs,
         priv_in: &S::PrivateInputs,
         groth_params: &'b groth16::MappedParameters<Bls12>,
+        post_type: Option<BellTaskType>,
     ) -> Result<MultiProof<'b>> {
         let partition_count = Self::partition_count(pub_params);
 
@@ -105,11 +108,28 @@ where
             vanilla_proofs,
             &pub_params.vanilla_params,
             groth_params,
-            pub_params.priority,
+            post_type,
         )?;
         info!("snark_proof:finish");
 
         Ok(MultiProof::new(groth_proofs, &groth_params.pvk))
+    }
+
+    /// prove is equivalent to ProofScheme::prove.
+    fn prove<'b>(
+        pub_params: &PublicParams<'a, S>,
+        pub_in: &S::PublicInputs,
+        priv_in: &S::PrivateInputs,
+        groth_params: &'b groth16::MappedParameters<Bls12>,
+    ) -> Result<MultiProof<'b>> {
+        // Is it ok to assign a default value of MerkleTree to this?
+        Self::prove_with_type(
+            pub_params,
+            pub_in,
+            priv_in,
+            groth_params,
+            Some(BellTaskType::MerkleTree),
+        )
     }
 
     fn prove_with_vanilla<'b>(
@@ -117,6 +137,7 @@ where
         pub_in: &S::PublicInputs,
         vanilla_proofs: Vec<S::Proof>,
         groth_params: &'b groth16::MappedParameters<Bls12>,
+        post_type: Option<BellTaskType>,
     ) -> Result<MultiProof<'b>> {
         let partition_count = Self::partition_count(pub_params);
 
@@ -129,7 +150,7 @@ where
             vanilla_proofs,
             &pub_params.vanilla_params,
             groth_params,
-            pub_params.priority,
+            post_type,
         )?;
         info!("snark_proof:finish");
 
@@ -234,7 +255,7 @@ where
         vanilla_proofs: Vec<S::Proof>,
         pub_params: &S::PublicParams,
         groth_params: &groth16::MappedParameters<Bls12>,
-        priority: bool,
+        post_type: Option<BellTaskType>,
     ) -> Result<Vec<groth16::Proof<Bls12>>> {
         let mut rng = OsRng;
         ensure!(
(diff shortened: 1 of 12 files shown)


storage-proofs-porep/src/stacked/vanilla/proof.rs
```diff
@@ -554,53 +556,86 @@ impl<'a, Tree: 'static + MerkleTreeTrait, G: 'static + Hasher> StackedDrg<'a, Tr
...
-                                .expect("failed to add final columns");
-                            trace!(
-                                "base data len {}, tree data len {}",
-                                base_data.len(),
-                                tree_data.len()
-                            );
+                                is_final = last;
+                            } else {
+                                // If we get here, this is a final column: build a sub-tree.
+                                if let Some((base_data, tree_data)) = builder.build_next()? {
```
porcuquine, line 609: It looks like you are building a sub-tree without having yet added the final batch of columns. I think you need to add the columns first, then start in on the building.

This brings up the point that with the tentative changes to neptune being used here, the API has been made more brittle. This mistake would not have been possible using the old API. You can probably fix as described above, but I would recommend also modifying `neptune` so it becomes an error to call `build_next()` before the final columns have been added. This would actually add some error-checking (enforcing that trees must be full) that was missing before.

If adding that check causes this code to fail at runtime, you will know it's working. Hopefully then fixing the problem here will stop that error, giving some confidence in the changes on both sides of the API boundary.

storage-proofs-porep/src/stacked/vanilla/proof.rs
```diff
@@ -928,43 +963,69 @@ impl<'a, Tree: 'static + MerkleTreeTrait, G: 'static + Hasher> StackedDrg<'a, Tr
...
+                            builder.add_leaves(&encoded)?;
+                            is_final = last;
+                        } else {
+                            // If we get here, this is a final leaf batch: build a sub-tree.
+                            info!(
+                                "building base tree_r_last with GPU {}/{}",
+                                i + 1,
+                                tree_count
+                            );
+                            if let Some((_, tree_data)) = builder.build_next()? {
```
porcuquine, line 1014: The comment from above applies here too, except instead of a missing call to `add_columns()`, there is a missing call to `add_leaves()`. The same fix and same recommendation for a corresponding enhancement to the new `neptune` API apply. Making the `neptune` change will mitigate the increased chance for mistake (as seen here) introduced by removing the need for the caller to explicitly add the final leaves/columns.

porcuquine (changes requested):
I noticed what appear to be changes in results introduced by not fully replicating the behavior of `add_final_columns()` and `add_final_leaves()` when replacing them with new API methods. It's plausible that this is the cause of some test failures.

storage-proofs-porep/src/stacked/vanilla/proof.rs
porcuquine: After looking at this again, I think the actual logic is correct here. However, the variable name (`is_final`) and comments are now inaccurate, leading to the confusion.

Consider that the comment on line 608 suggests possession of a 'final' column. This is a first-class concept inherited from the neptune API (that there be a 'final' batch of columns). However, the final column(s) were actually received and added on the _previous_ invocation of  `call`.

I would change `is_final` to something like `done_adding_columns` to make the meaning very clear. The comment should probably be split into two, one describing what happens in each branch of the conditional.

It may still make sense to add the check to neptune, to rule out the prior possibility, but I don't _think_ the code is 'wrong' after all. That change would make it impossible to create partial trees with implicit zero-padding, but I'm not sure we need to preserve that possibility (though it would technically be a breaking API change).

storage-proofs-porep/src/stacked/vanilla/proof.rs
porcuquine: The above correction applies here too: https://github.com/filecoin-project/rust-fil-proofs/pull/1460/files#r631920821

neithanmo:
i updated this with the integration rust-fil-proofs -> bellperson.  seems there are some errors when starting the scheduler in CI. . i just updated the scheduler to show a more verbose error message. 
in regards to the error during the tests, I will explain a bit more in the channel.

porcuquine (changes requested):

storage-proofs-porep/src/stacked/vanilla/proof_ext.rs
```diff
@@ -0,0 +1,67 @@
...
+
+    pub(crate) fn build(&mut self) -> Result<(), Error> {
+        let requirements = {
+            let resouce_req = ResourceReq {
+                resource: ResourceType::Gpu(ResourceMemory::All),
+                quantity: 1,
+                preemptible: self.num_iter > 1,
+            };
+            let task_req = TaskReqBuilder::new()
+                .with_task_type(TaskType::MerkleProof)
```
porcuquine, line 36: Assuming this type is meant to describe Merkle tree building, it is misnamed. Merkle proofs are later generated over these trees, but the act of building them shouldn't reference 'proof', since that is confusing.

storage-proofs-core/src/compound_proof.rs
```diff
@@ -77,12 +77,15 @@ where
...
-    fn prove<'b>(
+    // helper function it is almost the same as prove
+    // but with a taskType parameter which is used in bellperson to define
+    // the priority of the task and handling it accordingly.
+    fn prove_with_type<'b>(
         pub_params: &PublicParams<'a, S>,
         pub_in: &S::PublicInputs,
         priv_in: &S::PrivateInputs,
         groth_params: &'b groth16::MappedParameters<Bls12>,
+        post_type: Option<BellTaskType>,
```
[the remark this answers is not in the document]
vmx: (unrelated to @porcuquine's comment) It looks like you always pass in a type, so I don't think it needs to be wrapped in a `Option`.

storage-proofs-porep/src/stacked/vanilla/proof_ext.rs
neithanmo: I am open to set a more appropriate name here. also not sure if it should include the word __Merkle__ as what I have seen is that there are operations that not only use neptune but bellperson.

storage-proofs-core/src/compound_proof.rs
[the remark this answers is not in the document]
neithanmo: > `CompoundProof` is an abstraction that doesn't (and shouldn't) know anything whatsoever about specific proof types. So mention of `post_type` here, `MerkleTree` later, etc. is by definition inappropriate. It might be that `post_type` is just misnamed (will look more closely) — but the point stands: if we are referencing specific proof types of any kind here then something is wrong.

@porcuquine  I did not like the idea of refactoring that trait neither, as it seems to be a super Trait from which other traits inherit. but, at the moment it looked like the cleaning way to modify and pass the information as a task_type(which at the moment encodes deadlines and possible exclusive resources) to the scheduler, without having to refactor many other places.

@vmx Initially, it was defined as an optional param, I set it to MerkeProof, but I am not sure if that type should encompass all the processes that use the GPU but that are not Winning/Window post. It is important because all those processes will end up using the GPU assigned to MerkeTree operations.

PR #1466, opened 2021-06-02
fix: remove build conditional (not needed; not clear it was used)

cryptonemo:
I recall an error with this kind of conditional not actually working (but silently):

https://github.com/filecoin-project/rust-fil-proofs/blame/master/storage-proofs-porep/Cargo.toml#L43

I believe it's no longer needed in any case with asm support for all supported platforms in sha2 now available.


dignifiedquire:
this is about enabling asm by default, not because it is not supported, so this would be changing the default behavior we have right now afaik

PR #1474, opened 2021-06-16
feat: add a srs key loading bench

cryptonemo:
feat: separate aggregation benches from preprocessing

PR #1475, opened 2021-06-17
fix: use correct aggregate proof serialization format

cryptonemo:


nikkolasg (approved):

porcuquine (approved):
This looks straightforward. On the assumption that this is directly exercised by tests which show that these functions round-trip as expected, this seems fine to me.

dignifiedquire:
@cryptonemo can we add some tests to check that the sizes match exactly hard coded values, to avoid surprises.

cryptonemo:
> @cryptonemo can we add some tests to check that the sizes match exactly hard coded values, to avoid surprises.

For lack of a better place (i.e. a separate PR), I added it here: https://github.com/filecoin-project/rust-fil-proofs/pull/1474/commits/336e2fde39750ef3f4392a062e573bed7ed3c896

PR #1477, opened 2021-07-02
Fix: correct usage of BitMask in multicore sdr

qy3u:
This PR contains two commits:
1. Use the correct node index to set bpm. The prev node is the first node of parents not the sixth: https://github.com/filecoin-project/rust-fil-proofs/blob/00cb80440d14dd760805c14db597a306d4433fb2/storage-proofs-core/src/drgraph.rs#L168
2. Unset the bpm after using it. The current implemention has never unsetted it so it will always read all base parents in the main hash loop.


storage-proofs-porep/src/stacked/vanilla/create_label/multi.rs
```diff
@@ -77,13 +77,13 @@ fn fill_buffer(
     // Node 5 (prev node) will always be missing, and there tend to be
     // frequent close references.
     if cur_node > MIN_BASE_PARENT_NODE {
-        // Mark base parent 5 as missing
+        // Mark base parent 0 as missing
         // base_parent_missing.set_all(0x20);
-        base_parent_missing.set(5);
+        base_parent_missing.set(0);
```
porcuquine, line 82: Good catch. It looks like when the graph was changed (conditionally, depending on version), this code was not updated to also be conditional. So the old graph is always being used here. To make this fully correct now, it should be conditional here now too. The current formulation for the decision can be seen here: https://github.com/filecoin-project/rust-fil-proofs/blob/00cb80440d14dd760805c14db597a306d4433fb2/storage-proofs-core/src/drgraph.rs#L166-L169.

storage-proofs-porep/src/stacked/vanilla/create_label/multi.rs
```diff
@@ -77,13 +77,13 @@ fn fill_buffer(
...
 
         // Skip the last base parent - it always points to the preceding node,
         // which we know is not ready and will be filled in the main loop
-        for k in 0..BASE_DEGREE - 1 {
+        for k in 1..BASE_DEGREE {
```
porcuquine, line 86: This will need to be updated too. Probably the right thing is to have a single match which assigns to the relevant indices then uses those assignments elsewhere — similar to the code linked above.

porcuquine (changes requested):
Good catches. I _think_ this is good, but should probably go a little further (as noted in comments).

@cryptonemo We should make sure to run some full lifecycle tests on this (when done) to ensure there are no surprises.

storage-proofs-porep/src/stacked/vanilla/create_label/multi.rs
qy3u: You are right! The current implemention doesn't has a good compatibility with different versions of the graph (which it should be).I'll try to improve this PR to solve this quesion.

qy3u:
Hi @porcuquine. This PR has updated.Now,it stores the api_version in `ParentCache` and `CacheReader`, then it matchs the version stored to get corresponding indexs.

cryptonemo:
Thanks @qy3u -- I'll look closer and get some testing going on this early next week.

storage-proofs-porep/src/stacked/vanilla/memory_handling.rs
```diff
@@ -20,6 +22,7 @@ pub struct CacheReader<T> {
     window_size: usize,
     cursor: IncrementingCursor,
     consumer: AtomicU64,
+    pub api_version: ApiVersion,
```
dignifiedquire, line 25: lets make this private and add a method to get it out.

storage-proofs-porep/src/stacked/vanilla/utils.rs
```diff
@@ -36,6 +36,7 @@ impl<'a, T> UnsafeSlice<'a, T> {
     }
 
     #[inline]
+    #[allow(dead_code)]
```
dignifiedquire, line 39: lets delete this if it is never used

dignifiedquire (commented):
@cryptonemo can you run this for some performance checks, making sure we don't have any accidental regressions

cryptonemo:
> @cryptonemo can you run this for some performance checks, making sure we don't have any accidental regressions

Yes, this is on the list!

cryptonemo:
Still have wider testing to do, but the first 32GiB seal lifecycle test completed in 320m.  Compared to master taking 347m on the same hardware.  Testing again to see if this is consistent or something else.

porcuquine (approved):
The update appears to accomplish what I requested, and as far as I can see this all seems good.

This should not affect verification of historical proofs, even though the changes do relate to the previous upgrade. This PR is just propagating some changes from that upgrade which were not reflected in sealing.

The large performance impact is likely due to resetting `bpm` between node iterations.

In addition to the tests for performance regression, we should make sure to repeatedly observe that sealing completes consistently and correctly (i.e. valid PoRep, and subsequently PoSt — though I am less concerned about PoSt) are produced. I don't have specific reason to think something will have gone wrong, but there _is_ an operational change to how the data is being processed, and if for whatever reason that is introducing an error, this is where it would show up. It could be the case that such an error be non-deterministic. Given that the performance change results in much faster sealing, we should be skeptical in principle — even though, as noted, I believe this is probably a good and safe change.

Therefore, I approve, but contingent on a clean history of multiple runs with no failures. If we were to see any failures in testing, we should not just disregard and re-run the tests. We are not looking for presence of success but rather absence of failure.

Good work.

cryptonemo:
@qy3u Any chance you can back out the merge I did to bring this up to latest master?  I'd prefer to keep your original 3 commits and have those rebased against the latest `master` branch and pushed back here.  If you're able to do that today, I'd appreciate it -- otherwise, I'll keep what we have after that merge and finish testing on it.

cryptonemo:
This appears to test well, as-is.

cryptonemo (approved):

cryptonemo:
After merging this, CI tested well.  However, after cutting a new release with it, a CI test failed that exercised this code path (seal_lifecycle 2k).  Based on that alone, it's possible this will be reverted in our next shipped version.

PR #1480, opened 2021-07-09
fix: update dependencies

vmx:
Update to the latest versions of rust-gpu-tools, which is now using
`opencl3` instead of `ocl`. Also update to the latest version of
neptune, which got its API a bit cleaned up and doesn't need the
`BatcherType` anymore, but used that batchers directly.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -11,3 +11,8 @@ members = [
   "sha2raw",
   "filecoin-hashers",
 ]
+
+[patch.crates-io]
+rust-gpu-tools = { git = "https://github.com/filecoin-project/rust-gpu-tools", branch = "master" }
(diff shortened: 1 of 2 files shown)


cryptonemo:
Oops, I didn't realize this PR existed until after I updated my PR here: https://github.com/filecoin-project/rust-fil-proofs/pull/1482

I think this one will be closed in favor of that one.

vmx:
Nice things about having a separate commit/PR is that also other people that upgrade to never versions (perhaps for the scheduler work?) can easily see what needs to be done based on the Git history. Though, whatever is easiest.

vmx:
I just saw that the other PR doesn't do much more, so nevermind :)

cryptonemo:
Closed in favor of #1482

PR #1481, opened 2021-07-13
feat: use OnceCell for srs key caches to avoid duplicate generation

### LINKED ISSUE #1479: Repeat generation of aggregation keys
### Description

Currently, if multiple threads try to verify aggregate proofs with new parameters at the same time, we'll try to compute the keys multiple times (last one wins).

Ideally, we'd compute the keys once.

### Acceptance criteria

I can try to verify a thousand aggregate proofs at once without waiting a day or two (`(100s * 1000)/3600 = ~28 hours`).

### Risks + pitfalls

Extra complexity in the caches could introduce bugs. We also probably don't want to rely on nightly, which means either rolling our own "once" or using an external crate.

### Where to begin

Use something like https://doc.rust-lang.org/std/lazy/struct.SyncOnceCell.html (or https://crates.io/crates/conquer-once) or even just a mutex to define a per-key locked "cells" for every entry in our caches.

cryptonemo:
Resolves #1479

### DIFF
--- a/filecoin-proofs/src/caches.rs
+++ b/filecoin-proofs/src/caches.rs
@@ -29,6 +29,10 @@ type Cache<G> = HashMap<String, Arc<G>>;
 type GrothMemCache = Cache<Bls12GrothParams>;
 type VerifyingKeyMemCache = Cache<Bls12PreparedVerifyingKey>;
 
+const FIP0013_MIN_SNARKS: usize = 64;
+const FIP0013_MAX_SNARKS: usize = 8192;
+const PROOFS_TESTS_MIN_SNARKS: usize = FIP0013_MIN_SNARKS >> 5;
+
 const SRS_IDENTIFIER: &str = "srs-key";
 const SRS_VERIFIER_IDENTIFIER: &str = "srs-verifying-key";
 
@@ -52,7 +56,7 @@ pub struct SRSCache<G> {
 impl<G> SRSCache<G> {
     pub fn with_defaults(identifier: &str) -> Self {
         let mut data = HashMap::new();
-        let mut num_proofs_to_aggregate = 2;
+        let mut num_proofs_to_aggregate = PROOFS_TESTS_MIN_SNARKS;
 
         loop {
             for sector_size in &PUBLISHED_SECTOR_SIZES {
@@ -65,7 +69,7 @@ impl<G> SRSCache<G> {
             }
 
             num_proofs_to_aggregate <<= 1;
-            if num_proofs_to_aggregate > 8192 {
+            if num_proofs_to_aggregate > FIP0013_MAX_SNARKS {
                 break;
             }
         }
--- a/filecoin-proofs/src/caches.rs
+++ b/filecoin-proofs/src/caches.rs
@@ -49,15 +49,17 @@ lazy_static! {
         SRSCache::with_defaults(SRS_VERIFIER_IDENTIFIER);
 }
 
-// We have a separate SRSCache type for srs keys since they are cached
-// differently (as a hashmap per type, keyed by identifier consisting
-// of sector size and pow2 num proofs to aggregate)
+/// We have a separate SRSCache type for srs keys since they are
+/// cached differently (as a hashmap per type, keyed by identifier
+/// consisting of sector size and pow2 num proofs to aggregate).
 #[derive(Debug, Default)]
 pub struct SRSCache<G> {
     data: HashMap<String, OnceCell<Arc<G>>>,
 }
 
 impl<G> SRSCache<G> {
+    /// Initializes the cache by pre-populating the internal map with
+    /// all supported keys that could be looked up at a later time.
     pub fn with_defaults(identifier: &str) -> Self {
         let mut data = HashMap::new();
         let mut num_proofs_to_aggregate = PROOFS_TESTS_MIN_SNARKS;
@@ -69,7 +71,7 @@ impl<G> SRSCache<G> {
                     sector_size, num_proofs_to_aggregate, identifier,
                 );
                 trace!("inserting placeholder srs key with hash key {}", key);
-                data.insert(key.to_string(), OnceCell::new());
+                data.insert(key, OnceCell::new());
             }
 
             num_proofs_to_aggregate <<= 1;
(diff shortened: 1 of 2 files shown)


filecoin-proofs/src/caches.rs
```diff
@@ -27,14 +28,65 @@ type Bls12VerifierSRSKey = groth16::aggregate::VerifierSRS<Bls12>;
...
+                let key = format!(
+                    "STACKED[{}-{}]-{}",
+                    sector_size, num_proofs_to_aggregate, identifier,
+                );
+                trace!("inserting placeholder srs key with hash key {}", key);
+                data.insert(key.to_string(), OnceCell::new());
+            }
+
+            num_proofs_to_aggregate <<= 1;
+            if num_proofs_to_aggregate > 8192 {
```
dignifiedquire, line 68: shouldn’t this be a constant?

dignifiedquire:
did you test if this does solve the issue originally reported in some form? y

filecoin-proofs/src/caches.rs
cryptonemo: Yes, good call.

cryptonemo:
> did you test if this does solve the issue originally reported in some form? y

Yes, I ran all of our usual tests, which exercised a number of the identifiers and ran a 32GiB lifecycle test.  A 64GiB lifecycle test has been running for a very long time and may not be able to complete, as I need the machine to test the scheduler, but I'll give it a bit more time (not more than all of today).

cryptonemo:
> did you test if this does solve the issue originally reported in some form? y

Testing a variation on https://github.com/filecoin-project/filecoin-ffi/pull/185 and modified it to not warm up at first -- this is the real test in some sense, and this testing is not yet complete.

cryptonemo:
I tested various levels of parallelism and this is confirmed working in all cases I have found.

The grepped logs below show 1616 total calls to the lookup that will either block or specialize, and then the single logged line inside of the generator doing the specialize:

```
nemo@code4:~/pl/filecoin-ffi/rust$ cat LOG2 | grep -c "generating or waiting on specialize"
1616

nemo@code4:~/pl/filecoin-ffi/rust$ cat LOG2 | grep "specializing"
2021-07-14T10:39:05.538 TRACE filecoin_proofs::caches > get_stacked_srs_verifier_key specializing STACKED[34359738368-16384]
```

filecoin-proofs/src/caches.rs
```diff
@@ -27,14 +28,74 @@ type Bls12VerifierSRSKey = groth16::aggregate::VerifierSRS<Bls12>;
...
     static ref VERIFYING_KEY_MEMORY_CACHE: Mutex<VerifyingKeyMemCache> = Default::default();
-    static ref SRS_KEY_MEMORY_CACHE: Mutex<SRSKeyMemCache> = Default::default();
-    static ref SRS_VERIFIER_KEY_MEMORY_CACHE: Mutex<SRSVerifierKeyMemCache> = Default::default();
+    static ref SRS_KEY_MEMORY_CACHE: SRSCache<Bls12ProverSRSKey> =
+        SRSCache::with_defaults(SRS_IDENTIFIER);
+    static ref SRS_VERIFIER_KEY_MEMORY_CACHE: SRSCache<Bls12VerifierSRSKey> =
+        SRSCache::with_defaults(SRS_VERIFIER_IDENTIFIER);
+}
+
+// We have a separate SRSCache type for srs keys since they are cached
```
dignifiedquire, line 52: should be `///`

filecoin-proofs/src/caches.rs
```diff
@@ -27,14 +28,74 @@ type Bls12VerifierSRSKey = groth16::aggregate::VerifierSRS<Bls12>;
...
+// differently (as a hashmap per type, keyed by identifier consisting
+// of sector size and pow2 num proofs to aggregate)
+#[derive(Debug, Default)]
+pub struct SRSCache<G> {
+    data: HashMap<String, OnceCell<Arc<G>>>,
+}
+
+impl<G> SRSCache<G> {
+    pub fn with_defaults(identifier: &str) -> Self {
```
dignifiedquire, line 61: can you add a doc comment please?

filecoin-proofs/src/caches.rs
```diff
@@ -27,14 +28,74 @@ type Bls12VerifierSRSKey = groth16::aggregate::VerifierSRS<Bls12>;
...
+        let mut num_proofs_to_aggregate = PROOFS_TESTS_MIN_SNARKS;
+
+        loop {
+            for sector_size in &PUBLISHED_SECTOR_SIZES {
+                let key = format!(
+                    "STACKED[{}-{}]-{}",
+                    sector_size, num_proofs_to_aggregate, identifier,
+                );
+                trace!("inserting placeholder srs key with hash key {}", key);
+                data.insert(key.to_string(), OnceCell::new());
```
dignifiedquire, line 72: I don't think `to_string()` is needed here

filecoin-proofs/src/caches.rs
```diff
@@ -27,14 +28,74 @@ type Bls12VerifierSRSKey = groth16::aggregate::VerifierSRS<Bls12>;
...
+    /// Returns `None` for non existent entries, `Some(v)` for existing ones, where `v` is either
+    /// the result of running `generator` or already existing one.
+    pub fn get_or_init<F>(&self, key: &str, generator: F) -> Option<&Arc<G>>
+    where
+        F: FnOnce() -> Result<G>,
+    {
+        if let Some(cell) = self.data.get(key) {
+            trace!("generating or waiting on specialize for {}", key);
+            return Some(
+                cell.get_or_init(|| Arc::new(generator().expect("SRS specialize failed"))),
```
dignifiedquire, line 93: In which cases can this fail? It seems risky to panic here

dignifiedquire: probably better to change the return value to `Result<Option<&Arc<G>>>>`, and propagate it

dignifiedquire (changes requested):

filecoin-proofs/src/caches.rs
cryptonemo: I'm having trouble propagating the error because it's inside of the generator closure.  It doesn't seem to be happy allowing me to return a Result.

dignifiedquire: this works for me

```rust
    pub fn get_or_init<F>(&self, key: &str, generator: F) -> Result<Option<&Arc<G>>>
    where
        F: FnOnce() -> Result<G>,
    {
        if let Some(cell) = self.data.get(key) {
            trace!("generating or waiting on specialize for {}", key);
            let x = cell.get_or_try_init(|| -> Result<Arc<G>> { Ok(Arc::new(generator()?)) })?;
            return Ok(Some(x));
        }

        Ok(None)
    }
```

cryptonemo: Ah, I see, thanks.  I had this:

```rust
    pub fn get_or_init<F>(&self, key: &str, generator: F) -> Result<Option<&Arc<G>>>
    where
        F: FnOnce() -> Result<G>,
    {   
        if let Some(cell) = self.data.get(key) {
            trace!("generating or waiting on specialize for {}", key);
            let result = cell.get_or_init(|| -> Result<Arc<G>> {
                let generated = generator()?;
                
                Ok(Arc::new(generated))
            });
            
            return Ok(Some(result));
        }
        
        Ok(None)
    }

```

Which errors with this:

```
error[E0271]: type mismatch resolving `<[closure@filecoin-proofs/src/caches.rs:94:43: 98:14] as FnOnce<()>>::Output == Arc<G>`
  --> filecoin-proofs/src/caches.rs:94:31
   |
94 |             let result = cell.get_or_init(|| -> Result<Arc<G>> {
   |                               ^^^^^^^^^^^ expected enum `std::result::Result`, found struct `Arc`
   |
   = note: expected enum `std::result::Result<Arc<_>, anyhow::Error>`
            found struct `Arc<_>`
```

dignifiedquire (approved):

PR #1483, opened 2021-07-23
fix: switch to yastl thread pool

dignifiedquire:
This avoids all deadlocks as so far observed. Also uses a more flexible channel
for sync

Depends on https://github.com/filecoin-project/bellperson/pull/204

### DIFF
--- a/storage-proofs-porep/src/stacked/vanilla/proof.rs
+++ b/storage-proofs-porep/src/stacked/vanilla/proof.rs
@@ -32,6 +32,7 @@ use storage_proofs_core::{
     settings::SETTINGS,
     util::{default_rows_to_discard, NODE_SIZE},
 };
+use yastl::Pool;
 
 use crate::{
     encode::{decode, encode},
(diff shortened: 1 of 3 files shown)


DrPeterVanNostrand:
Looks ok to me

storage-proofs-porep/src/stacked/vanilla/proof.rs
```diff
@@ -59,6 +60,8 @@ lazy_static! {
     /// It might be possible to relax this constraint, but in that case, only one builder
     /// should actually be active at any given time, so the mutex should still be used.
     static ref GPU_LOCK: Mutex<()> = Mutex::new(());
+
+    static ref THREAD_POOL: Pool = Pool::new(num_cpus::get());
```
DrPeterVanNostrand, line 64: Here you are using `num_cpus::get()`whereas in `bellperson` you are using an env-var `BELLMAN_NUM_CPUS` to get the number of CPU's, should these two values always agree?

DrPeterVanNostrand (commented):
Looks ok to me

storage-proofs-porep/src/stacked/vanilla/proof.rs
cryptonemo: No, I don't think so.  The bellman var is bellman specific, and that's ok.  If we had a proofs specific one, it might be different, but proofs generally targets max parallelism from the available cores.

cryptonemo (approved):

cryptonemo:
Final testing is happening, but in fact, I have a machine that reliably hangs now during tests without this fix.  Performance has so far been measured to be 'about the same' to 'a hair faster'.  When my last test completes, assuming perf hasn't degraded after the rebases to latest, I'll merge.

PR #1485, opened 2021-07-28
chore: improve 'PRODUCER NOT READY' log

Aloxaf:
I met some miner reports that there are too many (sometimes even millions of) "PRODUCER NOT READY" logs for each layer.

I think we can lower its log level.

### DIFF
--- a/storage-proofs-porep/src/stacked/vanilla/create_label/multi.rs
+++ b/storage-proofs-porep/src/stacked/vanilla/create_label/multi.rs
@@ -308,7 +308,7 @@ fn create_layer_labels(
 
         // Keep track of which node slot in the ring_buffer to use
         let mut cur_slot = 0;
-        let mut _count_not_ready = 0;
+        let mut count_not_ready = 0;
 
         // Calculate nodes 1 to n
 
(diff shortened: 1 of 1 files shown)


storage-proofs-porep/src/stacked/vanilla/create_label/multi.rs
```diff
@@ -322,9 +322,9 @@ fn create_layer_labels(
 
             while producer_val < i {
                 if !printed {
-                    debug!("PRODUCER NOT READY! {}", i);
+                    trace!("PRODUCER NOT READY! {}", i);
```
cryptonemo, line 325: I also see this logged thousands of time per run.  I'd argue this individual log line could be removed.  Perhaps a map of the `i` value to specific `count` may be needed if this information is important and it could be logged at the end along with the total count?

Aloxaf: A map of `i` to `count` may be too large to be logged. I agree with remove this line.

cryptonemo (approved):

Issue #1495, opened 2021-08-18
The ParentCache doesn't seem to be generated correctly in circleci


qy3u:
Hi @cryptonemo @dignifiedquire . I just saw the CI test failure on my previous PR https://github.com/filecoin-project/rust-fil-proofs/pull/1477 .

After checking the detail of the failure in https://circleci.com/api/v1.1/project/github/filecoin-project/rust-fil-proofs/74927/output/106/0?file=true&allocation-id=611544a9648d5f76edbe1b19-0-build%2F7693C729
I found the error message: 'Error: corrupted cache: /var/tmp/filecoin-parents/v28-sdr-parent-3f0eef38bb48af1f48ad65e14eb85b4ebfc167cec18cd81764f6d998836c9899.cache, expected at least 3584, got 0 bytes'

According to the message, it seems that the test failed because the `ParentCache` file doesn't generated correctly(or flush to disk) in circleci. Besides,I have checkout to the commit that contains my PR and run the test locally, here is my command: `FIL_PROOFS_USE_MULTICORE_SDR=1 RUST_LOG=trace cargo test lifecycle_2kib --no-default-features --features gpu,blst --verbose --package filecoin-proofs --release -- --ignored --nocapture` . The test passed on my ubuntu machine.

[...]

PR #1496, opened 2021-08-19
Serialize parent's cache generation and access

cryptonemo:


storage-proofs-porep/src/stacked/vanilla/cache.rs
```diff
@@ -155,11 +157,26 @@ impl ParentCache {
...
+            if generated.get(&generation_key).is_none() {
+                generated.insert(generation_key);
+            }
             Self::open(len, cache_entries, graph, &path)
         } else {
             match Self::generate(len, cache_entries, graph, &path) {
-                Ok(c) => Ok(c),
+                Ok(c) => {
+                    generated.insert(generation_key);
+                    drop(generated);
```
dignifiedquire, line 176: this is not really needed, as it will be dropped as soon as the function ends

cryptonemo: I thought an explicit drop made sense at the time, but agreed, it's not needed

dignifiedquire (approved):

cryptonemo:
Resolves #1495

dignifiedquire (approved):

PR #1500, opened 2021-08-31
fix: use sync channels in pc2

dignifiedquire:


v-gu:
I noticed the unbounded behavior too, if GPU is too slow, then memory occupation will explode.

cryptonemo (approved):

PR #1501, opened 2021-09-06
Improve multicore SDR

dignifiedquire:
Based on #1477 this updates the parent selection logic in multicore sdr. 

It **does not** implement the bitmask update, as that is not thread safe, and only the thread that fills the buffer is allowed to set this value. This I missed in my earlier review.

### DIFF
--- a/storage-proofs-porep/src/stacked/vanilla/create_label/multi.rs
+++ b/storage-proofs-porep/src/stacked/vanilla/create_label/multi.rs
@@ -57,7 +57,7 @@ const SHA256_INITIAL_DIGEST: [u32; 8] = [
 fn fill_buffer(
     cur_node: u64,
     parents_cache: &CacheReader<u32>,
-    mut cur_parent: &[u32], // parents for this node
+    cur_parent: &[u32], // parents for this node
     layer_labels: &UnsafeSlice<'_, u32>,
     exp_labels: Option<&UnsafeSlice<'_, u32>>, // None for layer0
     buf: &mut [u8],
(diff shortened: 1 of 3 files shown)


qy3u:
It seems that https://github.com/filecoin-project/rust-fil-proofs/blob/9cc6444a78408a1fc19344a973d5ea8d91d63ad4/storage-proofs-porep/src/stacked/vanilla/create_label/multi.rs#L173-L176 makes that update the bitmask before increase consumer is thread safe.
Otherwise, the current `fill`& `get`implemention should not be thread safe too? 🤔

dignifiedquire:
The original code was 
```
*base_parent_missing = 0x20; // Mark base parent 5 as missing
```
instead of 
```
base_parent_missing.set(5);
```
which only sets the bit, without clearing. Which resulted in the bug you discovered.

dignifiedquire:
The last change speeds up replication from 12min to 11min on  my machine.

storage-proofs-porep/src/stacked/vanilla/create_label/multi.rs
```diff
@@ -74,50 +77,46 @@ fn fill_buffer(
...
+    let (predecessor_index, other_drg_parents) = match api_version {
+        ApiVersion::V1_0_0 => (BASE_DEGREE - 1, (0..BASE_DEGREE - 1)),
+        ApiVersion::V1_1_0 => (0, (1..BASE_DEGREE)),
+    };
+
     if cur_node > MIN_BASE_PARENT_NODE {
-        // Mark base parent 5 as missing
-        // base_parent_missing.set_all(0x20);
-        base_parent_missing.set(5);
+        // Mark base parent predrecessor as missing
```
porcuquine, line 86: Should be 'predecessor'.

porcuquine (changes requested):
@dignifiedquire and I went over this, and I believe it now clears the bitmask correctly (and in correct place). This is consistent with the observation that original code (no clearing) was slowest, but first attempted fix was much faster but failed tests. Given that we also have an understanding of the original transcription error that led to the original code, I'm pretty confident in this iteration.

Nevertheless, especially since we've been bitten once, we should make sure this code is tested at least as much as the first attempted fix was. Remember, the discovered test failures were non-deterministic and not expected to always manifest. We need to test this enough times to have reasonable expectation that any failures introduced will have showed up. If we see _any_ such failures, we need to take them seriously (as we did with the previous).

Even though I'm more confident in this than I was in the previous, I'm blocking merging here until I hear that we have successfully tested this (as thoroughly as we _eventually_ did the previous). The reason I am requesting changes rather than approving with 'conditions' is that the latter approach didn't prevent the code from being merged last time. So this is a slight process adjustment based on what we learned in the last round: we need a bit of extra diligence here.

cryptonemo (approved):
Ran the small lifecycle tests looped overnight without issue.  This is where the failure showed up both on CI and on my machine in the past.

dignifiedquire:
Rebased into clear commits, the different changes.

cryptonemo:
Did you want me to re-test here, or are we good to go?

porcuquine (approved):
Based on @cryptonemo's testing — which he believes is sufficient to have caught the previous error but uncovered no problems here — I am approving this now.

cryptonemo (approved):

PR #1504, opened 2021-09-22
fix: upgrade to latest versions

vmx:
The latest versions of the gpu dependencies contain support for CUDA. Therefore
the `gpu` feature is split into `cuda` and `opencl`. By default only `opencl` is
enabled.

BREAKING CHANGE: The `gpu` feature is replaced by the `opencl` feature.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -13,7 +13,7 @@ executors:
     resource_class: 2xlarge+
   gpu:
     machine:
-      image: ubuntu-1604-cuda-10.1:201909-23
+      image: ubuntu-2004-cuda-11.2:202103-01
     working_directory: ~/gpuci
     resource_class: gpu.nvidia.medium
   arm:
@@ -148,6 +148,11 @@ jobs:
   test_gpu_tree_building:
     executor: gpu
     environment: *setup-env
+    parameters:
+      cargo-args:
+        description: Addtional arguments for the cargo command
+        type: string
+        default: ""
     steps:
       - checkout
       - attach_workspace:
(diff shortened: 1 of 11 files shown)


dignifiedquire (changes requested):
can you update the circle ci config to test both features please?

vmx:
I thought I concentrate on OpenCL first as this is what we want to ship. Next step would then be CUDA with proper testing, updates of the README etc. So I would make the CUDA stuff a separate PR. Though perhaps it just works, so let's try it out.

vmx:
I've added a new test run for the `test_gpu_tree_building` as it's the only test that runs on the GPU. It now runs on OpenCL and on CUDA.

dignifiedquire (approved):

vmx:
@cryptonemo I leave the merge to you (I couldn't merge anyway due to "Required statuses must pass before merging").

cryptonemo:
I also don't recall where the required CI tests are checked/unchecked.  I'm ok merging this, but want to be sure that CI will not be broken on master from this point forward.  Any insight?

cryptonemo:
Oh wait, I think I found it: https://github.com/filecoin-project/rust-fil-proofs/settings/branch_protection_rules/2848255

If this seem correct (where I can uncheck ones after merging), I'll proceed.  Sanity check?

dignifiedquire:
the required tests changed with the `blst`/`pairing` change from the earlier PR

cryptonemo:
> the required tests changed with the `blst`/`pairing` change from the earlier PR

I see, I've removed all of the old ones then.

PR #1508, opened 2021-09-28
test(fil-proofs-tooling): warm up cache for window-post verify

dignifiedquire:


vmx (approved):
LGTM, once it's rust fmt'ed correctly.

cryptonemo (approved):
Looks good, although if we're doing this I'd measure and output the cold verify as well as the warm to get a better picture.

dignifiedquire:
> Looks good, although if we're doing this I'd measure and output the cold verify as well as the warm to get a better picture.

the difference is `1.9s` vs `6ms` on my machine :D

cryptonemo:
> > Looks good, although if we're doing this I'd measure and output the cold verify as well as the warm to get a better picture.
> 
> the difference is `1.9s` vs `6ms` on my machine :D

Yes.  This is information that should be reported in the bench.

PR #1510, opened 2021-09-30
expose 'multicore-sdr' feature

qy3u:
You can't use the multicore-sdr feature through filecoin-proofs in current implemention.

### DIFF
--- a/filecoin-proofs/Cargo.toml
+++ b/filecoin-proofs/Cargo.toml
@@ -72,6 +72,7 @@ opencl = [
     "filecoin-hashers/opencl",
     "fr32/opencl",
 ]
+multicore-sdr = ["storage-proofs-porep/multicore-sdr"]
 
 [[bench]]
 name = "preprocessing"


qy3u:
I noticed that `filecoin-proofs-api` has a `multicore-sdr` feature:
https://github.com/filecoin-project/rust-filecoin-proofs-api/blob/b8559d4141639f91c503cf06dac77bf77bd322e5/Cargo.toml?_pjax=%23js-repo-pjax-container%2C%20div%5Bitemtype%3D%22http%3A%2F%2Fschema.org%2FSoftwareSourceCode%22%5D%20main%2C%20%5Bdata-pjax-container%5D#L27

It exposed this feature through ``storage-proofs-porep``, **but all the functions are called through** ``filecoin_proofs_v1``.So, I really doubt that the multicore feature in ``filecoin-proofs-api`` will behave as it expected. 

Considering that `storage-proofs-porep` is not used directly in `filecoin-proofs-api`, I think that as a dependency, `storage-proofs-porep` can be removed in that crate.

cryptonemo (approved):

PR #1526, opened 2021-10-13
feat: allow for window post proving on a partition basis

dignifiedquire:
This is based on #1497 and attempts to reduce the changes to a minimum needed.


cryptonemo:
> @cryptonemo any thoughts why only darwin would show the testfailure?

CI flakiness is my guess.  The error is that it's missing parameters.  I suspect a re-run may resolve it(?)

PR #1527, opened 2021-10-15
fix: benchy: print help if no command is given

vmx:
Prior to this change benchy will panic if there were no options given.
Now it prints the help.

I wanted to do this for years, finally I did.


vmx:
Oh dear, why didn't I update to master?

dignifiedquire (approved):

PR #1528, opened 2021-10-15
Empty sector update tree r gpu

cryptonemo:


cryptonemo:
This may be slower in the general case, so it needs to be perf tested before it can be considered.  Working on that ...

cryptonemo:
> > Odd, the failing cuda tests all pass locally on my setup :-(
> 
> Nevermind, I can reproduce now

I'm confused.  Reproduced once, but can't again.  Something is going on.

cryptonemo:
Final testing showed comparablly equal performance for CPU tree building and improved building for GPU.

PR #1529, opened 2021-10-15
feat: update bellperson to v0.18.0

cryptonemo:
This PR updates bellperson, neptune (for using the latest bellperson as well) and the rust-toolchain version.


rust-toolchain
```diff
@@ -1 +1 @@
-1.51.0
+1.54.0
```
dignifiedquire, line 1: can you put this in its own commit? just so it is easier to track historically

PR #1535, opened 2021-11-04
reset consumer

firesWu:
Reset consumer every time when finish one layer.

If consumer not reset, make cause parent cache loading problem.

1. `cur_parent_ptr`  may point to a released memory location, causing panic.
2. `cur_parent_ptr` may point to a wrong parent cache window, resulting wrong pc1 result.


qy3u:
The bug you mentioned doesn't seem to be due to the consumer.
The value of the consumer has been correctly set to 1 [here](https://github.com/filecoin-project/rust-fil-proofs/blob/ec2536447c1bfef422b100afa80988e9bb136da8/storage-proofs-porep/src/stacked/vanilla/create_label/multi.rs?_pjax=%23js-repo-pjax-container%2C%20div%5Bitemtype%3D%22http%3A%2F%2Fschema.org%2FSoftwareSourceCode%22%5D%20main%2C%20%5Bdata-pjax-container%5D#L315-L316)

firesWu:
> The bug you mentioned doesn't seem to be due to the consumer value. The value of consumer has been correctly set to 1 [here](https://github.com/filecoin-project/rust-fil-proofs/blob/ec2536447c1bfef422b100afa80988e9bb136da8/storage-proofs-porep/src/stacked/vanilla/create_label/multi.rs?_pjax=%23js-repo-pjax-container%2C%20div%5Bitemtype%3D%22http%3A%2F%2Fschema.org%2FSoftwareSourceCode%22%5D%20main%2C%20%5Bdata-pjax-container%5D#L315-L316)

The thread that calls `create_label_runner` was created before consumer being reset to 1.  so we cannot be sure that the consumer used [here](https://github.com/filecoin-project/rust-fil-proofs/blob/ec2536447c1bfef422b100afa80988e9bb136da8/storage-proofs-porep/src/stacked/vanilla/create_label/multi.rs?_pjax=%23js-repo-pjax-container%2C%20div%5Bitemtype%3D%22http%3A%2F%2Fschema.org%2FSoftwareSourceCode%22%5D%20main%2C%20%5Bdata-pjax-container%5D#L88) and [here](https://github.com/filecoin-project/rust-fil-proofs/blob/ec2536447c1bfef422b100afa80988e9bb136da8/storage-proofs-porep/src/stacked/vanilla/memory_handling.rs?_pjax=%23js-repo-pjax-container%2C%20div%5Bitemtype%3D%22http%3A%2F%2Fschema.org%2FSoftwareSourceCode%22%5D%20main%2C%20%5Bdata-pjax-container%5D#L248) is already reset to 1.   @qy3u

qy3u:
I think you are right.
The consumer needs to be reset before `create_label_runner`.

vmx:
I've only had a quick look, but to me it sounds like it is the right thing to do, to also reset the consumer.

vmx (dismissed):
@firesWu Sorry for the long delay. I know had a closer look. You change looks absolutely right. Thanks for spotting this issue. I was recently also deep into that part of the code, but I haven't spotted this issue. Good job!

Could you please rebase your change to current master (so that we don't end up with strange merge commits)?

firesWu:
@vmx I have rebase my change to current master

vmx (approved):

dignifiedquire (approved):

PR #1549, opened 2021-12-23
Add flag for winpost

phoenixzhua:
This PR gives a solution to error "mined block in the past" due to windowpost and winpost happening at the same time. 

Add a flag to identify whether it is winning post or not.

related discussion: https://github.com/filecoin-project/lotus/discussions/7797

There is another related PR for this solution:  https://github.com/filecoin-project/bellperson/pull/246


storojs72:
Seems that this PR is actually a staled work. Should we may be close it for now?

PR #1557, opened 2022-01-25
fix: correctly set cache path for empty sector update proofs

arajasek:
**This PR is mostly intended to indicate where I think a bug exists**

I'm not entirely sure whether this is the right fix, but something around the cache path here seems to be wrong, causing most users on SnapNet to be unable to upgrade their sectors (see https://github.com/filecoin-project/lotus/issues/7968).

We may only want to be setting the path in some of the use cases for this method.


cryptonemo:
@arajasek Good catch! It looks like the issue is that we get the t_aux from one place, and persist it to another (after encode/remove/etc).  When re-instantiating after it's persisted to the new location, the cache path may be incorrect, so explicitly setting it at that time is preferred.

cryptonemo (approved):

PR #1576, opened 2022-03-11
build: use rust-toolchain 1.56.0

DrPeterVanNostrand:
This allows us to have dependencies that use Rust 2021 edition.


cryptonemo:
Place your bets folks -- the darwin CI job will timeout trying to download the SRS file ... again :-(

PR #1577, opened 2022-03-17
fix data race when create label

OneMore14:
After each layer calls create_layer_labels()，the parents_cache's cursor has drifted. At the beginning of creating the next layer label, ring_buf may read wrong data. And the error really happened widely. Although the labels are wrong, the sector can still pass C2 because [PoRep Challenges](https://spec.filecoin.io/#section-algorithms.sdr.porep-challenges) only challenge a small part of nodes. 

This error also causes a sector can't be unsealed correctly.

(This is my first pull request for open source project, if I did something wrong, please let me know, thank you~)

### DIFF
--- a/storage-proofs-porep/src/stacked/vanilla/create_label/multi.rs
+++ b/storage-proofs-porep/src/stacked/vanilla/create_label/multi.rs
@@ -487,6 +487,8 @@ pub fn create_labels_for_encoding<Tree: 'static + MerkleTreeTrait, T: AsRef<[u8]
             parents_cache.finish_reset()?;
         }
 
+        parents_cache.store_consumer(0);
+
         create_layer_labels(
             &parents_cache,
             replica_id.as_ref(),


OneMore14:
More explanation: each layer calls ```parents_cache.store_consumer(1);```. But before that statement, the codes spawn some threads to run ```create_label_runner()```. So this error may only occurs when multi core replication and layer 2-11(the cache cursor is default 0 when creating the first layer)

vmx:
@OneMore14 Thanks for the PR. I wonder how this change fixes issues. To me it seems that at this location on the code, the consumer is always already set to 0. Let me explain.

You are saying that it only happens at layers 2-11, so let's look only at this case. We get into the code path at https://github.com/filecoin-project/rust-fil-proofs/blob/27f15df5ae4345cc91b6175912dc955c8bade8e9/storage-proofs-porep/src/stacked/vanilla/create_label/multi.rs#L486-L488. It calls `parents_cache.finish_reset()`. When looking at the source of `finish_reser()`: https://github.com/filecoin-project/rust-fil-proofs/blob/27f15df5ae4345cc91b6175912dc955c8bade8e9/storage-proofs-porep/src/stacked/vanilla/memory_handling.rs#L181-L188 There at line 186 we already call `self.store_consumer(0);`.

What am I missing?

OneMore14:
@vmx   emmm... I'm sorry I did't find [this pr](https://github.com/filecoin-project/rust-fil-proofs/pull/1535) before. Actually we talk about the same issue. I found this error last year but I did't check the updated codes carefully. you can close this pr, sorry for waste your time

vmx:
@OneMore14 No worries. It's good to see that we fixed an issue that also others were seeing.

PR #1578, opened 2022-03-24
chore: update deps

Stebalien:
Hopefully this will let us use fewer duplicate deps.

Explicitly doesn't update:

- bellperson
- neptune
- bitvec 1.0 because I couldn't get it working and don't really care (only used in testing).

Importantly:

- Upgrades to clap 3
- Upgrades to bitvec 1
- Upgrades to blake2b/blake2s 1.0
- Upgrades byte-unit (requires rust 2021 edition support).
- Replaces cpuid-bool with cpufeatures (as recommended).
- Upgrades aes and fixes some fallout.
- Upgrades hwloc (in case that matters)
- Upgrades num-bigint (in case that matters)

### DIFF
--- a/fil-proofs-tooling/Cargo.toml
+++ b/fil-proofs-tooling/Cargo.toml
@@ -16,7 +16,7 @@ storage-proofs-post = { path = "../storage-proofs-post", version = "^11.0.0", de
 storage-proofs-update = { path = "../storage-proofs-update", version = "^11.0.0", default-features = false }
 filecoin-proofs = { path = "../filecoin-proofs", default-features = false }
 filecoin-hashers = { path = "../filecoin-hashers", default-features = false, features = ["poseidon", "blake2s", "sha256"] }
-clap = "2"
+clap = { version = "3.1.6", features = ["derive"] }
 serde = { version = "1.0", features = ["derive"] }
 serde_json = "1.0"
 toml = "0.5"
@@ -30,13 +30,13 @@ bellperson = "0.18.0"
 rand = "0.8"
 tempfile = "3.0.8"
 cpu-time = "1.0.0"
-git2 = "0.13.6"
+git2 = "0.14.2"
 heim = { git = "https://github.com/heim-rs/heim", rev = "b292f15", features = ["host", "memory", "cpu"] }
 async-std = "1.6"
-blake2s_simd = "0.5.6"
+blake2s_simd = "1.0.0"
 fil_logger = "0.1"
 log = "0.4.8"
-uom = "0.30"
+uom = "0.32.0"
 merkletree = "0.21.0"
 bincode = "1.1.2"
 anyhow = "1.0.23"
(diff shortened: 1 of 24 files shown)


fil-proofs-tooling/src/bin/gpu-cpu-test/main.rs
```diff
@@ -34,11 +34,24 @@ const POST_CONFIG: PoStConfig = PoStConfig {
...
-    pub enum Mode {
-        Threads,
-        Processes,
+#[derive(Debug, Clone)]
+pub enum Mode {
+    Threads,
+    Processes,
+}
+
+impl FromStr for Mode {
```
Stebalien, line 43: Technically, one can derive `ArgEnum`. But that doesn't actually work with `type_of_t`, so I'm not sure what one is supposed to do here. But this version _should_ actually work.

fr32/src/padding.rs
```diff
@@ -793,23 +793,23 @@ mod tests {
             let num_bits = rng.gen_range(1..data.len() * 8 - pos);
             let new_offset = rng.gen_range(0..8);
 
-            let mut bv = BitVec::<LittleEndian, u8>::new();
+            let mut bv = BitVec::<u8, Lsb0>::new();
```
Stebalien, line 796: I'm pretty sure my bitfield changes are correct, but double check.

storage-proofs-core/src/crypto/aes.rs
```diff
@@ -9,17 +10,19 @@ const IV: [u8; 16] = [0u8; 16];
...
     ensure!(key.len() == 32, "invalid key length");
 
-    let mode = Cbc::<Aes256, ZeroPadding>::new_var(key, &IV).context("invalid key")?;
+    let mode = cbc::Decryptor::<aes::Aes256>::new_from_slices(key, &IV).context("invalid key")?;
 
-    let res = mode.decrypt_vec(ciphertext).context("failed to decrypt")?;
+    let res = mode
+        .decrypt_padded_vec_mut::<ZeroPadding>(ciphertext)
+        .context("failed to decrypt")?;
     Ok(res)
```
Stebalien, line 26: Check this carefully.

Stebalien:
This turned into a bit of a "deal with all the deprecated things"-fest.

fr32/src/padding.rs
Stebalien: EH, I dropped this.

Stebalien:
> Hopefully this will let us use fewer duplicate deps.

Ha, ha ha. Who was I kidding.

Stebalien:
The test failures are timeouts. I'm not sure what the right approach there is.

vmx:
> The test failures are timeouts. I'm not sure what the right approach there is.

I'm working on that. :( that our current cache is broken. Sounds like I need to provide a quick fix to unblock folks.

vmx:
@Stebalien https://github.com/filecoin-project/rust-fil-proofs/pull/1582 that PR should make CI work again. Sadly it takes a long time to run (I'm also working on that problem).

fil-proofs-tooling/Cargo.toml
```diff
@@ -30,30 +30,31 @@ bellperson = "0.18.0"
...
 anyhow = "1.0.23"
 ff = "0.11.0"
 rand_xorshift = "0.3.0"
 bytefmt = "0.1.7"
 rayon = "1.3.0"
-flexi_logger = "0.16.1"
+flexi_logger = "0.22.3"
 typenum = "1.11.2"
 generic-array = "0.14.4"
 # `byte-unit = 4.0.13` is the final version that uses Rust 2018 edition.
```
vmx, line 50: This comment should be removed.

vmx (commented):
LGTM, but I'd like that @DrPeterVanNostrand has another at the hasher changs.

cryptonemo:
> LGTM, but I'd like that @DrPeterVanNostrand has another at the hasher changs.

Ditto, thanks!

cryptonemo:
Closing in favor of #1584

PR #1580, opened 2022-03-25
perf(filecoin-proofs): increase parallelism in window-post

someonegg:
This improves the parallelism when generating merkle tree.

### DIFF
--- a/filecoin-proofs/src/api/window_post.rs
+++ b/filecoin-proofs/src/api/window_post.rs
@@ -3,6 +3,7 @@ use std::collections::BTreeMap;
 use anyhow::{ensure, Context, Result};
 use filecoin_hashers::Hasher;
 use log::info;
+use rayon::prelude::{IntoParallelRefIterator, ParallelIterator};
 use storage_proofs_core::{
     compound_proof::{self, CompoundProof},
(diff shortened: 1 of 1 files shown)


cryptonemo:
Thanks for submitting, this, but this has been tried in the past and didn't appear to make a measurable difference: https://github.com/filecoin-project/rust-fil-proofs/pull/1286/files#diff-5bf9d3ebeba9a4d7a43ef35fae48b5c36e9e7fe50b4eec99515d8a1fc1f15010R360

The problem is that we're instantiating trees at this point, not building them in parallel.  If you can show positive measurements around this (and describe the I/O system you're on), it would help.

someonegg:
NFS protocol

When the page cache of nfs-server becomes the bottleneck, time between "generate_window_post: start" and "vanilla_proofs:start" will increase from 1 minute to more than 6 minutes.

According to the io log, this stage needs to read the last 32 bytes of each last-tree file.

cryptonemo:
> 

Can you comment on the improved time with this change?

someonegg:
With this change, the time is reduced to about 1 minute.

cryptonemo (approved):

Issue #1534, opened 2021-10-25
Darwin CI Generated Parameters


cryptonemo:
### Description

Currently, we generate test parameters on darwin each time CI tests are run.  These runs can now take over 80 minutes, but often are quicker.

It would be better if we can save/restore them similar to how we do on Linux (but simply re-using the existing code doesn't work on darwin).

PR #1583, opened 2022-03-25
chore: CI: use parameter cache also for macOS

vmx:
In CircleCI you cannot restore a cache on macOS that was saved on Linux.
The workaround is to persist those files in a workspace. That workspace
can then be attached on macOS. So the workflow for macOS is:

  1. (Linux) Generate parameters and cache them
  2. (Linux) Restore cache and preserve in workspace
  3. (macOS) Attach workspace

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -27,6 +27,21 @@ setup-env: &setup-env
 
 
 jobs:
+  # This job is needed for macOS, which doesn't support restoring caches that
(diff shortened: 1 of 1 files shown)


Stebalien (approved):

cryptonemo:
Resolves https://github.com/filecoin-project/rust-fil-proofs/issues/1534

cryptonemo (approved):

PR #1585, opened 2022-04-12
chore: CI: don't always run test_seal_lifecycle_upgrade

vmx:
It probably was a debug left-over from the sector upgrade work. The
`test_seal_lifecycle_upgrade` test is included when running with
the default test arguments, so there's no need to mention it separately.
It was also run in the non-ignore case. There's no point of running
it twice.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -201,7 +201,6 @@ jobs:
             ulimit -u 20000
             ulimit -n 20000
             cargo +<< pipeline.parameters.nightly-toolchain >> test --all --verbose --release << parameters.cargo-args >> -- --nocapture << parameters.test-args >>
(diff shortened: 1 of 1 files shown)


vmx:
> Have you verified in CI logs that it was run twice?

Yes. Search for `e.g. `test_seal_lifecycle_upgrade_16kib_sub_8_2_v1_1` in https://app.circleci.com/pipelines/github/filecoin-project/rust-fil-proofs/3825/workflows/bf530078-e258-4a32-9b7b-db0c5ecc04f7/jobs/86473/parallel-runs/0/steps/0-108 and https://app.circleci.com/pipelines/github/filecoin-project/rust-fil-proofs/3825/workflows/bf530078-e258-4a32-9b7b-db0c5ecc04f7/jobs/86475/parallel-runs/0/steps/0-108.

cryptonemo (approved):

PR #1594, opened 2022-04-26
chore: CI: split gpu tree building test runs

vmx:
Split the tree building CI runs into the ones from `filecoin-proofs`
`+ storage-proofs-update` and the rest for the ignored tests. This
should get the total runtime of those tests down to about 30mins.


cryptonemo (dismissed):

vmx:
Everyone, please don't merge this one, it might not be needed/different due to the other CI related PRs, I'd like to merge those first and then revisit that one again.

PR #1596, opened 2022-04-29
Use non-zk proofs for storage proofs

Kubuxu:


ribasushi:
@Kubuxu @porcuquine @cryptonemo is there a reason this one stalled? Asking as I was correcting someone earlier today, and stated that:

> Actually filecoin does not rely on ZK anywhere in its circuits. It is all "vanilla" SNARKs (some of them are computed via inefficient-for-the-case ZK-capable code)

Is what I said accurate, or do I need to walk that back? 😅

porcuquine:
You are fine regardless because Filecoin security doesn’t rely on zero knowledge.

cryptonemo:
Outdated and unlikely to be used.  May revisit at some point in the future though.

PR #1600, opened 2022-05-09
chore: remove unused dependencies

### LINKED ISSUE #1598: Check dependencies of all crates
### Description

Over the years there were several refactors, which the lead to dependencies in Cargo.toml files which aren't actually used. I propose looking into all crates and check which dependencies are still needed. Some might also be only needed as dev-dependency (similarly to what I did at https://github.com/filecoin-project/rust-fil-proofs/pull/1593).

### Acceptance criteria

All crates specify only the dependencies in their Cargo.toml they actually need.

### Risks + pitfalls

I don't see risks, I think our test coverage on CI is good enough.

### Where to begin

There is tooling to check for unused dependencies. From my experience they don't always work well. This is what I did for the PR mentioned above:

 - Comment out all dependencies and run `cargo build`. Comment the normal dependencies in again until the build works again.
 - Run `cargo test` and start to comment in the dev-dependencies until the tests compile.
 - Check CI that everything still works.

0x5459:
Remove unused dependencies. 
resolves #1598 .

### DIFF
--- a/fr32/Cargo.toml
+++ b/fr32/Cargo.toml
@@ -9,7 +9,6 @@ repository = "https://github.com/filecoin-project/rust-fil-proofs"
 
 [dependencies]
 anyhow = "1.0.23"
-bellperson = "0.18.0"
 byte-slice-cast = "1.0.0"
 byteorder = "1"
 ff = "0.11.0"
@@ -26,8 +25,8 @@ rand_xorshift = "0.3"
 
 [features]
 default = []
-cuda = ["bellperson/cuda"]
-opencl = ["bellperson/opencl"]
+cuda = []
+opencl = []
 
 [[bench]]
 name = "fr"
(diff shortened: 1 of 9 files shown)


dignifiedquire (approved):

fr32/Cargo.toml
```diff
@@ -26,8 +25,8 @@ rand_xorshift = "0.3"
 
 [features]
 default = []
-cuda = ["bellperson/cuda"]
-opencl = ["bellperson/opencl"]
+cuda = []
+opencl = []
```
vmx, line 29: You can remove the whole features section.

vmx (commented):
Just a minor thing, to clean things up even more.

fr32/Cargo.toml
0x5459: ok.

vmx (approved):
Thanks!

PR #1613, opened 2022-06-13
chore: lock versions to the correct minor, and update

Stebalien:
1. Update minimum versions for in-workspace crates.
2. Use the tilde constraint to ensure that we don't update to the next minor version.

Otherwise, the next minor release (which will re-release the changes on master) will break anything depending on this release.

NOTE: we should also go ahead and _release_ those minor releases. This is kind of hard to test otherwise.

### DIFF
--- a/fil-proofs-param/Cargo.toml
+++ b/fil-proofs-param/Cargo.toml
@@ -9,12 +9,12 @@ repository = "https://github.com/filecoin-project/rust-fil-proofs"
 readme = "README.md"
 
 [dependencies]
-storage-proofs-core = { path = "../storage-proofs-core", version = "^11.0.0", default-features = false}
-storage-proofs-porep = { path = "../storage-proofs-porep", version = "^11.0.0", default-features = false }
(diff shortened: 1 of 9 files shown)


arajasek:
is CI sad because it was opened from a fork?

vmx:
> is CI sad because it was opened from a fork?

No that's not the issue. I think I know how to fix that.

vmx:
@arajasek I was wrong, it was the issue. Though, this PR should be green soon.

arajasek:
@vmx Still sad :(

Stebalien:
Looks like something timed out.

Stebalien:
It's failing to fetch parameters. I'm seeing if I can get that working.

vmx:
> It's failing to fetch parameters. I'm seeing if I can get that working.

`ipget` doesnt' really work properly (reported here: https://github.com/ipfs/ipget/issues/103). The fix is to use curl instead. I'll backport https://github.com/filecoin-project/rust-fil-proofs/commit/3676e5b5a87fffee2741fc723cd23d6cbf079527.

PR #1614, opened 2022-06-14
chore: lock versions to the correct minor, and update

vmx:
1. Update minimum versions for in-workspace crates.
2. Use the tilde constraint to ensure that we don't update to the next
   minor version. The next minor version is going to revert the change.


vmx:
Not really needed anymore, lets use the original PR.

PR #1615, opened 2022-06-14
Updated deps

cryptonemo:


Cargo.toml
```diff
@@ -12,3 +12,7 @@ members = [
   "filecoin-hashers",
   "storage-proofs-update",
 ]
+
+[patch.crates-io]
+bellperson = { git = "https://github.com/filecoin-project/bellperson", branch = "forward-port-updates" }
+neptune = { git = "https://github.com/filecoin-project/neptune", branch = "updated-deps" }
```
cryptonemo, line 18: FIXME: Requires https://github.com/filecoin-project/bellperson/pull/274 and https://github.com/filecoin-project/neptune/pull/150 followed by crate releases

vmx (commented):
Many dependencies switched from `^` (which is the implicit default)` to `~`. I'd prefer if we would keep `^` (resp. the implicit default). This is what semver is about. This way one wouldn't need to update the Cargo.toml as often.

Though I also don't have a super strong opinion about it and could live with manually updating things more often.

DrPeterVanNostrand (approved):

PR #1618, opened 2022-07-01
chore: downgrade neptune to 5.1.x

vmx:
Downgrading `neptune` makes the master branch build again.

### DIFF
--- a/filecoin-hashers/Cargo.toml
+++ b/filecoin-hashers/Cargo.toml
@@ -18,7 +18,7 @@ anyhow = "1.0.34"
 serde = "1.0.117"
 rand = "0.8.0"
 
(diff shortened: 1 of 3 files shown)


cryptonemo:
I'm preferring #1615 to this and may merge.  We were waiting on @porcuquine to release Neptune v7, but could merge while it's still a neptune branch if causing issues.

storojs72:
Downgrading neptune is good workaround, but it doesn't free us from updating bellperson here in rust-fil-proofs. And unfortunately we will get this issue again and again when versions of bellperson become different. Should we may be prefer upgrading dependencies rather than downgrading?

cryptonemo:
Closing since https://github.com/filecoin-project/rust-fil-proofs/pull/1615 was merged

PR #1623, opened 2022-08-05
feat: update rust-toolchain to 1.62.0

cryptonemo:


vmx (changes requested):
What are the reasons for the upgrade?

I think it should be 1.62.1.

cryptonemo:
> What are the reasons for the upgrade?
> 
> I think it should be 1.62.1.

No particular reason -- we just did a release, so starting the next cycle with the latest toolchain seems reasonable since we'll be testing everything from here.

Please push the update to 1.62.1 if you'd like.

vmx:
> No particular reason -- we just did a release, so starting the next cycle with the latest toolchain seems reasonable since we'll be testing everything from here.

I'd say it's generally good practice in the Rust world to stay on the lowest possible version. But I can see how this produces less work for us.

cryptonemo:
Looks good, but I can't approve my own PR :-)

vmx (approved):

PR #1624, opened 2022-08-12
chore: use memmap2 instead of mapr

vmx:
`memmap2` is maintained and contains all features from `mapr`, so using
the `mapr` fork of `memmap` is no longer needed.

Here are the changes in detail:

 - `map_anon()` in `memmap2` is always `MAP_PRIVATE`, so a call to `private()`
is no longer needed.
 - using `mlock` is similar to the `MAP_LOCKED` setting, hence it is no longer
   set. In `memmap2` the `mlock()` is just called `lock()`.
 - `MAP_PRIVATE` is done via `map_copy_read_only()`.


storojs72 (approved):
Thanks. I believe it is good to merge this PR after this one - https://github.com/filecoin-project/rust-fil-proofs/pull/1625 that fixes Darwin CI job execution. Plus it looks like we can now archive https://github.com/filecoin-project/mapr repository as it is not used anymore. Who can do that (I cannot unfortunately...)? May be @dignifiedquire ?

dignifiedquire (approved):

cryptonemo (approved):

PR #1625, opened 2022-08-12
chore: update CircleCI to xcode 13.4.1

vmx:
Version 12.5.0 doesn't seem to be supported anymore, hence upgrade to the
latest non-beta release.

Simulators for iOS (which we don't need) are deleted, so that there's enough
space for the Filecoin params.


vmx:
It's currently running out of disk space, for more info see https://circleci.com/docs/using-macos#macos-vm-storage.

tawmaz (approved):

storojs72 (approved):

shawnrader (approved):

PR #1628, opened 2022-08-29
Tarpaulin coverage

storojs72:
This PR adds new CI job that actually evaluates test coverage of the Proofs codebase. 

[Tarpaulin](https://github.com/xd009642/tarpaulin) tool is used for this purpose.

Current total coverage, which is 37%, obtained by running regular tests (no ignored/isolated/big tests). Invoking other tests causes too long job execution.

I consider following next steps in this direction:
- add more regular simple tests to improve coverage, as it is rather low currently;
- enforce current value by CI, to prevent merging PRs that reduce coverage;
- add coverage results exporting to codecov.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -326,8 +326,8 @@ jobs:
             -v "${PWD}:/volume"
             -v "${FIL_PROOFS_PARAMETER_CACHE}:/var/tmp/filecoin-proof-parameters/"
             xd009642/tarpaulin
-            sh -c "apt update && apt install -y libssl-dev ocl-icd-opencl-dev libhwloc-dev && cargo tarpaulin --timeout 600 --release -v --ignored"
-
+            sh -c "apt update && apt install -y libssl-dev ocl-icd-opencl-dev libhwloc-dev && cargo tarpaulin --timeout 1800 --release -v --ignored"
+          no_output_timeout: 30m
 commands:
   ensure_filecoin_parameters:
     steps:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -326,7 +326,7 @@ jobs:
             -v "${PWD}:/volume"
             -v "${FIL_PROOFS_PARAMETER_CACHE}:/var/tmp/filecoin-proof-parameters/"
             xd009642/tarpaulin
-            sh -c "apt update && apt install -y libssl-dev ocl-icd-opencl-dev libhwloc-dev && cargo tarpaulin --timeout 1800 --release -v --ignored"
+            sh -c "apt update && apt install -y libssl-dev ocl-icd-opencl-dev libhwloc-dev && cargo tarpaulin --timeout 1800 --release -v"
           no_output_timeout: 30m
 commands:
   ensure_filecoin_parameters:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -307,6 +307,27 @@ jobs:
             cargo test --release --verbose --workspace -- --nocapture
           no_output_timeout: 2h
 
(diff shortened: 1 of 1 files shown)


vmx (approved):
Looks good as a first version.

I was wondering about the low coverage. I think tests should be excluded from coverage. Also I think excluding the `fil-proofs-tooling` crate would make sense, I wouldn't care much about test coverage for our tooling.

It would be interesting to see how much coverage we have if also the ignored tests are run, I've no clue if they test additional things. This can be a single local run, just to get an idea.

Another interesting thing would be to run it with other tooling like https://github.com/taiki-e/cargo-llvm-cov to see if the results roughly match.

vmx:
I had a look at the results. We need to make sure that the coverage is run with the configurations we use on CI and then merge the results. For example 

    storage-proofs-porep/src/stacked/vanilla/create_label/multi.rs: 71, 73, 82, 86, 88, 90, 92-94, 96-97, 101, 105, 107-108, 113, 118-120, 175, 195, 229, 249, 341, 439, 446, 448, 450-452, 454, 456, 459, 462-463, 468-471, 474-475, 477-478, 481-482, 487-488, 492-496, 498, 500-502, 507-508, 511, 513, 515-516, 518, 520, 525-528, 530, 545-546, 548-549, 552, 557, 566-569, 582-584, 586, 588, 590-592, 603, 607, 609-612, 616, 619, 628
|| storage

Looks really bad. If you e.g. [look at line 71](https://github.com/filecoin-project/rust-fil-proofs/blob/39e4302882797a3efe4f078873b9211eca8c5499/storage-proofs-porep/src/stacked/vanilla/create_label/multi.rs#L71) which isn't covered according to the test run. I'm sure that is covered by running e.g. the lifecycle test with multicore SDR enabled, which we [run on CI](https://github.com/filecoin-project/rust-fil-proofs/blob/39e4302882797a3efe4f078873b9211eca8c5499/.circleci/config.yml#L125).

vmx:
In regards to code coverage, https://github.com/filecoin-project/builtin-actors/pull/120 might also be interesting.

shawnrader:
Do you know what the cause is of the "instrumentation address clash" that shows up in the tarpaulin test results is?

storojs72:
@shawnrader, according to [this](https://github.com/xd009642/tarpaulin/issues/190#issuecomment-451421618) explanation from tarpaulin creator, such clashes occur when multiple lines of code are mapped to the same memory address. They are not considered as error, though may have impact on accuracy of coverage report. I believe that this is especially the case for multi-threaded code, that we have a lot in `rust-fil-proofs`.

storojs72:
> I think tests should be excluded from coverage. Also I think excluding the fil-proofs-tooling crate would make sense, I wouldn't care much about test coverage for our tooling.

@vmx, it makes sense, indeed. Of course, current result is just `referential` value of coverage. Let's think of it as a lower bound. It will grow with eventual excluding non-relevant code from coverage and adding more fine-grained unit tests.

Regarding including ignored tests - I would be great indeed, but the problem is that according to my [test](https://app.circleci.com/pipelines/github/filecoin-project/rust-fil-proofs/4241/workflows/06aa0dc3-9b32-41d6-93d5-e5a117157168/jobs/93365) run (yet unsuccessful) job execution takes more than one hour which makes it most long-running  which is bad from dev/exp point of view, so I decided to use only regular tests for coverage estimation. I would additionally experiment with other coverage tools ([this](https://github.com/taiki-e/cargo-llvm-cov)one that you mentioned, and [this](https://github.com/mozilla/grcov) one alternatively which is also LLVM-based), looking at accuracy of results that they can provide and time of job's execution

cryptonemo (approved):

Issue #1629, opened 2022-09-14
Window Posts no longer work for sectors on a read-only filesystem


clinta:
Something changed with v12 which seems to require writing to the sector store for window posts.

See: https://github.com/filecoin-project/lotus/issues/9298

Bisecting lotus narrowed down the issue to the upgrade in the rust-fil-proofs from 11.1 to 12.0.

This can be tested using `lotus-bench`, with the sealed secctor and cache located on a read-only filesystem. In previous versions this works without issue. After https://github.com/filecoin-project/lotus/commit/fa04496c81a7348e4ca75b3c0b9299cab4e898ec it fails with the following:

```
2022-09-14T03:41:29.046 INFO filcrypto::util::types > generate_fallback_sector_challenges: start
2022-09-14T03:41:29.046 INFO filecoin_proofs::api::post_util > generate_sector_challenges:start
2022-09-14T03:41:29.046 INFO filecoin_proofs::api::post_util > generate_sector_challenges:finish
2022-09-14T03:41:29.046 INFO filcrypto::util::types > generate_fallback_sector_challenges: end
2022-09-14T03:41:29.046 INFO filcrypto::util::types > generate_single_vanilla_proof: start
2022-09-14T03:41:29.046 INFO filecoin_proofs::api::post_util > generate_single_vanilla_proof:start: SectorId(24456)
2022-09-14T03:41:29.047 TRACE filecoin_proofs::types::private_replica_info > post: base tree size 153391689, base tree leafs 134217728, rows_to_discard 2, arities [8, 8, 0]
2022-09-14T03:41:29.047 INFO filcrypto::util::types > generate_single_vanilla_proof: end
2022-09-14T03:41:29.047Z	WARN	lotus-bench	lotus-bench/main.go:114	generate_single_vanilla_proof: merkle_tree failed: SectorId(24456)

Caused by:
    Read-only file system (os error 30)
```

solopine:
WinningPost can also fail if WinningPost machine uses readonly storage.

cryptonemo:
This part of the proofs code did not change (particularly around this point, though we never design or test with read-only storage in any cases).  Check `dmesg` on your system and see if you can better locate the cause. A common one may be that the system is trying to read past the end of storage (depends on how/why the storage became read-only).  I suspect a reboot and disk check may help.

cryptonemo:
[...]

PR #1630, opened 2022-09-15
test: make sure that WindowPoSt also works on read-only files

vmx:
This commit adapts the test to make the generated files read-only before a WindowPoSt is performed. This simulates the use case when you mount your sectors read-only after generation.

### DIFF
--- a/filecoin-proofs/tests/api.rs
+++ b/filecoin-proofs/tests/api.rs
@@ -865,6 +865,10 @@ fn winning_post<Tree: 'static + MerkleTreeTrait>(
     let challenges =
         generate_fallback_sector_challenges::<Tree>(&config, &randomness, &[sector_id], prover_id)?;
 
+    // Make sure that files can be read-only for a window post.
+    set_readonly_flag(replica.path(), true);
+    set_readonly_flag(cache_dir.path(), true);
+
     let single_proof = generate_single_vanilla_proof::<Tree>(
         &config,
         sector_id,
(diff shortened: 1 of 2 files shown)


cryptonemo (approved):

filecoin-proofs/tests/api.rs
```diff
@@ -1311,6 +1323,11 @@ fn window_post<Tree: 'static + MerkleTreeTrait>(
 
     let mut vanilla_proofs = Vec::with_capacity(replica_sectors.len());
 
+    // Make sure that files can be read-only for a window post.
+    for (_, _, _, cache_dir, _) in &sectors {
+        set_readonly_flag(cache_dir.path(), true);
```
storojs72, line 1328: I would also apply read-only setting to the replica file besides cache directory

storojs72 (commented):
We also need to test winning_post similarly - as it was mentioned in the original issue: https://github.com/filecoin-project/rust-fil-proofs/issues/1629#issuecomment-1246415319

cryptonemo:
Kicked off the CI again now that `merkletree` has been released

cryptonemo:
> We also need to test winning_post similarly - as it was mentioned in the original issue: [#1629 (comment)](https://github.com/filecoin-project/rust-fil-proofs/issues/1629#issuecomment-1246415319)

Not a bad idea, but not a requirement for this PR

vmx:
It now makes the replica files read-only and also tests Winning PoSt the same way.

storojs72 (approved):
LGTM!

cryptonemo (approved):

PR #1633, opened 2022-10-21
Use current process binding to limit cpus that threads can be bound to

clinta:
Core binding currently does not work if the lotus-worker process is limited to a subset of CPUs.

This PR prevents attempting to bind threads to a core that the process does not have access to.

### DIFF
--- a/storage-proofs-porep/src/stacked/vanilla/cores.rs
+++ b/storage-proofs-porep/src/stacked/vanilla/cores.rs
@@ -157,6 +157,7 @@ fn create_core_units(
     core_count: usize,
     group_count: usize,
     cores_per_unit: usize,
+    allowed_cores: &hwloc::CpuSet,
 ) -> Vec<Vec<usize>> {
     assert_eq!(0, core_count % group_count);
     // The number of cores that belong to a single group.
@@ -176,7 +177,7 @@ fn create_core_units(
     let core_units = (0..unit_count)
         .map(|i| {
             (0..cores_per_unit)
-                .map(|j| {
+                .filter_map(|j| {
                     // Every group gets a single unit assigned first. Only if all groups have
                     // already one unit, a second one will be assigned if possible. This would then
                     // be the second "round" of assignments.
@@ -184,10 +185,14 @@ fn create_core_units(
                     // The index of the core that is bound to a unit.
                     let core_index = (j + i * group_size) % core_count + (round * cores_per_unit);
                     assert!(core_index < core_count);
-                    core_index
+
+                    allowed_cores
+                        .is_set(core_index.try_into().ok()?)
+                        .then_some(core_index)
                 })
                 .collect::<Vec<_>>()
         })
+        .filter(|x| !x.is_empty())
         .collect::<Vec<_>>();
     debug!("Core units: {:?}", core_units);
     core_units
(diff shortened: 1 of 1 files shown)


storage-proofs-porep/src/stacked/vanilla/cores.rs
```diff
@@ -176,18 +177,22 @@ fn create_core_units(
...
                     assert!(core_index < core_count);
-                    core_index
+
+                    allowed_cores
+                        .is_set(core_index.try_into().ok()?)
+                        .then_some(core_index)
                 })
                 .collect::<Vec<_>>()
         })
+        .filter(|x| !x.is_empty())
```
vmx, line 195: This only filters out if a whole "group" is empty, right? Could you please add  a test case (or modify an existing one), when the allowed cores reduce the number of members of a group, but don't leave it completely empty? Or is this this something that never happens in reality?

storage-proofs-porep/src/stacked/vanilla/cores.rs
```diff
@@ -363,9 +378,45 @@ mod tests {
...
         );
 
-        let laptop = create_core_units(4, 1, 2);
+        let laptop = create_core_units(4, 1, 2, &(0..4).collect());
         assert_eq!(laptop, [[0, 1], [2, 3]]);
-        let laptop_not_filled = create_core_units(4, 1, 3);
+        let laptop_not_filled = create_core_units(4, 1, 3, &(0..4).collect());
         assert_eq!(laptop_not_filled, [[0, 1, 2]]);
+
+        let amd_limited_0 = create_core_units(16, 4, 4, &(0..8).collect());
```
vmx, line 386: Please add a test case for when the allowed cores are not a continuous range.

vmx (commented):
Thanks a lot for this clean and minimal patch. I left comments about the tests.

I like the additional description in your PR description, feel free to add it directly to the commit message.

vmx:
Somehow the CI didn't run. But I have hope that once you push another time, it should just work.

storage-proofs-porep/src/stacked/vanilla/cores.rs
clinta: Correct, this removes empty vecs from the top level which may exist after the filtering is done.

Test added: https://github.com/filecoin-project/rust-fil-proofs/blob/ae4019f2405a4acb21964e5027a6406c070b42eb/storage-proofs-porep/src/stacked/vanilla/cores.rs#L422-L433

clinta:
> Somehow the CI didn't run. But I have hope that once you push another time, it should just work.

Looks like it's still happening. Not sure why, as I didn't touch the Circle CI config.

clinta:
@vmx any suggestions for how to fix CI?

vmx:
Sorry for the delay. Thanks for the additional tests, that looks good. In regards to CI, I've contacted the CircleCI support as things seems to be triggered fine on other PRs. I created https://github.com/filecoin-project/rust-fil-proofs/pull/1636 in the meanwhile, to get CI running.

vmx:
> @vmx any suggestions for how to fix CI?

The reply I got from CircleCI support was:

> Can you have the user clinta sign out of CircleCI, clear their browser cache, and then log back in again before triggering another pipeline?

clinta:
Fixed the clippy and rustfmt errors, and it looks like CI is working now. Thanks for the assist. I had never setup a CircleCI account before, apparently that was required.

vmx:
> I had never setup a CircleCI account before, apparently that was required.

Interesting, I also didn't know that. That's good to know for future contributors. Thanks for looking into this!

vmx (approved):
Thanks for the PR and sorry for the delays due to travel and sickness.

PR #1634, opened 2022-10-21
Testing if a fork can trigger CI

vmx:
On some other PR the CI wasn't properly run, I hereby test if the issue was that the PR was from a fork and not from this repo. If that's that's the problem, then we should really fix it.


vmx:
It works.

Issue #1632, opened 2022-09-27
Introduce PoRepConfig::new_groth16()


vmx:
### Description

At a lot of places the `PoRepConfig` struct is created directly which almost always follows a pattern similar to:

```rust
let config = PoRepConfig {
    sector_size: SectorSize(SECTOR_SIZE_32_GIB),
    partitions: PoRepProofPartitions(
        *POREP_PARTITIONS
            .read()
            .expect("POREP_PARTITIONS poisoned")
            .get(&SECTOR_SIZE_32_GIB)
            .expect("unknown sector size") as usize,
    ),
    porep_id,
    api_version: ApiVersion::V1_1_0,
};
```

Instead introduce a `PoRepConfig::new_groth16()` which could look like this:

```rust
let config = PoRepConfig::new_groth16(SECTOR_SIZE_32_GIB, porep_id, ApiVersion::V1_1_0)
```

The reson for naming it `new_groth16` and not just `new` is, that the Halo2 work introduces [PoRepConfig::new_halo2](https://github.com/filecoin-project/rust-fil-proofs/blob/8f80afa751378aa2e113ebcbd1b5297af5a2e22e/filecoin-proofs/src/types/porep_config.rs#L59-L68).

### Acceptance criteria

`PoRepConfig::new_groth16()` is used everywhere where currently the `PoRepConfig` struct is constructed directly.

### Risks + pitfalls

I can't think of any.

### Where to begin

Introduce `PoRepConfig::new_groth16()` and use it everywhere where currently the `PoRepConfig` is used directly.

DrPeterVanNostrand:
I totally agree. Using constructors would prevent production proofs from being configured with incorrect setup values.

0xpwnstar:
/assign
I would like to work on this.

Issue #1631, opened 2022-09-27
Replace constant hashmaps with const fn


vmx:
### Description

Currently there are some constants defined as hashmaps in https://github.com/filecoin-project/rust-fil-proofs/blob/4ff161d97728fa7ba6e12d519a4c0ce0c1b08096/filecoin-proofs/src/constants.rs#L51-L120. They are wrapped in a `RwLock` which is only needed for [prodbench](https://github.com/filecoin-project/rust-fil-proofs/blob/4ff161d97728fa7ba6e12d519a4c0ce0c1b08096/fil-proofs-tooling/src/bin/benchy/prodbench.rs), which I even don't know whether anyone is still using/if it still works. So wither `prodbench` can be removed and if not, I think it doesn't really need the feature of being able to add custom constants.

This means that those hashmaps could be replaces with a constant functions. E.g. `POREP_PARTITIONS` could be replaces with something like:

```rust
pub const fn porep_partitions(sector_size: u64) -> usize {
       match sector_size {
           SECTOR_SIZE_2_KIB => 1,
           SECTOR_SIZE_4_KIB => 1,
            …
           _ => panic!("invalid sector size"),
       }
}
```

The advantage is that we don't need a `lazy_static`, but also that if you pass in an unsupported size, it will panic at compile-time, not at runtime.

This change would also remove a lot of boilerplate code. If you search for `POREP_PARTITIONS`, you'll find many occurrences of:

```rust
*POREP_PARTITIONS
    .read()
    .expect("POREP_PARTITIONS poisoned")
    .get(&SECTOR_SIZE_2_KIB)
    .expect("unknown sector size")
```

which could be replaced by `porep_partitions(SECTOR_SIZE_2_KIB)`.

### Acceptance criteria

The hashmaps mentioned above are replaced by const functions.

### Risks + pitfalls

Those are part of the public API, so crates like `filecoin-proofs-api` would also need to be adapted.

### Where to begin

Change one of those hashmaps to a const fn and get things compiled.

DrPeterVanNostrand:
It would be really cool (imho) if we made all proofs' code generic over `<const SECTOR_NODES: usize>`, which would allow Groth16's usage of the `const fn some_setup_param<const SECTOR_NODES: usize>()` pattern and would result in many simplifications around proof setup.

vmx:
It's a nice to have, but no one ever found the time, hence closing the issue.

PR #1635, opened 2022-10-26
chore: Introduce PoRepConfig::new_groth16()

hanbu97:
[https://github.com/filecoin-project/rust-fil-proofs/issues/1632](url)

### DIFF
--- a/fil-proofs-tooling/src/bin/benchy/prodbench.rs
+++ b/fil-proofs-tooling/src/bin/benchy/prodbench.rs
@@ -168,9 +168,8 @@ fn configure_global_config(inputs: &ProdbenchInputs) {
         .expect("POREP_PARTITIONS poisoned")
         .insert(inputs.sector_size_bytes(), inputs.porep_partitions);
     POREP_MINIMUM_CHALLENGES
-        .write()
-        .expect("POREP_MINIMUM_CHALLENGES poisoned")
-        .insert(inputs.sector_size_bytes(), inputs.porep_challenges);
+        .get_mut()
+        .insert(inputs.sector_size_bytes(), inputs.porep_challenges as usize);
 }
 
 pub fn run(
@@ -314,12 +313,11 @@ fn generate_params(i: &ProdbenchInputs) {
     );
     let dummy_porep_id = [0; 32];
 
-    cache_porep_params(PoRepConfig {
-        sector_size,
-        partitions,
-        porep_id: dummy_porep_id,
-        api_version: i.api_version(),
-    });
+    cache_porep_params(PoRepConfig::new_groth16(
+        i.sector_size_bytes(),
+        dummy_porep_id,
+        i.api_version(),
+    ));
 }
 
 fn cache_porep_params(porep_config: PoRepConfig) {
--- a/filecoin-proofs/src/constants.rs
+++ b/filecoin-proofs/src/constants.rs
@@ -1,5 +1,5 @@
-use std::collections::HashMap;
 use std::sync::RwLock;
+use std::{collections::HashMap, sync::RwLockWriteGuard};
 
 pub use storage_proofs_core::drgraph::BASE_DEGREE as DRG_DEGREE;
 pub use storage_proofs_porep::stacked::EXP_DEGREE;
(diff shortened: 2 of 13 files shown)


hanbu97:
https://github.com/filecoin-project/rust-fil-proofs/issues/1631
finished

shawnrader:
> Please rerun the workflow, clippy is fixed. Thank you.

I am still showing some clippy issues after re-running the workflow (some unused imports): https://app.circleci.com/pipelines/github/filecoin-project/rust-fil-proofs/4397/workflows/fd19d319-f5cf-422e-9885-8ff4b8314310/jobs/96937?invite=true#step-105-355

hanbu97:
Got it. Clippy in tests fixed.

fil-proofs-tooling/src/bin/benchy/prodbench.rs
```diff
@@ -167,10 +166,10 @@ fn configure_global_config(inputs: &ProdbenchInputs) {
         .write()
         .expect("POREP_PARTITIONS poisoned")
         .insert(inputs.sector_size_bytes(), inputs.porep_partitions);
-    POREP_MINIMUM_CHALLENGES
-        .write()
-        .expect("POREP_MINIMUM_CHALLENGES poisoned")
-        .insert(inputs.sector_size_bytes(), inputs.porep_challenges);
+    // POREP_MINIMUM_CHALLENGES
```
DrPeterVanNostrand, line 169: I think that it still makes sense to allow `benchy` to configure the number of PoRep challenges, as `benchy` is used to benchmark varrying PoRep configurations, not just the the default/production configurations.

filecoin-proofs/src/constants.rs
```diff
@@ -47,24 +47,25 @@ pub const PUBLISHED_SECTOR_SIZES: [u64; 10] = [
     SECTOR_SIZE_64_GIB,
 ];
 
+pub struct PorepMinimumChallenges;
+impl PorepMinimumChallenges {
+    pub const fn porep_partitions(sector_size: u64) -> usize {
```
DrPeterVanNostrand, line 52: I think that a better name for this function would be `from_sector_size`, however to allow for varying PoRep configurations during benchmarking (e.g. `benchy`), I think that it's better to leave this as a mutable `HashMap` rather than hardcoded constants.

filecoin-proofs/src/types/porep_config.rs
```diff
@@ -54,6 +55,22 @@ impl From<PoRepConfig> for SectorSize {
 }
 
 impl PoRepConfig {
+    /// construct PoRepConfig by groth16
+    pub fn new_groth16(sector_size: u64, porep_id: [u8; 32], api_version: ApiVersion) -> Self {
```
DrPeterVanNostrand, line 59: Until additional proof systems are supported in `rust-fil-proofs`, I think that it's better to exclude the "groth16` identifier from the function name as Groth16 is the only proof system currently supported. If/when additional proof systems are supported, I definitely think that it makes sense to use the function name `new_groth16`, but for now, I think that `new` is a better name.

DrPeterVanNostrand (changes requested):
Thank you!

I definitely like the idea of having a simple `PorepConfig` constructor.

My only concerns with this PR are:
1. `PorepConfig::new_groth16` should be named `new` until additional proof systems are supported.
2. Minimum PoRep challenges should be kept as a mutable setting to allow `benchy` to benchmark non-default PoRep configurations.

vmx:
>     1. `PorepConfig::new_groth16` should be named `new` until additional proof systems are supported.

I proposed in the original issue at https://github.com/filecoin-project/rust-fil-proofs/issues/1632 naming it `new_groth16()`. I still think it's a good idea (and less hassle for the future), but if @DrPeterVanNostrand feels strongly about it, we can also change it to just `new()`.

storojs72 (dismissed):
I support this change (though it still requires to fix merge conflicts), as it removes more lines of code than adds (+88; -218), that makes codebase a bit more readable.

hanbu97:
What should I do next, or some tasks else.

vmx:
@hanbu97 The comment about making things again a mutable HashMap is something you could work on. I'll talk with @DrPeterVanNostrand directly whether it should be `new` or `new_groth16` and let you know once we've come to a decision.

vmx (approved):
Thanks a lot, I really like the changes!

filecoin-proofs/src/constants.rs
```diff
@@ -47,24 +47,47 @@ pub const PUBLISHED_SECTOR_SIZES: [u64; 10] = [
...
+            .copied()
+            .collect(),
+        ))
+    }
+
+    pub fn get_mut(&self) -> RwLockWriteGuard<'_, HashMap<u64, usize>> {
+        self.0.write().expect("POREP_MINIMUM_CHALLENGES poisoned")
+    }
+
+    pub fn porep_partitions(&self, sector_size: u64) -> usize {
```
DrPeterVanNostrand, line 76: To me, the name of this function implies that we are deriving the partition count from the number of porep challenges. I think that a better name for this function would be something like `from_sector_size` or `for_sector_size` as we are retrieving the minimum challenge count for a given sector size.

hanbu97: I agree with you, will make it more readable. Changes done.

vmx (approved):
Looks good to me. I'm just waiting for @DrPeterVanNostrand for final approval.

filecoin-proofs/src/constants.rs
DrPeterVanNostrand: Awesome, thank you!

DrPeterVanNostrand:
Looks good to me. Thanks @hanbu97!

vmx (approved):
As @DrPeterVanNostrand also approved it, I hereby approve it.

@hanbu97 could you please rebase one more time? Best would be if you would squash it into a single commit.

DrPeterVanNostrand (approved):

hanbu97:
> @hanbu97 could you please rebase one more time? Best would be if you would squash it into a single commit.

When I tried to squash commits. Some error occurs.
> Unable to squash. Squashing replays all commits up to the last one required for the squash. A merge commit cannot exist among those commits.

It seems because I merged some changes at developing time. I will pay attention to it next time. If the squash is required I may create a new branch.

vmx:
@hanbu97 I dared to squash it into a single commit for you. Please let me know if you agree with that commit and the commit message.

I did a `git rebase -i master` and the fixed up the commits.

shawnrader (approved):

hanbu97:
Sure go ahead. Let's make it done. Thank you.

vmx (approved):

vmx:
Thanks again @hanbu97 and sorry that it took so long to get it merged.

PR #1636, opened 2022-11-07
(clone) use current process binding to limit thread cores

vmx:
*Note*: I create another PR in hope that it triggers the CI correctly. This is a clone of https://github.com/filecoin-project/rust-fil-proofs/pull/1633.

---

Use the current processes bound cores to limit the possible cores that threads can be bound to. This allows core binding to work properly when the lotus-worker service is limited to certain CPUs by cgroups.


vmx:
https://github.com/filecoin-project/rust-fil-proofs/pull/1633 was merged.

PR #1660, opened 2023-02-24
optimize: copy_parents_data only needs base parents

zhiqiangxu:
[`copy_parents_data_inner`](https://github.com/filecoin-project/rust-fil-proofs/blob/master/storage-proofs-porep/src/stacked/vanilla/graph.rs#L224) only uses base parents, so there's no need to fetch extension parents at the caller side.

### DIFF
--- a/storage-proofs-porep/src/stacked/vanilla/graph.rs
+++ b/storage-proofs-porep/src/stacked/vanilla/graph.rs
@@ -174,9 +174,9 @@ where
             let cache_parents = cache.read(node)?;
             Ok(self.copy_parents_data_inner(&cache_parents, base_data, hasher))
         } else {
-            let mut cache_parents = [0u32; DEGREE];
+            let mut cache_parents = [0u32; BASE_DEGREE];
 
-            self.parents(node as usize, &mut cache_parents[..])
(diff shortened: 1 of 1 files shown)


vmx (dismissed):
Thanks! That looks correct to me. But I'd like to have other team members double-check as this is such a central piece.

@zhiqiangxu Thanks for your contributions. You also always put helpful context in your PR description. It would be cool if you could also put it into the commit message next time (then it even automatically becomes the PR description).

zhiqiangxu:
> Thanks! That looks correct to me. But I'd like to have other team members double-check as this is such a central piece.
> 
> @zhiqiangxu Thanks for your contributions. You also always put helpful context in your PR description. It would be cool if you could also put it into the commit message next time (then it even automatically becomes the PR description).

I just modified the commit message and re-pushed the branch :)

DrPeterVanNostrand (dismissed):
Thanks for the PR. Looks good to me.

cryptonemo (approved):

vmx:
@zhiqiangxu sorry for taking so long, this PR slipped my attention. It's now merged.

Issue #1270, opened 2020-09-01
Window PoSt challenge generation should not depend on sector order


porcuquine:
### Description

Window PoSt challenge generation currently depends on the ordering of provided sectors.

This is considered to be a 'protocol bug' in that this is the behavior which has been specified and audited. However, it allows for miners to influence challenges by shuffling their sectors. One possible (and probably cleanest) solution is to make challenge-generation independent of sector ordering — at the cost of a breaking change (so extant Window PoSts would not verify with the new code, requiring a network reset or proofs upgrade).

In the current code, the function `generate_leaf_challenges` *appears* to do the right thing, but this function is only called from Election PoSt so is not actually used. https://github.com/filecoin-project/rust-fil-proofs/blob/4a07a868d077f5926b23c3f913f99d8b46b8cebc/storage-proofs/post/src/fallback/vanilla.rs#L192-L211

Notice, however, that in `generate_public_input` for Fallback PoSt, `challenge_index` depends on `i` (the sector's index within the submitted partition). This must *not* be the case. https://github.com/filecoin-project/rust-fil-proofs/blob/4a07a868d077f5926b23c3f913f99d8b46b8cebc/storage-proofs/post/src/fallback/compound.rs#L68-L76 Instead, the `challenge_index` can just be `n`. Each sector's challenge indexes should range from 0 to `challenge_count`.

The same can be seen in the vanilla proof: https://github.com/filecoin-project/rust-fil-proofs/blob/4a07a868d077f5926b23c3f913f99d8b46b8cebc/storage-proofs/post/src/fallback/vanilla.rs#L324-L333

### Acceptance criteria

Window PoSt challenge generation does not depend on the ordering of provided sectors.

### Risks + pitfalls

This will be a *breaking change* to proofs: a proof generated with the previous Window PoSt will not verify. Therefore, we will *either*:
- need a network reset; or
- need a proofs upgrade (with new RegisteredProof for modified Window PoSt).

This decision must be made at the project level, and we can support either path. If the proofs upgrade is selected, a new issue should be created; and both that new issue and the current should be addressed in the PR realizing this work.

Note that *because Winning PoSt uses only a single partition*, this change should not affect Winning PoSt challenge generation. Therefore, we don't technically need a new RegisteredProof. We should at least verify that this is true if we opt for a proofs upgrade.

[...]

PR #1661, opened 2023-02-24
feat: fix the open 'grindability issue'

cryptonemo:
Some more details here:
https://github.com/filecoin-project/rust-fil-proofs/issues/1270

### DIFF
--- a/filecoin-proofs/src/api/post_util.rs
+++ b/filecoin-proofs/src/api/post_util.rs
@@ -103,22 +103,24 @@ pub fn generate_fallback_sector_challenges<Tree: 'static + MerkleTreeTrait>(
 
             for n in 0..post_config.challenge_count {
                 let challenge_index = match post_config.typ {
-                    PoStType::Window => get_challenge_index(
-                        post_config.api_version,
-                        partition_index,
-                        i,
-                        num_sectors_per_chunk,
-                        post_config.challenge_count,
-                        n,
-                    ),
-                    PoStType::Winning => get_challenge_index(
-                        post_config.api_version,
-                        partition_index,
-                        i,
-                        post_config.sector_count,
-                        post_config.challenge_count,
-                        n,
-                    ),
+                    PoStType::Window => {
+                        let sector_index = partition_index * num_sectors_per_chunk + i;
+                        get_challenge_index(
+                            post_config.api_version,
+                            sector_index,
+                            post_config.challenge_count,
+                            n,
+                        )
+                    }
+                    PoStType::Winning => {
+                        let sector_index = partition_index * post_config.sector_count + i;
+                        get_challenge_index(
+                            post_config.api_version,
+                            sector_index,
+                            post_config.challenge_count,
+                            n,
+                        )
+                    }
                 };
                 let challenged_leaf = generate_leaf_challenge(
                     &public_params,
--- a/filecoin-proofs/src/api/post_util.rs
+++ b/filecoin-proofs/src/api/post_util.rs
@@ -9,7 +9,9 @@ use log::{debug, info};
 use storage_proofs_core::{
     cache_key::CacheKey, merkle::MerkleTreeTrait, proof::ProofScheme, sector::SectorId,
 };
-use storage_proofs_post::fallback::{self, generate_leaf_challenge, FallbackPoSt, SectorProof};
+use storage_proofs_post::fallback::{
+    self, generate_leaf_challenge, get_challenge_index, FallbackPoSt, SectorProof,
+};
 
 use crate::{
     api::as_safe_commitment,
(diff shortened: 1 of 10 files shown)


storage-proofs-post/src/fallback/utils.rs
```diff
@@ -0,0 +1,19 @@
+use storage_proofs_core::api_version::ApiVersion;
+
+/// Selects the challenge index used to determine the leaf challenge for PoSt
+pub fn get_challenge_index(
+    api_version: ApiVersion,
+    sector: usize,
+    sector_chunk_index: usize,
+    num_sectors_per_chunk: usize,
+    challenge_count: usize,
+    challenge_index: usize,
```
DrPeterVanNostrand, line 10: I would maybe change these argument names to something like:

```rust
fn get_challenge_index(
    api_version,
    sector_chunk_index,
    num_sectors_per_chunk,
    sector_index_in_chunk.
    challenge_count_per_sector,
    challenge_index,
)
```
or

```rust
fn get_challenge_index(
    api_version,
    sector_index,
    challenge_count_per_sector,
    challenge_index,
)
```
where the caller of `get_challenge_index` passes in the sector index in the PoSt proof `sector_index = chunk_index * num_sectors_per_chunk + sector_index_in_chunk`.

DrPeterVanNostrand (approved):
Looks good to me. My only optional change would be the naming used in `get_challenge_index`.

storage-proofs-post/src/fallback/utils.rs
cryptonemo: Nice, yeah I think something like that will be cleaner :+1:

filecoin-proofs/tests/api.rs
```diff
@@ -52,132 +52,164 @@ use filecoin_proofs::{
...
 
-    let mut porep_id = [0u8; 32];
-    porep_id[..8].copy_from_slice(&porep_id_v1_1.to_le_bytes());
-    assert!(!is_legacy_porep_id(porep_id));
-    seal_lifecycle::<SectorShape2KiB>(SECTOR_SIZE_2_KIB, &porep_id, ApiVersion::V1_1_0)
-}
+        assert!(match api_version {
+            ApiVersion::V1_0_0 => is_legacy_porep_id(porep_id),
+            ApiVersion::V1_1_0 | ApiVersion::V1_2_0 => !is_legacy_porep_id(porep_id),
+        });
```
vmx, line 82: I propose refactoring this into a function like:

```rust
fn to_porepid_verified(registered_seal_proof: u64, api_version: ApiVersion) -> [u8; 32] {
    let mut porep_id = [0u8; 32];
    porep_id[..8].copy_from_slice(&registered_seal_proof.to_le_bytes());
    assert!(match api_version {
        ApiVersion::V1_0_0 => is_legacy_porep_id(porep_id),
        ApiVersion::V1_1_0 | ApiVersion::V1_2_0 => !is_legacy_porep_id(porep_id),
    }
    porep_id
}
```

Which then can be used (with the comments above applied) within this loop as:

```rust
let porep_id = to_porepid_verified(registered_seal_proof, api_version);
```

storage-proofs-post/src/fallback/utils.rs
```diff
@@ -0,0 +1,14 @@
...
+/// Selects the challenge index used to determine the leaf challenge for PoSt
+pub fn get_challenge_index(
+    api_version: ApiVersion,
+    sector_index: usize,
+    challenge_count_per_sector: usize,
+    challenge_index: usize,
+) -> u64 {
+    (match api_version {
+        ApiVersion::V1_2_0 => challenge_index,
+        _ => sector_index * challenge_count_per_sector + challenge_index,
```
vmx, line 12: I propose making this exhaustive with using all available `ApiVersion`. This way it will fail at compile time when a new one is added. In the current code, a new `ApiVersion` would silently use the non v1_2_0, although I guess a v1_3_0, should use the v1_2_0 one.

vmx (changes requested):
Only minor things, except for the one comment about the exhaustive pattern matching.

storage-proofs-post/src/fallback/utils.rs
cryptonemo: Thanks, I took a shot at applying all review feedback!

cryptonemo:
Fun, the openssl-src package is failing clippy and causing CI issues :sweat:

cryptonemo:
> Fun, the openssl-src package is failing clippy and causing CI issues sweat

Hrm, we could just update `rust-toolchain` to something later like `1.67.1`.  Not a bad idea IMO, but then we have a host of other clippy issues incoming.  Let me see about a separate PR updating it and then possibly rebasing this after.

storage-proofs-post/src/fallback/utils.rs
```diff
@@ -0,0 +1,16 @@
...
+/// Selects the challenge index used to determine the leaf challenge for PoSt
+pub fn get_challenge_index(
+    api_version: ApiVersion,
+    sector_index: usize,
+    challenge_count_per_sector: usize,
+    challenge_index: usize,
+) -> u64 {
+    (match api_version {
+        ApiVersion::V1_2_0 => challenge_index,
+        ApiVersion::V1_0_0 | ApiVersion::V1_1_0 => {
```
vmx, line 12: Nitpick (feel free to keep it like that). I would expect the cases to be in the order of the versions => V1_2_0 after V1_0_0.

vmx (dismissed):

DrPeterVanNostrand (dismissed):

cryptonemo:
Resolves #1270

filecoin-proofs/src/api/post_util.rs
```diff
@@ -100,9 +102,16 @@ pub fn generate_fallback_sector_challenges<Tree: 'static + MerkleTreeTrait>(
             let mut challenges = Vec::new();
 
             for n in 0..post_config.challenge_count {
-                let challenge_index = ((partition_index * post_config.sector_count + i)
-                    * post_config.challenge_count
-                    + n) as u64;
+                let sector_index = match post_config.typ {
+                    PoStType::Window => partition_index * num_sectors_per_chunk + i,
+                    PoStType::Winning => partition_index * post_config.sector_count + i,
```
DrPeterVanNostrand, line 107: Maybe in the future (not in this PR) we should just put an assertion that if post-type is Winning-PoSt, that `post_config.sector_count == pub_sectors.len() == 1`, so `sector_index = 0` can be hardcoded.

DrPeterVanNostrand (approved):

vmx (approved):

Kubuxu (approved):
SGTM

PR #1663, opened 2023-02-27
feat(storage-proofs-post): add grindability tests

DrPeterVanNostrand:
- Adds Fallback-PoSt tests for version `ApiVersion::V1_2_0`.
- Tests that the challenge generation algorithm in `ApiVersion::V1_2_0` is incompatible with previous versions.


cryptonemo (approved):
Good catch, thanks!

vmx (approved):

PR #1670, opened 2023-03-03
feat: add API features

DrPeterVanNostrand:
This PR allows for optional API features to be enabled within a supported range of `ApiVersion`s.

This PR is currently a draft because:
- It should be merged into master after the grindability fix PR [#1661](#1661) is merged.
- `enum ApiFeature` contains no variants, i.e. this PR does not add prospective API features such as Synthetic-PoRep

### DIFF
--- a/storage-proofs-core/src/api_version.rs
+++ b/storage-proofs-core/src/api_version.rs
@@ -32,6 +32,20 @@ impl ApiVersion {
             ApiVersion::V1_2_0 => Version::new(1, 2, 0),
         }
     }
+
+    #[inline]
+    pub fn supports_feature(&self, feat: &ApiFeature) -> bool {
+        self >= &feat.first_supported_version()
+            && feat
+                .last_supported_version()
+                .map(|v_last| self <= &v_last)
+                .unwrap_or(true)
+    }
+
+    #[inline]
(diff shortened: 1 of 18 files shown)


vmx (commented):
I wonder if it would make sense to make the `PoRepConfig` API improvements a separate PR that we can merge already. And then making the API features PR a distinct one?

I don't want to cause lots of additional work, so take this as an idea, rather then a call for action.

storage-proofs-core/src/api_version.rs
```diff
@@ -56,6 +83,31 @@ impl FromStr for ApiVersion {
...
+
+impl ApiFeature {
+    #[inline]
+    pub fn first_supported_version(&self) -> ApiVersion {
+        /*
+        match self {
+            _ => unimplemented!(),
+        }
+        */
+        unimplemented!();
```
vmx, line 97: Some left-overs, but they also don't bother me to much.

vmx (dismissed):

cryptonemo (dismissed):

vmx (approved):

cryptonemo:
@DrPeterVanNostrand Please merge today

PR #1677, opened 2023-03-09
chore: update release.toml config

vmx:
This way releases can be made with the most recent version 0.25.5 of cargo-release. New major releases can be cut with simply running:

    cargo release major --execute

Prior to this change `fil-proofs-tooling` wasn't tagged. With this change it's tagged just like any other crates. That adds more consistency. It still isn't published to crates.io as intended.

The `consolidate-commits = false` option makes sure that each crate gets its own tag with a commit message that contains its version number. Without this setting, it would only say "Release" without further information.

### DIFF
--- a/fil-proofs-tooling/release.toml
+++ b/fil-proofs-tooling/release.toml
@@ -1,4 +1,0 @@
-disable-push = true
-disable-publish = true
-disable-tag = true
-no-dev-version = true
--- a/release.toml
+++ b/release.toml
@@ -1,3 +1,1 @@
-pre-release-commit-message = "chore({{crate_name}}): release {{version}}"
-pro-release-commit-message = "chore({{crate_name}}): starting development cycle for {{next_version}}"
-no-dev-version = true
+consolidate-commits = false


vmx:
I should note that this PR preserves the current behaviour. Those newer versions of cargo-release make it possible to have a single tag/commit for everything. As we release all version at once we might even want that. I leave the decision to @cryptonemo.

vmx:
I think this one is still worth having, in case someone wants to cut a release with a newer cargo-release version.

cryptonemo:
> I think this one is still worth having, in case someone wants to cut a release with a newer cargo-release version.

I cut releases with newer versions just fine

vmx:
> I cut releases with newer versions just fine

I get:

```console
$ cargo release --version
cargo-release 0.25.10
$ cargo release
error: Failed to parse `/home/vmx/src/pl/filecoin/rust-fil-proofs/release.toml`: TOML parse error at line 2, column 1
  |
2 | pro-release-commit-message = "chore({{crate_name}}): starting development cycle for {{next_version}}"
  | ^^^^^^^^^^^^^^^^^^^^^^^^^^
unknown field `pro-release-commit-message`, expected one of `allow-branch`, `sign-commit`, `sign-tag`, `push-remote`, `registry`, `release`, `publish`, `verify`, `owners`, `push`, `push-options`, `shared-version`, `consolidate-commits`, `pre-release-commit-message`, `pre-release-replacements`, `pre-release-hook`, `tag-message`, `tag-prefix`, `tag-name`, `tag`, `enable-features`, `enable-all-features`, `dependent-version`, `metadata`, `target`, `rate-limit`, `certs-source`
```

vmx:
> I don't usually use the github buttons because rebases are pretty easy, but maybe the 'Update with rebase' will do the trick?

I just rebased it locally, that was easy enough.

cryptonemo (approved):

Issue #1652, opened 2022-12-08
Make it possible to build for aarch64 on stable Rust 


vmx:
### Description

Currently you need a nightly Rust version to build for aarch64. It's just an optimization, but it's fully functional without. Hence I propose that we check for a nightly version similar to [what we do in bellperson](https://github.com/filecoin-project/bellperson/pull/269). This way we can e.g. use the stable Rust for CI.

This issue was triggered by the build failure at https://github.com/filecoin-project/rust-fil-proofs/pull/1651 and https://github.com/filecoin-project/filecoin-ffi/pull/345.

### Acceptance criteria

A successful build on aarch64 on a stable Rust version

### Risks + pitfalls

Getting things wired up correctly with the build.rs and checking for that feature might need a bit of fiddling.

Once done the `filecoin-ffi` project needs to be notified, that they don't need a nightly build anymore. Though they might want to put a note somewhere that using nightly on aarch64 will enable optimizations.

### Where to begin

Look at https://github.com/filecoin-project/bellperson/pull/269 for inspiration and do a similar thing for this repo.

PR #1679, opened 2023-03-10
fix: make builds on stable and aarch64 possible

vmx:
When using Rust nightly we are able to enable an optimization for aarch64.
With this change it's now possible to compile for aarch64 on stable rust,
while the optimization is still enabled when compiled with nightly.

A similar fix was made for bellperson at
https://github.com/filecoin-project/bellperson/pull/269

This fix can be tested even on x86_64 on Linux with cross compiling
for aarch64:

    cargo build --target aarch64-unknown-linux-gnu

Fixes https://github.com/filecoin-project/rust-fil-proofs/issues/1652.

### DIFF
--- a/storage-proofs-porep/src/stacked/vanilla/macros.rs
+++ b/storage-proofs-porep/src/stacked/vanilla/macros.rs
@@ -55,11 +55,14 @@ macro_rules! prefetch {
 
             x86::_mm_prefetch($val, x86::_MM_HINT_T0);
         }
-        #[cfg(target_arch = "aarch64")]
+        #[cfg(all(nightly, target_arch = "aarch64"))]
         unsafe {
             use std::arch::aarch64::{_prefetch, _PREFETCH_LOCALITY3, _PREFETCH_READ};
             _prefetch($val, _PREFETCH_READ, _PREFETCH_LOCALITY3);
         }
+        // Make sure that there are not warnings about unused variables.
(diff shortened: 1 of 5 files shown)


storage-proofs-porep/src/stacked/vanilla/macros.rs
```diff
@@ -55,11 +55,14 @@ macro_rules! prefetch {
...
         }
-        #[cfg(target_arch = "aarch64")]
+        #[cfg(all(nightly, target_arch = "aarch64"))]
         unsafe {
             use std::arch::aarch64::{_prefetch, _PREFETCH_LOCALITY3, _PREFETCH_READ};
             _prefetch($val, _PREFETCH_READ, _PREFETCH_LOCALITY3);
         }
+        // Make sure that there are not warnings about unused variables.
+        #[cfg(all(not(nightly), target_arch = "aarch64"))]
+        let _ = ($val);
```
DrPeterVanNostrand, line 65: I would maybe add a comment here (at least to me, this is non-obvious code) as to why there is a branch point for nightly, e.g. `When compiling for aarch64, we can only perform the prefetch optimization when nightly features are enabled.`

DrPeterVanNostrand (changes requested):
Looks good to me, just one comment regarding documentation.

cryptonemo (approved):

DrPeterVanNostrand (approved):

PR #1681, opened 2023-03-13
feat(filecoin-proofs): allow PC1 unreplicated data to be /dev/zero

### LINKED ISSUE #1664: seal_pre_commit_phase1 should allow /dev/zero as in_file
Currently passing /dev/zero to PC1 will fail with `in_path must be a file`. This would be a good optimization for CC sectors.

Bonus points if proofs would skip TreeD compute for all-zero data since we can easily precompute all levels.

DrPeterVanNostrand:
Allows PC1's unreplicated data file to be `/dev/zero`.

Closes #1664 (however this PR does not include skipping TreeD building for empty sectors, which can be handled in another PR).

### DIFF
--- a/filecoin-proofs/src/api/seal.rs
+++ b/filecoin-proofs/src/api/seal.rs
@@ -67,10 +67,18 @@ where
 {
     info!("seal_pre_commit_phase1:start: {:?}", sector_id);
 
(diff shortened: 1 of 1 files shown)


cryptonemo (approved):

cryptonemo:
Merge at will

Issue #1667, opened 2023-02-28
Additional data encode/decode methods


magik6k:
Currently methods for encoding/decoding data to/from sectors are quite limited.
* Unseal/UnsealRange allow reading unsealed data from sealed sectors, but require TreeR generation, which is slow and expensive, and cannot be stored
  * Even if there was a way to generate TreeR/SectorKey (https://github.com/filecoin-project/rust-fil-proofs/issues/1666), there is no method to decode data from the sealed sector.
  * If that method exists ideally it would look like UnsealRange, taking sealed/key streams, which would allow unsealing services to generate sector key before fetching data from SPs, and decode (parts of) data with the key in a streaming fashion after getting the key.
* Snapdeals encode/decode methods could also use streaming support.

magik6k:
This is needed to enable https://github.com/filecoin-project/lotus/issues/10316

cryptonemo:
Noting this is for normal sectors, not snapdeals

DrPeterVanNostrand:
ELI5: Lotus wants an API for encoding/decoding of "normal" (i.e. non CC) sectors.
- Accept a range of unsealed data and encoding key and return sealed data.
- Accept a range of sealed data and encoding key and return unsealed data.
- Lotus would like proofs API to accept file descriptors (e.g. to [un]sealed data and sealing keys).

magik6k:
Apparently this is quite easy to do in Go, approximate code that may work:
```go
import (
	"github.com/consensys/gnark-crypto/ecc/bls12-381/fr"
)

func encode(keyBytes []byte, valueBytes []byte) fr.Element {
	// Convert bytes to field elements
	key := new(fr.Element).SetBytes(keyBytes[:])
	value := new(fr.Element).SetBytes(valueBytes[:])

	// Add the field elements
	key.Add(key, value)

	return *key
}
```

We should probably test that + a snapdeals variant before adding rust-side APIs which may turn out to be redundant

vmx:
SnapDeals decode is merged and available through the FFI via https://github.com/filecoin-project/filecoin-ffi/pull/402. Normal replica encode/decode will be implemented directly in Go.

PR #1682, opened 2023-03-15
feat(filecoin-proofs): add APIs for retrieving sealing and sector update keys

### LINKED ISSUE #1666: Method to generate sector key
There seems to be no good way to generate a sector key today - which is needed for snapdeals unsealing, and possibly for future unsealing service-type things.

One could call Precommit1 and Precommit2 with null bytes (PC1 alone doesn't generate the sector key - https://github.com/filecoin-project/rust-fil-proofs/issues/1665)

DrPeterVanNostrand:
This PR adds:
- `SealPreCommitPhase1Output::sealing_key_path` - returns path to "sealing key" (i.e. last layer)
- `SealPreCommitPhase1Output::read_sealing_key_range` - returns a range of nodes of the sealing key
- `filecoin_proofs::api::update::read_sector_key_range` - returns a range of nodes of the "sector update key" (i.e. the originally sealed sector, also called "old replica" within the context of sector updating)

Closes #1666
(and relates to #1667)

### DIFF
--- a/filecoin-proofs/src/api/update.rs
+++ b/filecoin-proofs/src/api/update.rs
@@ -1,4 +1,4 @@
-use std::fs::{self, File};
+use std::fs::{self, File, OpenOptions};
 use std::io::Write;
 use std::path::Path;
 
@@ -7,6 +7,7 @@ use bincode::{deserialize, serialize};
 use filecoin_hashers::{Domain, Hasher};
 use generic_array::typenum::Unsigned;
 use log::{info, trace};
+use memmap2::MmapOptions;
 use merkletree::merkle::get_merkle_tree_len;
 use merkletree::store::StoreConfig;
 use storage_proofs_core::{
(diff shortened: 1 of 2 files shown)


DrPeterVanNostrand:
Hi @magik6k, if you have a chance, could you offer any feedback on this key retrieval API from a Lotus perspective?

Is retrieving a range of nodes from the sector update key sufficient, or would a different API be better for Lotus? I remember that you mentioned using file descriptors to retrieve keys, however that would require quite a bit of API refactoring.

filecoin-proofs/src/api/update.rs
```diff
@@ -281,6 +282,28 @@ pub fn remove_encoded_data<Tree: 'static + MerkleTreeTrait<Hasher = TreeRHasher>
...
+pub fn read_sector_key_range<Tree: MerkleTreeTrait>(
+    replica_old_path: &Path,
+    node_range: std::ops::Range<usize>,
+) -> Result<Vec<<Tree::Hasher as Hasher>::Domain>> {
+    let replica_old_file = OpenOptions::new()
+        .read(true)
+        .open(replica_old_path)
+        .with_context(|| format!("could not open path={:?}", replica_old_path))?;
+    let replica_old = unsafe {
+        MmapOptions::new()
```
cryptonemo, line 295: Might make sense to mmap just the range that's required here

magik6k (commented):
I don't think the key-range-reading methods are useful by themselves to lotus / other proofs api consumers. For #1667 I'd imagine the proofs api would just have methods taking os pipes with (fr32-aligned) streams of sealed data + sector key data, and would get a decoded stream of unsealed data.

`sealing_key_path` Looks good, and should make fixing snapdeals unsealing much less hacky.

DrPeterVanNostrand:
Replaced by PR #1687

PR #1684, opened 2023-03-21
fix: remove DRG PoRep

vmx:
The DRG PoRep was never used, hence remove it as it is only maintenance burden without adding any value.


DrPeterVanNostrand (dismissed):
I'm happy with removing DRG-PoRep. I think that it still exists mainly for posterity's sake. A more minimal change would be to enable it with a feature flag (off by default); but either are fine with me. And as you said, if for some reason DRG-PoRep needs to be referenced in the future, it will always exist in git history.

cryptonemo (approved):

PR #1685, opened 2023-03-22
fix(storage-proofs-porep): return last layer from Labels

DrPeterVanNostrand:
An off-by-one error had caused `Labels::labels_for_last_layer` to return the second last layer.


cryptonemo (approved):
Thanks!

PR #1687, opened 2023-03-24
feat(filecoin-proofs): add encode/decode range apis

### LINKED ISSUE #1666: Method to generate sector key

DrPeterVanNostrand:
Adds API functions for:
- Retrieving sealing key file
- Retrieving node ranges of sealing keys, data/unencoded files, and replica/encoded files (PoRep and SnapDeals)
- Performing PoRep and SnapDeals encoding/decoding for ranges of nodes

Closes #1666, #1667
Replaces #1682

### DIFF
--- a/filecoin-proofs/src/api/seal.rs
+++ b/filecoin-proofs/src/api/seal.rs
@@ -1,5 +1,6 @@
 use std::fs::{self, metadata, File, OpenOptions};
 use std::io::Write;
+use std::ops::Range;
 use std::path::{Path, PathBuf};
 
 use anyhow::{ensure, Context, Result};
@@ -32,7 +33,10 @@ use storage_proofs_porep::stacked::{
 
 use crate::POREP_MINIMUM_CHALLENGES;
 use crate::{
-    api::{as_safe_commitment, commitment_from_fr, get_base_tree_leafs, get_base_tree_size},
+    api::{
+        as_safe_commitment, commitment_from_fr, get_base_tree_leafs, get_base_tree_size,
+        read_nodes_from_file,
+    },
     caches::{
         get_stacked_params, get_stacked_srs_key, get_stacked_srs_verifier_key,
         get_stacked_verifying_key,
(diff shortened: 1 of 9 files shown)


storage-proofs-porep/src/stacked/mod.rs
```diff
@@ -1,3 +1,5 @@
+pub use crate::encode::{decode, encode};
```
vmx, line 1: To me it feels a bit strange to make things public from higher up the hierarchy. What about making the `mod encode` in `src/lib.rs` public instead?

filecoin-proofs/tests/api.rs
```diff
@@ -2121,3 +2132,99 @@ fn test_aggregate_proof_encode_decode() -> Result<()> {
...
+fn test_sealing_encode_decode_ranges<R: Rng, Tree: MerkleTreeTrait>(
+    rng: &mut R,
+    sector_size: u64,
+    pc1_output: &SealPreCommitPhase1Output<Tree>,
+    replica_path: &Path,
+) -> Result<()> {
+    let sector_nodes = (sector_size >> 5) as usize;
+    let node_range_len = 11;
+
+    for _ in 0..5 {
```
vmx, line 2145: Those are arbitrary values, aren't they? I'd put in a comment that they are arbitrary.

vmx (commented):
I see that lots of functions are public. Do they all need to public (honest question, I haven't checked myself)? I generally strive for as little public API as possible, so that we can make changes more easily without breaking things.

DrPeterVanNostrand:
@magik6k I updated the key retrieval API to include partial PoRep and SnapDeals encoding/decoding. Could you please take a look at these API additions when you have a moment?

filecoin-proofs/src/api/seal.rs
```diff
@@ -556,6 +560,92 @@ pub fn seal_commit_phase2<Tree: 'static + MerkleTreeTrait>(
...
+                .map(|data_node| storage_proofs_porep::encode(key, data_node))
+        })
+        .collect()
+}
+
+/// Decodes a range of replica nodes (stored on disk) into unreplicated data nodes using the last
+/// sealing layer as the encoding key.
+pub fn decode_replica_range<Tree: MerkleTreeTrait>(
+    pc1_output: &SealPreCommitPhase1Output<Tree>,
+    replica_path: &Path,
```
magik6k, line 633: I'd prefer if encode/decode just operated on byte streams. Lotus / other programs can make sure that those byte streams are properly aligned at node boundaries - those APIs should work with pipe FDs.

This is needed for cases where the last sealing layer, and replica are on different computers, and we want to decode bits of data on the fly

magik6k: It would also be preferable if `pc1_output: &SealPreCommitPhase1Output` was specified as an FD / Path to the last layer / sectorkey

filecoin-proofs/src/api/util.rs
```diff
@@ -38,3 +45,27 @@ pub fn get_base_tree_size<Tree: MerkleTreeTrait>(sector_size: SectorSize) -> Res
...
+
+    let byte_offset = node_range.start * NODE_SIZE;
+    let byte_len = (node_range.end - node_range.start) * NODE_SIZE;
+    let bytes = unsafe {
+        MmapOptions::new()
+            .offset(byte_offset as u64)
+            .len(byte_len)
+            .map(&file)
+            .with_context(|| format!("failed to mmap range of file: {}", path.display()))?
+    };
```
magik6k, line 64: Oh, that probably makes supporting data over pipes.. hard

I guess APIs which operate on buffers are also fine, we can manage data reading on Go side - passing large enough buffers over Cgo should make the overhead small enough to be mostly ignored.

magik6k (commented):
I'm not a fan of requiring data to be in files, this makes those methods significantly less useful (or at least much less efficient)

Would it be possible to make those methods just operate on byte buffers?

vmx:
> Would it be possible to make those methods just operate on byte buffers?

@magik6k Do I then understand correctly that you don't really function to do it on specific files, but you rather need access to the functionality to encode/decode the way we do it for replicas and for the snapped data. Would it then be more of an FFI API to [`encode`/`decode`](https://github.com/filecoin-project/rust-fil-proofs/blob/963eb08e55b9700657d8533418bf6332ac77d197/storage-proofs-porep/src/encode.rs) (resp. the same for snapped data)?

If you then have access to the buffers on the Go side. Might it make sense to those operations directly in Go instead of calling out over the FFI? I think there are high performance libraries out there to do field arithmetic in Go.

vmx:
> some of the methods could be made public for integration tests only.

Would it be possible to move tests into the files then? I prefer if only methods that need to be public are public and not making the public just for testing. The reason is that technically it would be a breaking change whenever we change them.

cryptonemo (approved):

vmx (changes requested):
I had a quick chat with @magik6k. API wise what he is after an API for doing encoding/decoding with buffers and/or with pipes.

I have no idea how hard it would be to do it with pipes, so for simplicity, I propose doing it with buffers for now and going from there.

cryptonemo:
> I had a quick chat with @magik6k. API wise what he is after an API for doing encoding/decoding with buffers and/or with pipes.
> 
> I have no idea how hard it would be to do it with pipes, so for simplicity, I propose doing it with buffers for now and going from there.

I had discussed this with @DrPeterVanNostrand previously, and given that all of our existing APIs are path based, it makes sense to first add those.  From the FFI, there are some APIs that support pipes, but meeting the criteria of what's needed and implementing the most optimal solutions aren't the same thing here.

It's intrusive to export the descriptors as originally requested, and even buffers could be a next step IMO.

DrPeterVanNostrand:
I think that the requested FD-based FFI is possible given some changes are made to this PR.

What if something like the following was added to `filecoin-ffi` for both pairs (Sealing and SnapDeals) of encoding/decoding functions?

```rust
#[ffi_export]
unsafe fn decode_replica_range(
    replica_fd: libc::c_int,
    sealing_key_fd: libc::c_int,
    num_nodes: usize,
) -> Result<Vec<[u8; 32]>> {
    // This function assumes that both the replica and key files are open and have been seeked to the desired node offset.
    let replica_file = File::from_raw_fd(replica_fd);
    let key_file = File::from_raw_fd(sealing_key_fd);

    let decoded: Vec<[u8; 32]> =
        filecoin_proofs::decode_replica_range(replica_file, key_file, num_nodes)
            .map(|decoded| {
                decoded.into_iter().map(|node| Fr::from(node).to_repr()).collect()
            })?;

    // Keep file descriptors alive.
    let _ = replica_file.into_raw_fd();
    let _ = sealing_key_file.into_raw_fd();

    Ok(decoded)
}
```

This would expose only the "decode/encode range" functions in `filecoin-proofs` and have the desired functionality of operating on FDs (whose opening/closing and seeking is managed by Lotus).

cryptonemo:
> I think that the requested FD-based FFI is possible given some changes are made to this PR (with the caveat that I am not super familiar working with `filecoin-ffi`).
> 
> What if something like the following was added to `filecoin-ffi` for both pairs (Sealing and SnapDeals) of encoding/decoding functions?
> 
> ```rust
> #[ffi_export]
> unsafe fn decode_replica_range(
>     replica_fd: libc::c_int,
>     sealing_key_fd: libc::c_int,
>     num_nodes: usize,
> ) -> Result<Vec<[u8; 32]>> {
>     // This function assumes that both the replica and key files are open and have been seeked to the desired node offset.
>     let replica_file = File::from_raw_fd(replica_fd);
>     let key_file = File::from_raw_fd(sealing_key_fd);
> 
>     let decoded: Vec<[u8; 32]> =
>         filecoin_proofs::decode_replica_range(replica_file, key_file, num_nodes)
>             .map(|decoded| {
>                 decoded.into_iter().map(|node| Fr::from(node).to_repr()).collect()
>             })?;
> 
>     // Keep file descriptors alive.
>     let _ = replica_file.into_raw_fd();
>     let _ = sealing_key_file.into_raw_fd();
> 
>     Ok(decoded)
> }
> ```
> 
> This would expose only the "decode/encode range" functions in `filecoin-proofs` and have the desired functionality of operating on FDs (whose opening/closing and seeking is managed by Lotus).

:+1: Given that's how we've exposed interfaces like that from FFI in the past, I think it's a good first step because the underlying implementation then has room to wiggle or be re-factored over time as needed.

DrPeterVanNostrand:
@vmx
> Would it be possible to move tests into the files then?

These functions could be tested via unit tests, I used integration tests to keep the convention that (almost all of) `filecoin-proofs`'s testing is done via integration tests and because the required test setup functionality is in the integration tests.

storage-proofs-porep/src/encode.rs
```diff
@@ -20,3 +20,7 @@ pub fn decode<T: Domain>(key: T, value: T) -> T {
     result -= key;
     result.into()
 }
+
+pub fn decode_fr(key: &mut Fr, value: Fr) {
+    *key = value - *key;
```
vmx, line 25: Nitpick (no need to change): For symmetry with `encode_fr` it could be `*key = = value` and it could be called from `decode()` like `encode_fr` is called from `encode`.

vmx (dismissed):
Those APIs are so much better than the last iteration. The replica encode/decode is pretty straight forward. I've been talking with @magik6k about it. So perhaps they are implemented in Go directly, so you wouldn't need to cal out to the FFI.

I'd hope we could also do the SnapDeals one in Go, but that much is more difficult as it involves Poseidon hashing.

I approve, but it should only be merged when @magik6k agrees that those are the APIs he needs.

storage-proofs-porep/src/encode.rs
DrPeterVanNostrand: Yup, makes sense. You mean something like:
```rust
pub fn decode<T: Domain>(key: T, value: T) -> T {
    let value: Fr = value.into();
    let mut key: Fr = key.into();
    decode_fr(&mut key, &value);
    value.into()
}

pub fn decode_fr(key: &mut Fr, value: &Fr) {
    *key = value - key;
}
```

vmx:
This PR was superseded by https://github.com/filecoin-project/rust-fil-proofs/pull/1704. The normal replica encode/decode will be done directly in Go.

PR #1688, opened 2023-03-30
fix: Benchy test without git code cannot output result

lovel8:
Description: When the benchy program is generated on the compilation server and copied to a non-git code server for testing, the report result cannot be output.

### DIFF
--- a/fil-proofs-tooling/Cargo.toml
+++ b/fil-proofs-tooling/Cargo.toml
@@ -26,7 +26,6 @@ bellperson = "0.24.0"
 rand = "0.8"
 tempfile = "3.0.8"
 cpu-time = "1.0.0"
-git2 = "0.14.2"
 heim = { git = "https://github.com/heim-rs/heim", rev = "b292f15", features = ["host", "memory", "cpu"] }
 async-std = "1.6"
 blake2s_simd = "1.0.0"
@@ -48,6 +47,9 @@ structopt = "0.3.12"
 humansize = "1.1.0"
 blstrs = "0.6.0"
 time = "0.3.9"
+
+[build-dependencies]
+vergen = { version = "8.1.1", features = ["build", "git", "gitcl"] }
 
 [features]
 default = ["opencl", "measurements"]
--- a/fil-proofs-tooling/build.rs
+++ b/fil-proofs-tooling/build.rs
@@ -1,0 +1,8 @@
+use std::error::Error;
+use vergen::EmitBuilder;
+
+fn main() -> Result<(), Box<dyn Error>> {
+    // Emits the `VERGEN_GIT_SHA` and `VERGEN_GIT_COMMIT_TIMESTAMP` environment variables.
+    EmitBuilder::builder().all_git().emit()?;
+    Ok(())
+}
--- a/fil-proofs-tooling/src/metadata.rs
+++ b/fil-proofs-tooling/src/metadata.rs
@@ -1,6 +1,5 @@
 use anyhow::{anyhow, Result};
-use chrono::{DateTime, TimeZone, Utc};
-use git2::Repository;
+use chrono::{DateTime, Utc};
 use serde::Serialize;
 
 /// Captures metadata about the current setup.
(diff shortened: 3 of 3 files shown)


vmx (commented):
What about  making the Git metadata optional at https://github.com/filecoin-project/rust-fil-proofs/blob/963eb08e55b9700657d8533418bf6332ac77d197/fil-proofs-tooling/src/metadata.rs#L10.

It could be changed to `Option<GitMetadata>` and if retrieving the data fails, it would just be None.

lovel8:
> What about making the Git metadata optional at
> 
> https://github.com/filecoin-project/rust-fil-proofs/blob/963eb08e55b9700657d8533418bf6332ac77d197/fil-proofs-tooling/src/metadata.rs#L10
> 
> .
> It could be changed to `Option<GitMetadata>` and if retrieving the data fails, it would just be None.

If the git information is empty, the purpose of the code corresponding to this parameter will not be achieved.
Using precompilation to build git information into the program in advance feels like a good choice ( this modify PR)

vmx:
> If the git information is empty, the purpose of the code corresponding to this parameter will not be achieved.
> Using precompilation to build git information into the program in advance feels like a good choice ( this modify PR)

Thanks for the explanation. I only though about the case where you have something like a source tarball without a `.git` directory where you also kind of know the version as you compile it. I haven't thought of compiled binaries.

fil-proofs-tooling/Cargo.toml
```diff
@@ -49,6 +49,9 @@ humansize = "1.1.0"
 blstrs = "0.6.0"
 time = "0.3.9"
 
+[build-dependencies]
+vergen = "3.1.0"
```
vmx, line 53: I only had a quick look. Would it be possible to use the most current version of `vergen`?

lovel8: Sure, Update current last version vergen@V8.1.1

fil-proofs-tooling/src/metadata.rs
```diff
@@ -27,25 +25,14 @@ impl<T> Metadata<T> {
...
-        let head = repo.head()?;
-        let commit = head.peel_to_commit()?;
-        // Unwrap is OK as the given seconds won't be out of range.
-        let date = Utc.timestamp_opt(commit.time().seconds(), 0).unwrap();
-
         Ok(GitMetadata {
-            hash: commit.id().to_string(),
-            date,
+            hash: env!("VERGEN_GIT_SHA").to_string(),
+            date: env!("VERGEN_GIT_COMMIT_TIMESTAMP").to_string(),
```
vmx, line 35: For backwards compatibility, I propose something like:

```rust
/// Captures git specific metadata about the current repo.
#[derive(Debug, Serialize)]
#[serde(rename_all = "kebab-case")]
pub struct GitMetadata {
    hash: String,
    date: DateTime<Utc>,
}

impl GitMetadata {
    pub fn new() -> Result<Self> {
        // Unwrap is OK as vergen returns a valid timestamp.
        let date = env!("VERGEN_GIT_COMMIT_TIMESTAMP").parse::<DateTime<Utc>>().unwrap();
        Ok(GitMetadata {
            hash: env!("VERGEN_GIT_SHA").to_string(),
            date,
        })
    }
}
```

fil-proofs-tooling/build.rs
```diff
@@ -0,0 +1,8 @@
+use std::error::Error;
+use vergen::EmitBuilder;
+
+fn main() -> Result<(), Box<dyn Error>> {
+    // Emit the instructions
```
vmx, line 5: Could you please change this comment to something like

    // Emits the `VERGEN_GIT_SHA` and `VERGEN_GIT_COMMIT_TIMESTAMP` environment variables.

This way if someone searches where those come from, they'll find this comment.

fil-proofs-tooling/Cargo.toml
```diff
@@ -49,6 +49,10 @@ humansize = "1.1.0"
 blstrs = "0.6.0"
 time = "0.3.9"
 
+[build-dependencies]
+vergen = { version = "8.1.1", features = ["build", "git", "gitcl"] }
+time = "^0.3.20"
```
vmx, line 54: I don't think `time` is needed as dependency here..

fil-proofs-tooling/Cargo.toml
vmx, line 1: The `git2` dependency isn't used anymore, so it can be removed from the `Cargo.toml`.

benchy
vmx, line 1: Looks like you accidentally checked in the `benchy` binary.

vmx (commented):
Thanks! Things look good. Once you've removed the `benchy` binary from the commit, it's good to go.

vmx (approved):

PR #1692, opened 2023-04-21
fix: TreeD size and rows to discard calculation was wrong

vmx:
For the TreeD the amount of nodes for the full merkle tree and not for the subtrees needs to be used. The `node_count` variable was shadowed by another instance. This is fixed by renaming the initial variable.


cryptonemo (approved):
Good catch. Fortunately for TreeD it's a no-op.  I wonder if in the future we should plumb `None` down instead, but given how merkletree is structured, makes sense for now.

vmx:
> Fortunately for TreeD it's a no-op.

What is a no-op? The rows to discard?

cryptonemo:
> > Fortunately for TreeD it's a no-op.
> 
> What is a no-op? The rows to discard?

The value passed in (the 'bug') isn't used.

PR #1697, opened 2023-05-08
chore: make tests more deterministic

vmx:
Adding a piece is using random field elements. Instead of using different ones on every test run, use the same random ones on every run. This makes debugging issues easier.


cryptonemo:
Instead of removing the randomness, I'd prefer to see it made optional

vmx:
Deterministic tests are nice for debugging. Making it optional just adds complexity. If someone needs determinism, then a change like this is small enough. Hence I'm closing this PR.

PR #1703, opened 2023-05-15
refactor: introduce new from trait for PoseidonDomain

vmx:
Add a new `From` conversion into the field element repr representation. This makes it also possible to hide the internal representation.

### DIFF
--- a/storage-proofs-porep/src/stacked/vanilla/create_label/multi.rs
+++ b/storage-proofs-porep/src/stacked/vanilla/create_label/multi.rs
@@ -754,6 +754,9 @@ mod tests {
             .read_at(final_labels.len() - 1)
             .expect("read_at");
         dbg!(&last_label);
-        assert_eq!(expected_last_label.to_repr(), last_label.0);
+        assert_eq!(
+            expected_last_label.to_repr(),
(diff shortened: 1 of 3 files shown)


storage-proofs-porep/src/stacked/vanilla/create_label/multi.rs
```diff
@@ -754,6 +754,9 @@ mod tests {
             .read_at(final_labels.len() - 1)
             .expect("read_at");
         dbg!(&last_label);
-        assert_eq!(expected_last_label.to_repr(), last_label.0);
+        assert_eq!(
+            expected_last_label.to_repr(),
+            <<Fr as PrimeField>::Repr>::from(last_label)
```
DrPeterVanNostrand, line 759: It's just an aesthetic preference, but these should also work using the current `Domain` trait.
```rust
assert_eq!(expected_last_label.to_repr(), last_label.as_ref())
assert_eq!(expected_last_label.to_repr(), last_label.into_bytes()
```

DrPeterVanNostrand (dismissed):

DrPeterVanNostrand (approved):

PR #1704, opened 2023-05-17
feat: add function to decode a range from an updated sector

vmx:
Currently it's only possible to decode a full empty updated sector. With this new function, it's possible to also decode only a range.

You can pass in any readable stream, i.e. file descriptors but also pipes or even slices.

BREAKING CHANGE: `EmptySectorUpdate::decode_from` no longer takes `h` as argument.

---

This is based on @DrPeterVanNostrand work at https://github.com/filecoin-project/rust-fil-proofs/pull/1687.

In case you wonder why the `rhos()` function was refactored. This makes it easier callable, as it now only needs the number of nodes as input. The new `Rhos` will then also be used for the upcoming binaries.

### DIFF
--- a/filecoin-proofs/src/api/update.rs
+++ b/filecoin-proofs/src/api/update.rs
@@ -22,7 +22,7 @@ use storage_proofs_core::{
 };
 use storage_proofs_porep::stacked::{PersistentAux, TemporaryAux};
 use storage_proofs_update::{
-    constants::{TreeDArity, TreeDDomain, TreeRDomain, TreeRHasher},
+    constants::{h_default, TreeDArity, TreeDDomain, TreeRDomain, TreeRHasher},
     phi,
     vanilla::Rhos,
     EmptySectorUpdate, EmptySectorUpdateCompound, PartitionProof, PrivateInputs, PublicInputs,
@@ -174,6 +174,7 @@ pub fn encode_into<Tree: 'static + MerkleTreeTrait<Hasher = TreeRHasher>>(
             sector_key_path,
             sector_key_cache_path,
             staged_data_path,
+            h_default(config.nodes_count),
         )?;
 
     let mut comm_d = [0; 32];
@@ -233,7 +234,8 @@ pub fn decode_from_range<R: Read, S: Read, W: Write>(
     let comm_d_domain = TreeDDomain::try_from_bytes(&comm_d[..])?;
     let comm_r_domain = TreeRDomain::try_from_bytes(&comm_r[..])?;
     let phi = phi(&comm_d_domain, &comm_r_domain);
-    let rho_invs = Rhos::new_inv_range(&phi, nodes_count, nodes_offset, num_nodes);
+    let h = h_default(nodes_count);
+    let rho_invs = Rhos::new_inv_range(&phi, h, nodes_count, nodes_offset, num_nodes);
 
     let bytes_length = num_nodes * NODE_SIZE;
 
@@ -311,6 +313,7 @@ pub fn decode_from<Tree: 'static + MerkleTreeTrait<Hasher = TreeRHasher>>(
         <Tree::Hasher as Hasher>::Domain::try_from_bytes(&p_aux.comm_c.into_bytes())?,
         comm_d_new.into(),
         <Tree::Hasher as Hasher>::Domain::try_from_bytes(&p_aux.comm_r_last.into_bytes())?,
+        h_default(config.nodes_count),
     )?;
 
     info!("decode_from:finish");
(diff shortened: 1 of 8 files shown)


storage-proofs-core/src/util.rs
```diff
@@ -177,6 +184,46 @@ pub fn default_rows_to_discard(leafs: usize, arity: usize) -> usize {
     }
 }
 
+// Based on
+// https://stackoverflow.com/questions/73145503/iterator-for-reading-file-chunks/73145594#73145594
+/// Chunks the given reader to the given size.
+///
+/// If the end is reached and there are a few bytes left, that don't fill a full chunk, those bytes
+/// are returned.
+pub struct ChunkIterator<R> {
```
DrPeterVanNostrand, line 193: Will this be used exclusively in `filecoin-proofs`? If so, it may make sense for it to be defined there, rather than in `storage-proofs-core`.

storage-proofs-update/src/vanilla.rs
```diff
@@ -252,6 +254,88 @@ where
     }
 }
 
+// Computes all `2^h` rho values for the given `phi` in the given range. Each rho corresponds to
+// a `high` value, where `high` is the `h` high bits of a node-index.
+#[derive(Debug)]
+pub struct Rhos {
+    rhos: HashMap<usize, Fr>,
```
DrPeterVanNostrand, line 261: No `Vec`? I believe that this hashmap's keys are the integers `0..2^h`, so I think that this hashmap is just adding a layer of abstraction on top of indexing into a vector, i.e. `rhos_vec[high_bits]` is replaced with `rhos_hashmap[high_bits]`?

Unless `Rhos` will be used slightly differently than `fn rhos` is used now (where all `2^h` rhos are generated at once in a vector and the vector elements are referenced throughout vanilla encoding).

storage-proofs-update/src/vanilla.rs
```diff
@@ -252,6 +254,88 @@ where
...
+    pub fn new_inv(phi: &TreeRDomain, nodes_count: usize) -> Self {
+        Self::new_inv_range(phi, nodes_count, 0, nodes_count)
+    }
+
+    /// Generate the `rho`s for a certain number of nodes and range.
+    ///
+    /// Those are used for encoding.
+    ///
+    /// All inputs are in number of nodes.
+    pub fn new_range(phi: &TreeRDomain, nodes_count: usize, offset: usize, num: usize) -> Self {
```
DrPeterVanNostrand, line 286: The parameters names `nodes_count`, `offset` and `num` in these two methods are a little unclear imo.

DrPeterVanNostrand (changes requested):

storage-proofs-update/src/vanilla.rs
DrPeterVanNostrand: Nvm, I see below that it is not guaranteed that the keys are `0..2^h`.

filecoin-proofs/src/api/update.rs
```diff
@@ -203,6 +209,84 @@ pub fn encode_into<Tree: 'static + MerkleTreeTrait<Hasher = TreeRHasher>>(
...
+    {
+        let chunk_offset = chunk_index * chunk_size;
+
+        // The end of the intended decoding range was reached.
+        if chunk_offset > bytes_length {
+            break;
+        }
+
+        let input_chunk = input_chunk_result?;
+        let sector_key_chunk = sector_key_chunk_result?;
```
DrPeterVanNostrand, line 254: I think adding `context` to these read results (at least "input_data/sector_key read failure") would be a useful for determining where exactly an io error occurred.

DrPeterVanNostrand (changes requested):

storage-proofs-core/src/util.rs
vmx: Very good point. I've moved it to `filecoin-proofs`.

cryptonemo (approved):
The decode parts (and testing) looks good to me.  I'll approve, but let Jake confirm the change regarding the Rhos.  Thanks!

storage-proofs-update/src/vanilla.rs
```diff
@@ -252,6 +254,90 @@ where
...
+    rhos: HashMap<usize, Fr>,
+    /// The amount of right shift that is needed to get the `h` high bits of a node.
+    bits_shr: usize,
+}
+
+impl Rhos {
+    /// Generate the `rho`s for a certain number of nodes.
+    ///
+    /// Those are used for encoding. The `nodes_count` is the total number of nodes of the sector.
+    pub fn new(phi: &TreeRDomain, nodes_count: usize) -> Self {
```
DrPeterVanNostrand, line 270: Can these constructors accept an `h` parameter, which gets passed to `Self::calc_bits_shr` rather than forcing `h = h_default(sector_nodes)` (and call `h_default` at the api level and pass it to the constructors)?

PR #1705, opened 2023-05-17
Add APIs to generate TreeC and TreeR separately

vmx:
This adds APIs for TreeC and TreeRLast. You can also review the commits individually.


Cargo.toml
```diff
@@ -12,3 +12,6 @@ members = [
   "filecoin-hashers",
   "storage-proofs-update",
 ]
+
+[patch.crates-io]
+merkletree = { git = "https://github.com/filecoin-project/merkletree" }
```
DrPeterVanNostrand, line 17: Maybe there should be a `TODO` here indicating that this change is temporary.

filecoin-proofs/src/api/seal.rs
```diff
@@ -1130,3 +1135,118 @@ pub fn verify_batch_seal<Tree: 'static + MerkleTreeTrait>(
...
+    replica_path: R,
+    output_dir: O,
+) -> Result<<TreeR::Hasher as Hasher>::Domain>
+where
+    O: AsRef<Path>,
+    R: AsRef<Path>,
+{
+    let leaf_count = sector_size as usize / NODE_SIZE;
+    let tree_count = get_base_tree_count::<TreeR>();
+    let nodes_count = leaf_count / tree_count;
```
DrPeterVanNostrand, line 1154: It may be a good idea to name this `base_tree_leafs` or something similar (I know that the name `nodes_count` is used in `StackedDrg::generate_tree_...`, but that naming is bad imo).

Isn't this value the same as `base_tree_leafs` below?

DrPeterVanNostrand: This comment also applies to the below function.

filecoin-proofs/tests/api.rs
```diff
@@ -1833,6 +1848,23 @@ fn compare_elements(path1: &Path, path2: &Path) -> Result<(), Error> {
     Ok(())
 }
 
+/// Compare the elements of trees in two different directories. If the tree is split into multiple
+/// files, only the first file is checked.
+fn compare_elements_with_cache_key<Tree: 'static + MerkleTreeTrait>(
```
DrPeterVanNostrand, line 1853: The purpose of this function is just to check equivalency of the first base trees in `dir_a` and `dib_b`? Is that an intentional constraint (maybe for performance reasons) or does it make sense for the function to check full tree equivalency?

vmx: For performance reasons I thought it's good enough to check the first file. If the result is different, I'd expect it to be wildly different, hence already on the first file.

cryptonemo: I'm agreeing with Jake here, for consistency, it makes sense to check the entire tree.  Test sector sizes are very quick to check, so it can be avoided at prod sector sizes in tests where possible (although those should be on the order of a few minutes in any case)

vmx: Would it make sense to introduce file hashing instead? Given that's really about "the files are equal", that they are correct field elements was already checked.

vmx: I've added hashing the files.

filecoin-proofs/src/api/seal.rs
vmx: > Isn't this value the same as `base_tree_leafs` below?

No it's the binary vs. the octa tree.

DrPeterVanNostrand: Oh, right. Thanks

filecoin-proofs/tests/api.rs
DrPeterVanNostrand: Yup, works for me.

DrPeterVanNostrand (approved):

PR #1706, opened 2023-05-17
Add API to do only SDR

vmx:
This PR adds an API to do SDR without doing anything else.

Please note that it's split into two commits. The first one adds the API, the second one is a refacoring, which makes the API a bit clearer IMHO. It passes only on the path, which is the only thing that's really needed.

### DIFF
--- a/filecoin-proofs/src/api/seal.rs
+++ b/filecoin-proofs/src/api/seal.rs
@@ -1251,3 +1251,33 @@ where
 
     Ok(tree_c.root())
 }
+
+pub fn sdr<P, Tree: 'static + MerkleTreeTrait>(
+    porep_config: &PoRepConfig,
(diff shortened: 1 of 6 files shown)


cryptonemo (dismissed):
Looks good, but just noting that the reason the StoreConfig was passed around instead of the path is because it was meant to contain all aspects of the Store, including the path.

vmx:
> it was meant to contain all aspects of the Store, including the path.

Noted.  I found it confusing that all aspects, except for the `rows_to_discard`, are overridden later on.

vmx:
@cryptonemo I've rebased that one. I've squased the addition after your last review into the first commit. This was just a input argument order change nothing else.

cryptonemo (approved):

PR #1707, opened 2023-05-24
fix: make add_piece taking less time

vmx:
`add_piece` operates in 64 bytes pieces. It's faster to operate in bigger chunks. This commits increases the buffer size to 4KiB. This makes adding a 32GiB piece about more than 2x faster. On the hardware I used it goes down from 14min to about 6min.

### DIFF
--- a/filecoin-proofs/src/commitment_reader.rs
+++ b/filecoin-proofs/src/commitment_reader.rs
@@ -7,11 +7,13 @@ use rayon::prelude::{ParallelIterator, ParallelSlice};
 
 use crate::{constants::DefaultPieceHasher, pieces::piece_hash};
 
+const BUFFER_SIZE: usize = 4096;
+
 /// Calculates comm-d of the data piped through to it.
 /// Data must be bit padded and power of 2 bytes.
 pub struct CommitmentReader<R> {
     source: R,
-    buffer: [u8; 64],
+    buffer: Vec<u8>,
     buffer_pos: usize,
     current_tree: Vec<<DefaultPieceHasher as Hasher>::Domain>,
 }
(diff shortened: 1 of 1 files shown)


filecoin-proofs/src/commitment_reader.rs
```diff
@@ -7,11 +7,13 @@ use rayon::prelude::{ParallelIterator, ParallelSlice};
 
 use crate::{constants::DefaultPieceHasher, pieces::piece_hash};
 
+const BUFFER_SIZE: usize = 4096;
```
Kubuxu, line 10: Have you tried bumping it even further, for example, up to 1MiB-8MiB?
Even with the 4KiB, it is just 64 SHA invocations, the branch predictor is probably just warming up at that point.

vmx: I've tried 8MiB it didn't make a difference.

I think yet another difference would make it if the hashing would be done in parallel. But a simple "par chunker" didn't really help, as then the pieces are to small. One need to implement it manually. I decided that it's not worth it and that this improvement is already good enough for how simple it is.

cryptonemo:
Can we check if lotus is using this method?  I recall it not being used, but may be wrong there.  In any case, it should help speed up our big tests.

cryptonemo:
> > Can we check if lotus is using this method?
> 
> `write_with_alignment` seems to be the only API call in the FFI using it. It is then called `WriteWithAlignment` on the Go side. If I grep the Lotus repo for that, I only find it used in tests.
> 
> Why is that important? It's not a breaking change, it just improves things. So if Lotus would use it, it would also be beneficial for them.

I just wanted to know if we'd see the speed-ups outside of tests.

cryptonemo (approved):

PR #1708, opened 2023-05-25
refactor: do not call circuit_proofs directly

vmx:
`circuit_proofs` should not be called directly according to the comment at
 https://github.com/filecoin-project/rust-fil-proofs/blob/95e8c982540b14077d58ee9ddc076f7602b4805d/storage-proofs-core/src/compound_proof.rs#L228-L232

The correct call is `prove_with_vanilla`, use that instead.

This leads to a clearer call path, hence an easier to follow code.

### DIFF
--- a/filecoin-proofs/src/api/seal.rs
+++ b/filecoin-proofs/src/api/seal.rs
@@ -524,17 +524,12 @@ pub fn seal_commit_phase2<Tree: 'static + MerkleTreeTrait>(
         _,
     >>::setup(&compound_setup_params)?;
 
-    trace!("snark_proof:start");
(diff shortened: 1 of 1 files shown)


cryptonemo:
I don't know that that's correct.  The comment is indicating that it shouldn't be overridden, not wired internally.

vmx:
Although I still find this cleaner, it funnily enough gets in my way with integrating the supraseal-c2. So @cryptonemo, if you don't like this change, we can also close it.

vmx:
I'm closing this one for now, as I currently exploit the fact of calling `circuit_proofs` directly in another branch.

PR #1713, opened 2023-07-08
Remove unnecessary copy of proof bytes

lemmih:
`new_from_reader` unnecessarily copies the bytes into a `Vec<u8>` in this case.


cryptonemo (approved):
Good catch, thanks!

PR #1714, opened 2023-08-08
fix: downgrade harmless warning to trace logging

cryptonemo:


cryptonemo:
Related to https://github.com/filecoin-project/lotus/issues/11143

PR #1715, opened 2023-09-01
feat: add fallback option if no t_aux file exists

vmx:
The only information from the `t_aux` file that is used during Commitphase 1 (C1) is the number of rows that are discarded from TreeC. If that value is either the default, or set via `FIL_PROOFS_ROWS_TO_DISCARD` accordingly, then all other parts of `t_aux` can be inferred from the function call.

When `FIL_PROOFS_T_AUX_IMPLICIT_FALLBACK=1` is set, then those inferred values will be used, in case the `t_aux` file cannot be read or doesn't exist.

### DIFF
--- a/filecoin-proofs/src/api/seal.rs
+++ b/filecoin-proofs/src/api/seal.rs
@@ -26,6 +26,7 @@ use storage_proofs_core::{
     parameter_cache::SRS_MAX_PROOFS_TO_AGGREGATE,
     proof::ProofScheme,
     sector::SectorId,
+    settings::SETTINGS,
     util::{default_rows_to_discard, NODE_SIZE},
     Data,
 };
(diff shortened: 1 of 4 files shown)


storage-proofs-porep/src/stacked/vanilla/params.rs
```diff
@@ -838,6 +839,60 @@ impl<Tree: MerkleTreeTrait, G: Hasher> Clone for TemporaryAux<Tree, G> {
...
+        };
+
+        let tree_size = get_merkle_tree_len(sector_nodes, Tree::Arity::to_usize())
+            .expect("Tree must have enough leaves and have an arity of power of two");
+
+        let tree_r_last_config = StoreConfig {
+            path: cache_path.clone(),
+            id: CacheKey::CommRLastTree.to_string(),
+            size: Some(tree_size),
+            // A default 'rows_to_discard' value will be chosen for tree_r_last, unless the user
```
cryptonemo, line 870: I think it's possible that this part/assumption may break things.  It's reasonable, but may require some more thought/testing across those cases.

filecoin-proofs/src/api/seal.rs
```diff
@@ -444,16 +445,35 @@ pub fn seal_commit_phase1_inner<T: AsRef<Path>, Tree: 'static + MerkleTreeTrait>
...
         let t_aux_path = cache_path.as_ref().join(CacheKey::TAux.to_string());
-        let t_aux_bytes = fs::read(&t_aux_path)
-            .with_context(|| format!("could not read file t_aux={:?}", t_aux_path))?;
-
-        let mut res: TemporaryAux<_, _> = deserialize(&t_aux_bytes)?;
-
-        // Switch t_aux to the passed in cache_path
-        res.set_cache_path(cache_path);
-        res
+        match fs::read(&t_aux_path) {
```
cryptonemo, line 452: As-is, this won't work for Sector Updates (and maybe other parts?).  I don't think that path will be much more complicated, but does require more testing, as `t_aux` is very much not optional in at least that part of the code.

cryptonemo (changes requested):

storage-proofs-porep/src/stacked/vanilla/params.rs
vmx: Why would it break things? This is copy and pasted from the original code. The only difference is that we take the default value/the env variable and don't read it from disk.

Is this a comment or blocking a merge? If it's blocking, what do you suggest?

filecoin-proofs/src/api/seal.rs
vmx: Isn't the Sector Updates part self-contained at https://github.com/filecoin-project/rust-fil-proofs/blob/2ec0a0fe4005a6f2f6659bbbc167a1d41ee855e0/filecoin-proofs/src/api/update.rs? This part should only concern the C1.

So would be adding a test that removes the `t_aux` and then runs C1 be sufficient?

vmx:
I thought about this again. Would it help that we make it clear in the README that you're on your own if you set the `FIL_PROOFS_T_AUX_IMPLICIT_FALLBACK` env variable? When you use the code in the current stack and there is a `t_aux`, nothing changes, it will also error if there is none.

If you don't have a `t_aux` and set `FIL_PROOFS_T_AUX_IMPLICIT_FALLBACK`, then it uses the default value, which is either the hard-coded one, or the one you can set dynamically via `FIL_PROOFS_ROWS_TO_DISCARD`. So even in case someone has set `FIL_PROOFS_ROWS_TO_DISCARD` to a custom value, I'd expect it to be set all the time. This means that even if then `t_aux` would be deleted `FIL_PROOFS_T_AUX_IMPLICIT_FALLBACK` would be set, it would still return the correct result.

filecoin-proofs/src/api/seal.rs
cryptonemo: The sector updates code currently relies on t_aux being available and if it's not available from an earlier step in sealing, future sector updates won't work.

vmx: Sorry, now I get it. The problem is **not** about the code paths. Each works fine independently. The problem is that if someone seals a CC sector and removes their `t_aux`, they could then still C1, but they later could **not** do a sector update.

This means that I also need to take care of the sector update code path. Would adding a similar fallback there be an OK solution?

vmx:
Superseded by https://github.com/filecoin-project/rust-fil-proofs/pull/1717.

PR #1717, opened 2023-09-14
Allow sealing without t_aux files

vmx:
This PR introduces a new feature called `fixed-rows-to-discard`, when enabled, no `t_aux` files are written or read. It contains quite a lot of changes as some refactorings lead to less `StoreConfig`s and hence make it easier to reason about the whole change that it still does the correct thing.

This PR supersedes https://github.com/filecoin-project/rust-fil-proofs/pull/1715.

### DIFF
--- a/filecoin-proofs/src/api/mod.rs
+++ b/filecoin-proofs/src/api/mod.rs
@@ -739,8 +739,7 @@ where
     // Make sure p_aux exists and is valid.
     let _ = util::get_p_aux::<Tree>(cache)?;
 
-    // Make sure t_aux exists and is valid.
-    let t_aux = util::get_t_aux::<Tree>(cache)?;
+    let t_aux = util::get_t_aux::<Tree>(cache, metadata.len())?;
 
     // Verify all stores/labels within the Labels object.
     let cache = cache_path.as_ref().to_path_buf();
--- a/filecoin-proofs/src/api/seal.rs
+++ b/filecoin-proofs/src/api/seal.rs
@@ -286,6 +286,8 @@ where
         _,
     >>::setup(&compound_setup_params)?;
 
+    // Silence Clippy warning for the case where `t_aux` is not written.
+    #[allow(unused_variables)]
     let (tau, (p_aux, t_aux)) = StackedDrg::<Tree, DefaultPieceHasher>::replicate_phase2(
         &compound_public_params.vanilla_params,
         labels,
@@ -299,7 +301,8 @@ where
 
     // Persist p_aux and t_aux here
     util::persist_p_aux::<Tree>(&p_aux, cache_path.as_ref())?;
-    util::persist_t_aux::<Tree>(&t_aux, cache_path.as_ref())?;
+    #[cfg(not(feature = "fixed-rows-to-discard"))]
+    util::persist_t_aux(&t_aux, cache_path.as_ref())?;
 
     let out = SealPreCommitOutput { comm_r, comm_d };
 
@@ -414,7 +417,7 @@ pub fn seal_commit_phase1_inner<T: AsRef<Path>, Tree: 'static + MerkleTreeTrait>
     );
 
     let p_aux = util::get_p_aux::<Tree>(cache_path.as_ref())?;
-    let t_aux = util::get_t_aux::<Tree>(cache_path.as_ref())?;
+    let t_aux = util::get_t_aux::<Tree>(cache_path.as_ref(), u64::from(porep_config.sector_size))?;
 
     // Convert TemporaryAux to TemporaryAuxCache, which instantiates all
     // elements based on the configs stored in TemporaryAux.
@@ -1167,12 +1170,13 @@ where
         path: PathBuf::from(output_dir.as_ref()),
         id: CacheKey::CommRLastTree.to_string(),
         size: Some(size),
-        // A default 'rows_to_discard' value will be chosen for tree_r_last, unless the user
-        // overrides this value via the environment setting (FIL_PROOFS_ROWS_TO_DISCARD). If
-        // this value is specified, no checking is done on it and it may result in a broken
-        // configuration. *Use with caution*. It must be noted that if/when this unchecked
-        // value is passed through merkle_light, merkle_light now does a check that does not
-        // allow us to discard more rows than is possible to discard.
+        // A default 'rows_to_discard' value will be chosen for tree_r_last, unless the
+        // `fixed-rows-to-discard` feature is not enabled and the user overrides this value via
+        // the environment setting (FIL_PROOFS_ROWS_TO_DISCARD). If this value is specified, no
+        // checking is done on it and it may result in a broken configuration. *Use with caution*.
+        // It must be noted that if/when this unchecked value is passed through merkle_light,
+        // merkle_light now does a check that does not allow us to discard more rows than is
+        // possible to discard.
         rows_to_discard: default_rows_to_discard(base_tree_leafs, TreeR::Arity::to_usize()),
     };
 
--- a/filecoin-proofs/src/api/update.rs
+++ b/filecoin-proofs/src/api/update.rs
@@ -95,7 +95,8 @@ pub fn encode_into<Tree: 'static + MerkleTreeTrait<Hasher = TreeRHasher>>(
     let config = SectorUpdateConfig::from_porep_config(porep_config);
 
     let p_aux = util::get_p_aux::<Tree>(sector_key_cache_path)?;
-    let t_aux = util::get_t_aux::<Tree>(sector_key_cache_path)?;
+    let t_aux =
+        util::get_t_aux::<Tree>(sector_key_cache_path, u64::from(porep_config.sector_size))?;
 
     let (tree_d_new_config, tree_r_last_new_config) =
         get_new_configs_from_t_aux_old::<Tree>(&t_aux, new_cache_path, config.nodes_count)?;
@@ -138,6 +139,7 @@ pub fn encode_into<Tree: 'static + MerkleTreeTrait<Hasher = TreeRHasher>>(
     let mut p_aux = p_aux;
     p_aux.comm_r_last = comm_r_last_domain;
     util::persist_p_aux::<Tree>(&p_aux, new_cache_path)?;
+    #[cfg(not(feature = "fixed-rows-to-discard"))]
     util::persist_t_aux::<Tree>(&t_aux, new_cache_path)?;
 
     info!("encode_into:finish");
@@ -272,7 +274,7 @@ pub fn remove_encoded_data<Tree: 'static + MerkleTreeTrait<Hasher = TreeRHasher>
     info!("remove_data:start");
 
     let p_aux = util::get_p_aux::<Tree>(replica_cache_path)?;
-    let t_aux = util::get_t_aux::<Tree>(replica_cache_path)?;
+    let t_aux = util::get_t_aux::<Tree>(replica_cache_path, u64::from(config.sector_size))?;
 
     let (_, tree_r_last_new_config) =
         get_new_configs_from_t_aux_old::<Tree>(&t_aux, sector_key_cache_path, config.nodes_count)?;
@@ -295,6 +297,7 @@ pub fn remove_encoded_data<Tree: 'static + MerkleTreeTrait<Hasher = TreeRHasher>
     let mut p_aux = p_aux;
     p_aux.comm_r_last = tree_r_last_new;
     util::persist_p_aux::<Tree>(&p_aux, sector_key_cache_path)?;
+    #[cfg(not(feature = "fixed-rows-to-discard"))]
     util::persist_t_aux::<Tree>(&t_aux, sector_key_cache_path)?;
 
     info!("remove_data:finish");
@@ -337,7 +340,7 @@ pub fn generate_single_partition_proof<Tree: 'static + MerkleTreeTrait<Hasher = 
         h: config.h,
     };
 
-    let t_aux_old = util::get_t_aux::<Tree>(sector_key_cache_path)?;
+    let t_aux_old = util::get_t_aux::<Tree>(sector_key_cache_path, u64::from(config.sector_size))?;
 
     let (tree_d_new_config, tree_r_last_new_config) =
         get_new_configs_from_t_aux_old::<Tree>(&t_aux_old, replica_cache_path, config.nodes_count)?;
@@ -429,7 +432,7 @@ pub fn generate_partition_proofs<Tree: 'static + MerkleTreeTrait<Hasher = TreeRH
         h: config.h,
     };
 
-    let t_aux_old = util::get_t_aux::<Tree>(sector_key_cache_path)?;
+    let t_aux_old = util::get_t_aux::<Tree>(sector_key_cache_path, u64::from(config.sector_size))?;
 
     let (tree_d_new_config, tree_r_last_new_config) =
         get_new_configs_from_t_aux_old::<Tree>(&t_aux_old, replica_cache_path, config.nodes_count)?;
@@ -571,7 +574,7 @@ pub fn generate_empty_sector_update_proof<Tree: 'static + MerkleTreeTrait<Hasher
         h: config.h,
     };
 
-    let t_aux_old = util::get_t_aux::<Tree>(sector_key_cache_path)?;
+    let t_aux_old = util::get_t_aux::<Tree>(sector_key_cache_path, u64::from(config.sector_size))?;
 
     let (tree_d_new_config, tree_r_last_new_config) =
         get_new_configs_from_t_aux_old::<Tree>(&t_aux_old, replica_cache_path, config.nodes_count)?;
--- a/filecoin-proofs/src/api/util.rs
+++ b/filecoin-proofs/src/api/util.rs
@@ -83,9 +83,7 @@ pub(crate) fn get_p_aux<Tree: MerkleTreeTrait>(
     Ok(p_aux)
 }
 
-/// Instantiates t_aux from the specified cache_dir for access to  labels and tree_d, tree_c,
-/// tree_r_last store configs.
-pub(crate) fn get_t_aux<Tree: MerkleTreeTrait>(
+fn read_t_aux_file<Tree: MerkleTreeTrait>(
     cache_path: &Path,
 ) -> Result<TemporaryAux<Tree, DefaultPieceHasher>> {
     let t_aux_path = cache_path.join(CacheKey::TAux.to_string());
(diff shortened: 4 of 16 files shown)


filecoin-proofs/src/api/mod.rs
```diff
@@ -58,80 +53,40 @@ pub use winning_post::*;
 pub use storage_proofs_update::constants::partition_count;
 
 // Ensure that any associated cached data persisted is discarded.
-pub fn clear_cache<Tree: MerkleTreeTrait>(cache_dir: &Path) -> Result<()> {
+pub fn clear_cache(cache_dir: &Path) -> Result<()> {
```
cryptonemo, line 56: Note that `clear_cache`, `clear_layer_data`, and `clear_synthetic_proofs` are exposed API methods, not just internal.  In other words, dropping (or potentially changing) them will cause issues to users of them.

vmx: Yes, I know. `clear_cache` and `clear_layer_data` did the same thing, hence one was dropped (of course it need to be reflected in proofs API as well as FFI. The way they work is indeed slightly different. Two things that are different:

 - Trees which consist of several files are now deleted by a glob. So in case someone would have a file that is names similarly to a tree file, but is not really part of it, that file would be deleted. Does that really happen?
 - The trees were opened. If I understand it correctly, then it could have happened that if a file was corrupt that it wasn't deleted. Now it's just deleted. I think that's a win.

I would argue that such a massive simplification of the code, and also for people to explain "xyz" will happen justifies that change.

filecoin-proofs/src/api/mod.rs
```diff
@@ -58,80 +53,40 @@ pub use winning_post::*;
...
-    let result = TemporaryAux::<Tree, DefaultPieceHasher>::clear_synthetic_proofs(&t_aux);
-
+    let result = stacked::clear_synthetic_proofs(cache_dir);
     info!("clear_synthetic_proofs:finish");
-
     result
 }
 
+/// Generate the replica id as expected for Stacked DRG.
+pub fn generate_replica_id<H: Hasher, T: AsRef<[u8]>>(
```
DrPeterVanNostrand, line 72: Was this moved out of `storage-proofs-porep::stacked` just to reduce the API surface of that crate?

vmx: I've moved that as the `stacked/vanilla/params.rs` looked like a "catch all" for all sorts of (unrelated) things. So I took the chance to move things around, while I was working on that file.

filecoin-proofs/src/api/mod.rs
cryptonemo: The API signatures are different, so I'd prefer we not disturb them for now.  If they weren't exposed, it's an easy internal re-factor.  They didn't do the same thing originally, but ended up doing the same btw.  They may not again at some point later.

I don't think glob delete is a good idea because we don't 'own' the directory we're using, necessarily.  If it was entirely created and managed internally, it could be different.  I see it as our API is given access to a dir in which we persist and use the things we require and leave whatever else may be there up to others.

vmx: > The API signatures are different, so I'd prefer we not disturb them for now. If they weren't exposed, it's an easy internal re-factor. They didn't do the same thing originally, but ended up doing the same btw. They may not again at some point later.
> 
> I don't think glob delete is a good idea because we don't 'own' the directory we're using, necessarily. If it was entirely created and managed internally, it could be different. I see it as our API is given access to a dir in which we persist and use the things we require and leave whatever else may be there up to others.

So it seems like the difference was needed => we should take the chance to make a change and unify it into a single function all the way up. It it turns out something else is needed in the future, it can be added again.



> The API signatures are different, so I'd prefer we not disturb them for now. If they weren't exposed, it's an easy internal re-factor. They didn't do the same thing originally, but ended up doing the same btw. They may not again at some point later.

 
> I don't think glob delete is a good idea because we don't 'own' the directory we're using, necessarily. If it was entirely created and managed internally, it could be different. I see it as our API is given access to a dir in which we persist and use the things we require and leave whatever else may be there up to others.

I would solve this by communicating in a change long that files with such and such prefix are going to be removed.

The problem if we don't use globs is, that we either need a new API for the "no t_aux" case or we have different behaviour for the "no t_aux" case, which would be painful to maintain.

I totally agree that we should not break things where it isn't needed. But IMHO that's really more of theoretic breakage then practical one. It's not like we delete all files. We delete files with a specific prefix and extension.

Would an option be to discuss this with the Lotus team and see if they would be OK with that change?

vmx: > The API signatures are different, so I'd prefer we not disturb them for now. If they weren't exposed, it's an easy internal re-factor. They didn't do the same thing originally, but ended up doing the same btw. They may not again at some point later.
> 
> I don't think glob delete is a good idea because we don't 'own' the directory we're using, necessarily. If it was entirely created and managed internally, it could be different. I see it as our API is given access to a dir in which we persist and use the things we require and leave whatever else may be there up to others.

I had another look. The `clear_caches()` function is not used at all. Not in `rust-fil-proofs` and not in `rust-filecoin-proofs-api`. So I'd really be in favour of removing it. let's take the chance and reducing the API surface in cases it isn't even used. I can also open a separate PR if you like (which I'd actually prefer).

`clear_layer_data()` and `clear_cache()` have the same signature and do the same thing. The `clear_layer_data()` is exposed through `rust-filecoin-proofs-api`, so yes, it's a breaking change, but we can easily just call `clear_cache()` directly from `rust-filecoin-proofs-api` or (which I'd prefer and happy to do a PR` remove it there. I really prefer if we don't drag on cruft, which isn't really needed. We should strive for a minimal API surface where possible.

vmx:
As requested on a side-channel, let's start with a PR with the refactorings first. That can be found at https://github.com/filecoin-project/rust-fil-proofs/pull/1719.

filecoin-proofs/src/api/util.rs
```diff
@@ -83,8 +83,45 @@ pub(crate) fn get_p_aux<Tree: MerkleTreeTrait>(
...
+    sector_bytes: u64,
+) -> Result<TemporaryAux<Tree, DefaultPieceHasher>> {
+    trace!(
+        "Instantiating TemporaryAux from default values with cache_path at {:?}",
+        cache_path
+    );
+    let t_aux_path =
+        merkletree::store::StoreConfig::data_path(cache_path, &CacheKey::TAux.to_string());
+    if Path::new(&t_aux_path).exists() {
+        log::warn!(
```
DrPeterVanNostrand, line 102: Everything looks good to me. I'm just wondering if this should fallback to `bincode::deserialize()` rather than a warning, or at least warn/fail if the taux on disk disagrees with the default taux values? But maybe that breaks something that I'm unaware of, and there's a good reason why this is just a warning.

filecoin-proofs/src/api/update.rs
```diff
@@ -272,6 +277,9 @@ pub fn remove_encoded_data<Tree: 'static + MerkleTreeTrait<Hasher = TreeRHasher>
     info!("remove_data:start");
 
     let p_aux = util::get_p_aux::<Tree>(replica_cache_path)?;
+    #[cfg(feature = "fixed-rows-to-discard")]
```
cryptonemo, line 280: In order to clean-up all of the switching around this, why not just update the `get_t_aux` method to take the sector_bytes and ignore it inside if not needed (using the switching)?

filecoin-proofs/src/api/util.rs
vmx: The idea is to fail early. If there is a `t_aux` file, then something with your setup/environment is wrong and you should really fix it. An implicit fallback could lead to hard to debug errors.

filecoin-proofs/src/api/update.rs
vmx: Good idea!

filecoin-proofs/src/api/util.rs
cryptonemo: If someone uses this code to do a sector update on a previously sealed sector, why fail if `t_aux` exists?

filecoin-proofs/src/api/mod.rs
cryptonemo: Not following the latest here, but `clear_cache` is exposed via FFI as well (saying it's not used in an `api` test doesn't mean it's not used)

filecoin-proofs/src/api/util.rs
vmx: We don't fail, we just warn that it is ignored. Of course things then might fail further down the line, hence the warning.

When we talked about it, I though we wanted a clear separating between "you run it with t_aux" and "you run it without t_aux". The "no t_aux" code path, never ever touches that file. In case there is a t_aux file, I would expect that it's more often accidental, rather then intentional.

But if we want to blur that distinction, I can make it read in the t_aux file. Though I'd prefer not doing that.

filecoin-proofs/src/api/mod.rs
vmx: Looks like I searched accidentally for `clear_caches` (with the "s"). Yes you're right, it's part of the FFI.

filecoin-proofs/src/api/util.rs
cryptonemo: > When we talked about it, I though we wanted a clear separating between "you run it with t_aux" and "you run it without t_aux". The "no t_aux" code path, never ever touches that file. In case there is a t_aux file, I would expect that it's more often accidental, rather then intentional.

Since the case of the 'no t_aux' doesn't touch the file, it shouldn't care that it's present.  Warning that it's present and unused is ok, but any failure caused by it being present will be a problem considering that just about every sealed and updated sector that exists today has it present.  There's no world where people can run this and be expected to re-seal everything, or manually remove every t_aux on disk in their operation to get the benefit.  Sure, someone may choose to start an operation with it enabled and that's the only legitimate case where 'no t_aux' files exist on disk.

DrPeterVanNostrand: I was thinking that falling back to `bincode::deserialize()` would  make the `fixed-rows-to-discard` feature mean something like "for all sectors going forward seal using the default value for `rows_to_discard`, but for all previously sealed sectors (e.g. during sector updates) use that sector's t_aux file (if it exists)".

vmx: Alright, I've added the fallback. I squashed the commits as I wanted to change the commit to reflect the reality. The diff between before and after the force push is pretty clean though: https://github.com/filecoin-project/rust-fil-proofs/compare/2ee6eda8e577728a3f8dcf3e732859e1442ec201..b02359f97effeca0f84ff95543284527b04975d4

filecoin-proofs/src/api/util.rs
```diff
@@ -113,3 +160,67 @@ pub(crate) fn persist_t_aux<Tree: MerkleTreeTrait>(
...
+            .into_path();
+
+        // Check if getting t_aux if no such file exists returns the default values.
+        let default_t_aux = TemporaryAux::<SectorShape32GiB, DefaultPieceHasher>::new(
+            sector_size / NODE_SIZE,
+            11,
+            cache_dir.clone(),
+        );
+        let t_aux = get_t_aux::<SectorShape32GiB>(&cache_dir, sector_size as u64)
+            .expect("t_aux should have been read");
```
DrPeterVanNostrand, line 204: Same comment: "t_aux should not have been read"?

filecoin-proofs/src/api/util.rs
DrPeterVanNostrand: Thanks for making the change. Everything looks good to me

filecoin-proofs/src/api/util.rs
```diff
@@ -113,3 +160,67 @@ pub(crate) fn persist_t_aux<Tree: MerkleTreeTrait>(
...
+
+    use crate::SectorShape32GiB;
+
+    /// Testing whether the default values are set if there's no `t_aux` file.
+    #[test]
+    fn test_get_t_aux_defaults() {
+        let dir_does_not_exist = Path::new("/path/does/not/exist");
+        let sector_size: usize = 34359738368;
+        let t_aux = get_t_aux::<SectorShape32GiB>(dir_does_not_exist, sector_size as u64)
+            .expect("t_aux should have been read");
```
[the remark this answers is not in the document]
vmx: I was doing "expect as pre-condition" style: https://doc.rust-lang.org/std/error/index.html#common-message-styles (got there from https://stackoverflow.com/questions/66362625/why-is-rusts-expect-called-expect.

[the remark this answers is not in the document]
DrPeterVanNostrand: Ok, sounds good

filecoin-proofs/src/api/seal.rs
```diff
@@ -299,7 +301,8 @@ where
 
     // Persist p_aux and t_aux here
     util::persist_p_aux::<Tree>(&p_aux, cache_path.as_ref())?;
-    util::persist_t_aux::<Tree>(&t_aux, cache_path.as_ref())?;
+    #[cfg(not(feature = "fixed-rows-to-discard"))]
```
cryptonemo, line 304: Last nit on this -- could we instead have a `persist_t_aux` defined for this feature config that is more or less an inline'd no-op?  I think the switching at this level is still not optimal because it's not clear that `fixed-rows-to-discard` means we don't need `t_aux`, although this is an initial case where that's true.  It also consolidates all switching on this config related to the p_aux/t_aux to the util methods.

vmx: > I think the switching at this level is still not optimal because it's not clear that `fixed-rows-to-discard` means we don't need `t_au`

I think of it as the reverse. If I would read the code and see a `persist_t_aux` function call, i would think "oh, t_aux is always persisted". Only if I then look at the implementation I see that it's actually not persisted in some cases. In the current code I see it right away that if that feature is enabled, nothing will be persisted. I would even see that there isn't even a persist_t_aux implementation at all.

filecoin-proofs/src/api/util.rs
```diff
@@ -100,7 +98,56 @@ pub(crate) fn get_t_aux<Tree: MerkleTreeTrait>(
...
+        "Instantiating TemporaryAux from default values with cache_path at {:?}",
+        cache_path
+    );
+    let t_aux_path = cache_path.join(CacheKey::TAux.to_string());
+    if Path::new(&t_aux_path).exists() {
+        log::warn!(
+            "`t_aux` file exists, use that file instead of the default values. t_aux={:?}",
+            &t_aux_path
+        );
+        read_t_aux_file(cache_path)
```
cryptonemo, line 120: Is this what we wanted here?  I thought if it existed with the config option set, it would warn loudly that it exists and is being ignored and then continue with the generated default?

filecoin-proofs/src/api/util.rs
```diff
@@ -113,3 +160,67 @@ pub(crate) fn persist_t_aux<Tree: MerkleTreeTrait>(
...
+        assert_eq!(
+            t_aux.tree_r_last_config.rows_to_discard,
+            expected_rows_to_discard
+        );
+    }
+
+    /// Testing whether the values from the `t_aux` file are used in case there is any.
+    #[test]
+    fn test_get_t_aux_file_exists() {
+        let sector_size = 34359738368;
```
cryptonemo, line 192: Can we use the existing constants that are defined?

filecoin-proofs/src/api/util.rs
vmx: That's how I understood you and @DrPeterVanNostrand. If we would continue with the default values (and the values in t_aux were different) then things might fail at later stages. I thought that's what you wanted to prevent.

cryptonemo: I could see the case for either -- but if we're going with the 'supraseal will never touch t_aux` idea, then I think what I said is consistent.  If it's better to support what's there, I'm ok with that, but it's a different model

storage-proofs-core/src/settings.rs
```diff
@@ -10,6 +10,7 @@ lazy_static! {
 
 const SETTINGS_PATH: &str = "./rust-fil-proofs.config.toml";
 const PREFIX: &str = "FIL_PROOFS";
+pub const DEFAULT_ROWS_TO_DISCARD: u32 = 2;
```
cryptonemo, line 13: Could this constant be moved to `storage-proofs-core/src/util.rs` where it would be better suited in the `default_rows_to_discard` method instead of the other way around?  Then settings would use that constant

filecoin-proofs/src/api/util.rs
cryptonemo: > That's how I understood you and @DrPeterVanNostrand. If we would continue with the default values (and the values in t_aux were different) then things might fail at later stages. I thought that's what you wanted to prevent.

I'm ok with it -- it's just different than the 'fixed-rows-to-discard requiring the defaults only and ignoring t_aux' model we also talked about.  I don't know yet which has the bigger potential for breakage :thinking:

vmx: I would prefer the "supraseal will never touch t_aux" idea. But wasn't it then previously doing what you now describe? If I look at https://github.com/filecoin-project/rust-fil-proofs/blob/2ee6eda8e577728a3f8dcf3e732859e1442ec201/filecoin-proofs/src/api/util.rs#L101-L106 (that's the old version), there is a warning that the file exists, but then it keeps on going with the defaults.

What I would prefer (but I can see the case of not doing that), would be to error in case the file exists, to indicate that you should enable this feature only on new sectors.

cryptonemo: Ok, I was confused on the old version then -- I thought it would error if `t_aux` existed, which is the worst option since it always exists for current sectors.  I think I said I'd prefer it to warn loudly that it was being ignored and continue with the defaults in that case.

That said, I can now see the case for warning, but then using the existing (i.e. as it is now).  It's a fine point, but we need (documented) clarity either way.  Maybe I'm leaning toward where it is now (using existing t_aux if found) because it has a better chance of working on installs where a non-standard rows_to_discard was used on the existing sealed sectors (and going forward with supraseal, it would start using the default) :thinking:

storage-proofs-core/src/settings.rs
vmx: Works for me, just a note that then you kind of have a circular dependency. `util` depends on `settings` and `settings` depend on `util`.

cryptonemo: No, if the constant is moved to util, settings will use the constant and util does not require anything from settings.

cryptonemo (approved):

PR #1720, opened 2023-09-22
refactor: clearer Synthetic PoRep separation

vmx:
The Synthetic PoRep code is interleaved with the original PoRep code. This commit refactors the code, so that it's clearer what the differences between Synthetic and Non-Synthetic PoRep is, without the need to read through large parts of the code, looking for if-statements.

It also becomes more apparent in the code, that there are two "modes" of the `prove_layers` function in case of the Synthetic PoRep. One generation step in case there's no seed and one extracting step when there's a seed.

There no longer is a fallback in case from Synthetic PoRep to Non-Synthetic PoRep in case the Synthetic PoRep Proofs file cannot be read, but when the labels are set. I consider that a feature as if this case happens, your system likelt has some problems, so it's better to error early.

### DIFF
--- a/storage-proofs-porep/src/stacked/vanilla/params.rs
+++ b/storage-proofs-porep/src/stacked/vanilla/params.rs
@@ -132,9 +132,9 @@ pub struct PublicInputs<T: Domain, S: Domain> {
 
 impl<T: Domain, S: Domain> PublicInputs<T, S> {
     /// If the porep challenge randomness `self.seed` is set, this method returns the porep
-    /// challenges for partition `k`; otherwise if `self.seed` is `None`, returns the entire
-    /// synthetic challenge set. Note synthetic challenges are generated in a single partition
-    /// `k = 0`.
+    /// challenges for partition `k` (for syth and non-syth poreps); otherwise if `self.seed` is
+    /// `None`, returns the entire synthetic challenge set. Note synthetic challenges are generated
+    /// in a single partition `k = 0`.
     pub fn challenges(
         &self,
         layer_challenges: &LayerChallenges,
--- a/storage-proofs-porep/src/stacked/vanilla/params.rs
+++ b/storage-proofs-porep/src/stacked/vanilla/params.rs
@@ -132,7 +132,7 @@ pub struct PublicInputs<T: Domain, S: Domain> {
 
 impl<T: Domain, S: Domain> PublicInputs<T, S> {
     /// If the porep challenge randomness `self.seed` is set, this method returns the porep
-    /// challenges for partition `k` (for syth and non-syth poreps); otherwise if `self.seed` is
+    /// challenges for partition `k` (for synth and non-synth poreps); otherwise if `self.seed` is
     /// `None`, returns the entire synthetic challenge set. Note synthetic challenges are generated
     /// in a single partition `k = 0`.
     pub fn challenges(
--- a/storage-proofs-porep/src/stacked/vanilla/proof.rs
+++ b/storage-proofs-porep/src/stacked/vanilla/proof.rs
@@ -105,66 +105,87 @@ impl<'a, Tree: 'static + MerkleTreeTrait, G: 'static + Hasher> StackedDrg<'a, Tr
         partition_count: usize,
     ) -> Result<Vec<Vec<Proof<Tree, G>>>> {
         assert!(layers > 0);
-
-        if !layer_challenges.use_synthetic {
-            // This needs to be relaxed now since the layers may not exist in the synth porep case
-            assert_eq!(t_aux.labels.len(), layers);
-        }
-
-        let graph_size = graph.size();
-
         // Sanity checks on restored trees.
         assert!(pub_inputs.tau.is_some());
-        // Skip this check in the case of synthetic porep
-        if t_aux.tree_d.is_some() {
+
+        if layer_challenges.use_synthetic {
+            // If there are no synthetic vanilla proofs stored on disk yet, generate them.
+            if pub_inputs.seed.is_none() {
+                info!("generating synthetic vanilla proofs in a single partition");
+
+                let vanilla_proofs = Self::prove_layers_generate(
+                    graph,
+                    pub_inputs,
+                    p_aux.comm_c,
+                    t_aux,
+                    layer_challenges,
+                    layers,
+                    partition_count,
+                )?;
+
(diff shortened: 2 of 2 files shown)


vmx:
As this PR wasn't reviewed yet, I did dare to rebase it. I also pushed a new commit. It's easiest reviewed if you look at those two commits separately. The second one is again easiest with whitespace only changes hidden: https://github.com/filecoin-project/rust-fil-proofs/pull/1720/commits/4f4621681dc7188074956db78b5c20cf786e9c67?diff=split&w=1

storage-proofs-porep/src/stacked/vanilla/proof.rs
```diff
@@ -105,66 +105,107 @@ impl<'a, Tree: 'static + MerkleTreeTrait, G: 'static + Hasher> StackedDrg<'a, Tr
...
+            // corresponding to the porep challlenge set.
+            else {
+                Self::read_porep_proofs_from_synth(
+                    graph.size(),
+                    pub_inputs,
+                    layer_challenges,
+                    t_aux.synth_proofs_path(),
+                    partition_count,
+                )
+                .map_err(|error| {
```
DrPeterVanNostrand, line 153: In the case where Synth-PoRep is enabled and the challenge seed is set, but synthetic challenge proofs do not exist on disk, I think the fallback behavior should be to run normal PoRep (rather than return an error). In the event where, while waiting for the challenge seed on chain, a user's synthetic challenge proofs were moved/removed from their expected location, returning an error here would run in an endless loop (assuming the user were to rerun `prove_layers` following the read synth-proofs error).

vmx: If a user runs a Synthetic PoRep, you would do that in order to free space. So you would remove the tree files/layers once the proofs are generated. If you move/remove those proofs and fallback to non-Synthetic PoRep, it would likely fail, because the original files are not there anymore. I would find that behaviour more confusing/hard to follow that failing early on the Synthetic PoRep code path.

This is kind of a discussion, whether we want "fail early" behaviour, which I prefer as I find it easier to debug/finding root cases if it happend. Vs. "best effort", which might then make things fail later on in unexpected ways.

DrPeterVanNostrand: Yeah, I think that it's totally debatable, and that the error makes sense too. Do you you have an opinion about falling back to standard PoRep only when the SDR layers weren't deleted? Or do you think an error should be returned in that case too?

vmx: > Do you you have an opinion about falling back to standard PoRep only when the SDR layers weren't deleted? Or do you think an error should be returned in that case too?

That might be a good middle ground. Though if it would be me, I think would still return early as I really see it as some misconfiguration/user error which under normal operations should never happen. And If someone did something wrong, they probably want to find out about it early. I'm also thinking about potential bug reports, which might be easier to tackle with early failure.

vmx:
The force push has just been a rebase, without any other code changes.

storage-proofs-porep/src/stacked/vanilla/params.rs
```diff
@@ -132,9 +132,9 @@ pub struct PublicInputs<T: Domain, S: Domain> {
 
 impl<T: Domain, S: Domain> PublicInputs<T, S> {
     /// If the porep challenge randomness `self.seed` is set, this method returns the porep
-    /// challenges for partition `k`; otherwise if `self.seed` is `None`, returns the entire
-    /// synthetic challenge set. Note synthetic challenges are generated in a single partition
-    /// `k = 0`.
+    /// challenges for partition `k` (for syth and non-syth poreps); otherwise if `self.seed` is
```
DrPeterVanNostrand, line 135: Spelling: `syth` -> `synth`.

storage-proofs-porep/src/stacked/vanilla/proof.rs
DrPeterVanNostrand: I'm ok with erroring as it would force a user to immediately diagnose and fix the issue. My thinking for why standard PoRep should be run if the labels are still present on disk (with a logged warning saying "synth-porep was not run") is mainly because I think that re-sealing/labeling should be avoided when possible. However, if that fallback behavior is repeatedly taken, a user may be mislead into thinking that they are using synth-porep when in reality they aren't.

Maybe @cryptonemo has an opinion, but if not, I'm happy to approve.

cryptonemo: My thought on it is that we should fallback to PoRep if it's possible and warn/log very loudly that something is misconfigured and SynthPoRep cannot be used.  I agree that will complicate potential code/testing paths, but shouldn't complicate much past having the user reading the logs and clearly seeing there's a problem.

Maybe the more ideal answer from a user perspective is to check if the synth_proofs data exists *in the clear data call (when enabled of course)* and log loudly there that they cannot be cleared because synth_proofs data doesn't exist -- and then we have the logic to fall back to PoRep because something is wildly misconfigured ???

vmx: > I agree that will complicate potential code/testing paths, but shouldn't complicate much past having the user reading the logs and clearly seeing there's a problem.

I disagree here. Especially with Ni-PoRep there's yet another code path. I find the current code pretty clean, you can clearly see the different PoRep cases and reason about what happens when. With certain fallback it likely be a lot of additional back and forth, which I think is just not worth it.

From a users perspective I see it as: if i've misconfigured wildly I want to immediately know it, rather then having a system that tries random things just in hope it might recover somehow, but might still end up in a broken state. An example for today: SupraSeal's PC2 does not support Synthetic PoRep. When run with Lotus and SyntheticPoRep enabled, it runs into a seemingly endless loop, but then does something (the wrong thing). Wouldn't it be much better if it would just say: "error, you try to run Synthetic PoRep although it's not supported"?

So if there's *any* chance you could live with the current solution I'd prefer keeping it the current way.

If you can't live with it, should we then also start thinking about how it will look like if we also add Ni-PoRep. Should there also be fallbacks?

cryptonemo: Ok, that's fair.  Thanks!

cryptonemo (approved):

PR #1721, opened 2023-09-26
refactor: refactor p_aux/t_aux handling into the util module

vmx:
`p_aux` and `t_aux` is dealt with throughout the code. In the update proofs API they are nicely put into functions. Move those functions into the common utility module and then use those in all places.

The update proofs API puts a restriction on using the Poseidon hasher. This restriction is kept as it makes sense for all cases where `p_aux` and `t_aux` is used.

### DIFF
--- a/filecoin-proofs/src/api/fake_seal.rs
+++ b/filecoin-proofs/src/api/fake_seal.rs
@@ -1,15 +1,14 @@
 use std::fs::File;
-use std::io::Write;
 use std::path::Path;
 
-use anyhow::{Context, Result};
-use bincode::serialize;
+use anyhow::Result;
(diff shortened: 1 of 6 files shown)


vmx:
To me it made sense to restrict it to the PoseidonHasher. I think it made things clearer. Though it would be a breaking change for filecoin proofs API, hence I'll redo that PR without restricting things to the PoseidonHasher.

vmx:
Ready for review, it doesn't contain any breaking changes now.

DrPeterVanNostrand:
Looks good to me, but Nemo should probably also do a final review.

cryptonemo (approved):

PR #1722, opened 2023-09-26
ci: update macos runner to Apple silicon

galargh:
CircleCI is removing support for Apple Intel executors on Oct 2 - see https://discuss.circleci.com/t/macos-resource-deprecation-update/46891. After that date only Apple Silicon executors will be available. This PR ensures the macos builds are executed on Apple Silicon executors only.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -289,7 +289,7 @@ jobs:
     macos:
       xcode: "13.4.1"
     working_directory: ~/crate
-    resource_class: large
+    resource_class: macos.x86.medium.gen2
     environment: *setup-env
     steps:
       - run:


vmx:
The commit message and description are not describing what the change is doing. It's still using x86 [which will be deprecated in January](https://discuss.circleci.com/t/macos-intel-support-deprecation-in-january-2024/48718). I suggest we move to a `macos.m1.medium.gen1` so that we don't need to change in January again.

galargh:
As per https://github.com/filecoin-project/rust-fil-proofs/pull/1722#issuecomment-1739139554, it turns out I misunderstood the announcement (timeline) slightly. I'm going to close this PR now and recreate it with correct commit messages and description. I'm sorry for the extra noise.

PR #1730, opened 2023-10-31
refactor: split challenges and layer information

vmx:
At the moment the challenges and the information about the number of layers is put into a single `LayerChallenges` struct. This misleads into thinking that the challenge generation is related to the number of layers, but it is *not*. `LayerChallenges` now only contains the information about challenges.

This makes the code eaier to follow, as it's now clearer, whether a function needs the layers or the challenges.

### DIFF
--- a/storage-proofs-porep/src/stacked/vanilla/proof.rs
+++ b/storage-proofs-porep/src/stacked/vanilla/proof.rs
@@ -100,15 +100,15 @@ impl<'a, Tree: 'static + MerkleTreeTrait, G: 'static + Hasher> StackedDrg<'a, Tr
         pub_inputs: &PublicInputs<<Tree::Hasher as Hasher>::Domain, <G as Hasher>::Domain>,
         p_aux: &PersistentAux<<Tree::Hasher as Hasher>::Domain>,
         t_aux: &TemporaryAuxCache<Tree, G>,
-        layer_challenges: &LayerChallenges,
-        layers: usize,
+        challenges: &LayerChallenges,
+        num_layers: usize,
         partition_count: usize,
     ) -> Result<Vec<Vec<Proof<Tree, G>>>> {
-        assert!(layers > 0);
+        assert!(num_layers > 0);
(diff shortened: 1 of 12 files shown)


storage-proofs-porep/src/stacked/vanilla/proof.rs
```diff
@@ -100,42 +100,40 @@ impl<'a, Tree: 'static + MerkleTreeTrait, G: 'static + Hasher> StackedDrg<'a, Tr
...
                 assert_eq!(partition_count, 1);
 
                 let comm_r = pub_inputs.tau.as_ref().expect("tau is set").comm_r;
                 // Derive the set of challenges we are proving over.
-                let challenges = layer_challenges.derive_synthetic(
-                    graph.size(),
-                    &pub_inputs.replica_id,
-                    &comm_r,
-                );
+                let challenge_positions =
```
DrPeterVanNostrand, line 119: At least to me, renaming `challenges` to `challenge_positions` (here and below) makes the code slightly less clear.

DrPeterVanNostrand (changes requested):
Just one comment about naming, but otherwise looks good to me.

storage-proofs-porep/src/stacked/vanilla/proof.rs
vmx: Agreed, but `challenges` is what you put in as parameter, which is of type `LayerChallenges`. So I didn't want to re-assign something with the same name. I also don't really like the name, but it's the best i could come up with.

Would another name work? `challenge_offsets`? `derived`? `derived_challenges`? `derived_positions`? `derived_offsets`?

Or should we rename the input variable?

If you've any suggestions, please let me know.

DrPeterVanNostrand: Yeah, that makes sense. Maybe here `synth_challenges` and below `partition_challenges`, or just `challenge_nodes` in both?

DrPeterVanNostrand (approved):

PR #1731, opened 2023-10-31
refactor: better select challenges function

vmx:
The function to select the challenges changes a lot over time and now is hardly comprehensible. Hence it's refacored into a more understandable version. It uses a `div_ceil` function which is expecte to be included in the Rust standard library at some point in the future.

### DIFF
--- a/filecoin-proofs/src/parameters.rs
+++ b/filecoin-proofs/src/parameters.rs
@@ -104,20 +104,21 @@ pub fn setup_params(porep_config: &PoRepConfig) -> Result<stacked::SetupParams> 
     })
 }
 
+/// Division of x by y, rounding up.
+/// x and y must be > 0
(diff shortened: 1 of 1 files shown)


vmx:
> Just FYI if you're interested, I believe `challenges_count_all` is a leftover from when PoRep's design included a specific number of challenges per layer (and `challenges_count_all` meant the "total number of challenges across all layers per partition"). And a potential future refactor that I would also support is renaming `challenges_count_all` to something like `partition_challenge_count`.

Exactly! I tackle that in a later commit on the ni-porep work: https://github.com/filecoin-project/rust-fil-proofs/pull/1734/commits/d2505cb15fd3c83192caff3ca569b66767394812.

DrPeterVanNostrand (approved):

PR #1732, opened 2023-10-31
refactor: StackedCircuit only needs the number of layers

vmx:
There's no need to store the full public parameters within the `StackedCircuit` struct, as it only needs the number of layers.

### DIFF
--- a/filecoin-proofs/src/types/porep_config.rs
+++ b/filecoin-proofs/src/types/porep_config.rs
@@ -101,7 +101,7 @@ impl PoRepConfig {
 
         Ok(
             <StackedCompound<Tree, DefaultPieceHasher> as CacheableParameters<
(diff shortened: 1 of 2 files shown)


DrPeterVanNostrand (approved):
Looks good to me.

Theoretically I think that `StackedCircuit` doesn't even need to store `num_layers` because the circuit prover implicitly stores `num_layers` in the circuit's private inputs (i.e. in the column proofs) and the circuit verifier generates circuit public inputs by passing `pub_params` into `StackedCompound::generate_public_inputs`.

vmx:
@DrPeterVanNostrand good point! I've created https://github.com/filecoin-project/rust-fil-proofs/pull/1735 as I'm not sure if that's what you meant and also not sure if we really want to do that.

DrPeterVanNostrand (approved):

PR #1733, opened 2023-10-31
fix: remove with_feature

vmx:
The `PoRepConfig::with_feature()` function is not used anywhere, hence remove it.


DrPeterVanNostrand (approved):

cryptonemo (approved):
As we use more features in the future, this is a forward thinking API since we need decent ways to enable them.

PR #1734, opened 2023-10-31
feat: non-interactive PoRep

vmx:
Adding non-interactive PoRep. It's similar to the interactive PoRep, but
it performance 12.6x the amount of proofs. In order to not overwhelm the
system with too many parallel proofs, they are split into batches of 10,
which then leads to a similar resource consumption per batch as for a
single interactive PoRep.

### DIFF
--- a/filecoin-proofs/src/api/seal.rs
+++ b/filecoin-proofs/src/api/seal.rs
@@ -33,7 +33,6 @@ use storage_proofs_porep::stacked::{
 use storage_proofs_update::vanilla::prepare_tree_r_data;
 use typenum::{Unsigned, U11, U2};
 
-use crate::POREP_MINIMUM_CHALLENGES;
 use crate::{
     api::util::{get_aggregate_target_len, pad_inputs_to_target, pad_proofs_to_target},
     api::{as_safe_commitment, commitment_from_fr, get_base_tree_leafs, get_base_tree_size, util},
@@ -570,19 +569,50 @@ pub fn seal_commit_phase2<Tree: 'static + MerkleTreeTrait>(
 
     // Verification is cheap when parameters are cached,
     // and it is never correct to return a proof which does not verify.
-    verify_seal::<Tree>(
-        porep_config,
-        comm_r,
-        comm_d,
-        prover_id,
-        sector_id,
-        ticket,
-        seed,
-        &buf,
-    )
-    .context("post-seal verification sanity check failed")?;
-
-    let out = SealCommitOutput { proof: buf };
+    // Non-interactive PoRep is an aggregated proof, hence we also need a different code path for
+    // the verifucation.
+    let out = if porep_config.feature_enabled(ApiFeature::NonInteractivePoRep) {
+        let aggregated = aggregate_seal_commit_proofs::<Tree>(
+            porep_config,
+            &[comm_r],
+            &[seed],
+            &[SealCommitOutput { proof: buf }],
+            groth16::aggregate::AggregateVersion::V2,
+        )?;
+        let inputs = get_seal_inputs::<Tree>(
+            porep_config,
+            comm_r,
+            comm_d,
+            prover_id,
+            sector_id,
+            ticket,
+            seed,
+        )?;
+        let is_valid = verify_aggregate_seal_commit_proofs::<Tree>(
+            porep_config,
+            aggregated.clone(),
+            &[comm_r],
+            &[seed],
+            inputs,
+            groth16::aggregate::AggregateVersion::V2,
+        )
+        .context("post-seal aggregation verification sanity check failed")?;
+        ensure!(is_valid, "post seal aggregation verifies");
+        SealCommitOutput { proof: aggregated }
+    } else {
+        verify_seal::<Tree>(
+            porep_config,
+            comm_r,
+            comm_d,
+            prover_id,
+            sector_id,
+            ticket,
+            seed,
+            &buf,
+        )
+        .context("post-seal verification sanity check failed")?;
+        SealCommitOutput { proof: buf }
+    };
 
     info!("seal_commit_phase2:finish: {:?}", sector_id);
     Ok(out)
@@ -892,6 +922,28 @@ pub fn verify_seal<Tree: 'static + MerkleTreeTrait>(
 ) -> Result<bool> {
     info!("verify_seal:start: {:?}", sector_id);
 
+    // Non-interactive PoReps are aggregated, but it should be possible to use the usual PoRep
+    // APIs, hence branch out here and not one layer higher.
+    if porep_config.feature_enabled(ApiFeature::NonInteractivePoRep) {
+        let inputs = get_seal_inputs::<Tree>(
+            porep_config,
+            comm_r_in,
+            comm_d_in,
+            prover_id,
+            sector_id,
+            ticket,
+            seed,
+        )?;
+        return verify_aggregate_seal_commit_proofs::<Tree>(
+            porep_config,
+            proof_vec.to_vec(),
+            &[comm_r_in],
+            &[seed],
+            inputs,
+            groth16::aggregate::AggregateVersion::V2,
+        );
+    }
+
     ensure!(comm_d_in != [0; 32], "Invalid all zero commitment (comm_d)");
     ensure!(comm_r_in != [0; 32], "Invalid all zero commitment (comm_r)");
     ensure!(!proof_vec.is_empty(), "Invalid proof bytes (empty vector)");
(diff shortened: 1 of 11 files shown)


fil-proofs-tooling/src/bin/benchy/porep.rs
```diff
@@ -78,7 +78,9 @@ fn get_porep_config(sector_size: u64, api_version: ApiVersion, use_synthetic: bo
     // Replicate the staged sector, write the replica file to `sealed_path`.
     let mut config = PoRepConfig::new_groth16(sector_size, arbitrary_porep_id, api_version);
     if use_synthetic {
-        config.enable_feature(ApiFeature::SyntheticPoRep);
+        config
```
cryptonemo, line 81: We should be able to re-factor these config updates using `new_groth16_with_features` now

vmx: That's a bigger refactor, hence I've opened a separate PR for it at https://github.com/filecoin-project/rust-fil-proofs/pull/1742.

vmx:
OK, I've rebased the commit and added a proper commit message, which I've also added to the description of this PR.

filecoin-proofs/src/constants.rs
```diff
@@ -144,6 +104,43 @@ lazy_static! {
...
+        _ => panic!("invalid sector size"),
+    }
+}
+
+/// Returns the minimum number of challenges used for the non-interactive PoRep fo a certain sector
+/// size.
+pub(crate) const fn get_porep_non_interactive_minimum_challenges(sector_size: u64) -> usize {
+    match sector_size {
+        SECTOR_SIZE_2_KIB | SECTOR_SIZE_4_KIB | SECTOR_SIZE_16_KIB | SECTOR_SIZE_32_KIB
+        | SECTOR_SIZE_8_MIB | SECTOR_SIZE_16_MIB | SECTOR_SIZE_512_MIB | SECTOR_SIZE_1_GIB => 4,
```
cryptonemo, line 123: Was the test sector size challenge count specified in the spec?  Can't recall

cryptonemo (commented):
Generally looks good to me -- if/when Jake approves, ship it!

filecoin-proofs/src/constants.rs
vmx: No the test sector size is not specified in the spec, but I thought I pick one different from the interactive porep.

filecoin-proofs/src/types/porep_config.rs
```diff
@@ -78,8 +78,8 @@ impl PoRepConfig {
         porep_id: [u8; 32],
         api_version: ApiVersion,
         api_features: Vec<ApiFeature>,
-    ) -> Self {
-        Self {
+    ) -> Result<Self> {
```
DrPeterVanNostrand, line 81: Mentioning this just in case you prefer it; an early idea for `ApiFeature` was to have a way to check for conflicting features in `storage_proofs_core::api_version`. Based on your changes, I think that it would look something like...
```rust
impl ApiFeature {
    fn conflicting_features(&self) -> std::slice::Iter<Self> {
        match self {
            ApiFeature::SyntheticPoRep => [ApiFeature::NonInteractivePoRep].iter(),
            ApiFeature::NonInteractivePoRep => [ApiFeature::SyntheticPoRep].iter(),
            // _ => [].iter(),
        }
    }
}

pub fn check_conflicting_api_features(enabled_feats: &[ApiFeature]) -> Result<()> {
    let n = enabled_feats.len();
    if n < 2 {
        return Ok(());
    }
    for i in 0..n - 1 {
        let feat = &enabled_feats[i];
        let rem = &enabled_feats[i + 1..];
        for conflict in feat.conflicting_features() {
            if rem.contains(conflict) {
                return Err(anyhow!(
                    "The features `{feat}, {conflict}` cannot be enabled simultaneously"
                ));
            }
        }
    }
    Ok(())
}
```

```rust
impl PoRepConfig {
    pub fn new_groth16_with_features(
        sector_size: u64,
        porep_id: [u8; 32],
        api_version: ApiVersion,
        mut api_features: Vec<ApiFeature>,
    ) -> Result<Self> {
        api_features.dedup();
        storage_proofs_core::api_version::check_conflicting_api_features(&api_features)?;
        let partitions = if api_features.contains(&ApiFeature::SyntheticPoRep) {
            // Synth-PoRep partition count
        } else if api_features.contains(&ApiFeature::NonInteractivePoRep) {
            // NI-PoRep partition count
        } else {
            // Interactive-PoRep partition count
        };
        Ok(PoRepConfig {
            sector_size: SectorSize(sector_size),
            partitions,
            porep_id,
            api_version,
            api_features,
        })
    }
}
```

DrPeterVanNostrand: I'm not saying that I prefer the above, but it does have the benefit of specifying conflicting features in the same file as `enum ApiFeature` (and where the features' `ApiVersion` ranges are specified).

filecoin-proofs/src/constants.rs
```diff
@@ -156,6 +153,9 @@ pub const MINIMUM_RESERVED_BYTES_FOR_PIECE_IN_FULLY_ALIGNED_SECTOR: u64 =
 /// The minimum size a single piece must have before padding.
 pub const MIN_PIECE_SIZE: UnpaddedBytesAmount = UnpaddedBytesAmount(127);
 
+/// The maximum number of challenges per partition the circuits can work with.
+pub(crate) const MAX_CHALLENGES_PER_PARTITION: u8 = 18;
```
DrPeterVanNostrand, line 157: To me, the word `MAX` implies that a PoRep circuit can prove less than 18 challenges, whereas a name like `POREP_CHALLENGES_PER_PARTITION` would imply that a PoRep circuit can prove exactly 18 challenges.

vmx: Yes, that's the reason I used `MAX`, as less challenges also seem to work.

filecoin-proofs/src/types/porep_config.rs
vmx: That sounds like a good idea, as it would "scale" better in case there are more features in the future. Let me try that out.

vmx: Alright, I've pushed a new commit, which is a combination of what I had an your idea.

filecoin-proofs/src/api/seal.rs
```diff
@@ -570,19 +569,50 @@ pub fn seal_commit_phase2<Tree: 'static + MerkleTreeTrait>(
...
+            aggregated.clone(),
+            &[comm_r],
+            &[seed],
+            inputs,
+            groth16::aggregate::AggregateVersion::V2,
+        )
+        .context("post-seal aggregation verification sanity check failed")?;
+        ensure!(is_valid, "post seal aggregation verifies");
+        SealCommitOutput { proof: aggregated }
+    } else {
```
DrPeterVanNostrand, line 602: If the logic to verify an aggregate proof for NI-PoRep is also being added to `verify_seal`, would something like this work to avoid the duplicate code?
```rust
proof.write(&mut buf)?;

// Non-interactive PoRep is an aggregated proof.
if porep_config.feature_enabled(ApiFeature::NonInteractivePoRep) {
    buf = aggregate_seal_commit_proofs::<Tree>(
        porep_config,
        &[comm_r],
        &[seed],
        &[SealCommitOutput { proof: buf }],
        groth16::aggregate::AggregateVersion::V2,
    )?;
}

let is_valid = verify_seal::<Tree>(..., &buf)
    .context("post-seal verification sanity check failed")?;
ensure!(is_valid, "post seal aggregation verifies");

info!("seal_commit_phase2:finish: {:?}", sector_id);
Ok(SealCommitOutput { proof: buf })
```
But, there may be a reason that I'm not seeing as to why the aggregate proof verification logic for NI-PoRep is needed both here and `verify_seal`.

DrPeterVanNostrand (approved):
Looks good to me. Thanks for the `ApiFeature` addition. I just have the one comment about `verify_seal`, but I'll approve now anyway (and if you think changes should be made, will approve again).

filecoin-proofs/src/api/seal.rs
DrPeterVanNostrand: Just adding, by "duplicate logic" I mean the calls to `get_seal_inputs` and `verify_aggregate_seal_commit_proofs` in this `if`-branch.

vmx: Wow, that was a very good catch! That happens when you work on the code for weeks and in the end you don't see the obvious. I've changed it and pushed it (let's hope I didn't break anything).

DrPeterVanNostrand (approved):

PR #1735, opened 2023-11-02
refactor: get number of layers from private inputs

vmx:
Instead of passing in the number of layers through the public params, get the number of layers from the private inputs that are required anyway. The number of layers equals the number of columns of a columns proof.

---

I first wasn't really convinced that it makes sense, but after seeing that now [`circuit()` doesn't need the public params anymore](https://github.com/filecoin-project/rust-fil-proofs/compare/stacked-circuit-num-layers...stacked-circuit-remove-num-layers?expand=1#diff-ffb695bda5cb5ab675888466cd786c2dd9f6d1e9e60d52ce0aa4300c3020170bL296) it looks like a win to me.

### DIFF
--- a/storage-proofs-porep/src/stacked/circuit/params.rs
+++ b/storage-proofs-porep/src/stacked/circuit/params.rs
@@ -99,7 +99,6 @@ impl<Tree: MerkleTreeTrait, G: 'static + Hasher> Proof<Tree, G> {
     pub fn synthesize<CS: ConstraintSystem<Fr>>(
         self,
         mut cs: CS,
-        layers: usize,
         comm_d: &AllocatedNum<Fr>,
         comm_c: &AllocatedNum<Fr>,
         comm_r_last: &AllocatedNum<Fr>,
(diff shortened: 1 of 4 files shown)


storage-proofs-porep/src/stacked/circuit/params.rs
```diff
@@ -137,12 +136,12 @@ impl<Tree: MerkleTreeTrait, G: 'static + Hasher> Proof<Tree, G> {
...
 
         // Private Inputs for the DRG parent nodes.
-        let mut drg_parents = Vec::with_capacity(layers);
+        let num_layers = drg_parents_proofs[0].len();
+        let mut drg_parents = Vec::with_capacity(num_layers);
 
         for (i, parent) in drg_parents_proofs.into_iter().enumerate() {
             let (parent_col, inclusion_path) =
                 parent.alloc(cs.namespace(|| format!("drg_parent_{}_num", i)))?;
-            assert_eq!(layers, parent_col.len());
```
DrPeterVanNostrand, line 145: These asserts aren't totally necessary, but I still think they're useful as a sanity check that the Prover's witness (column proofs) was constructed correctly. Circuits can be hard to debug after they've been synthesized (e.g. an error during Groth16 proving because the Groth16 parameters don't match the synthesized circuit) and I've found that these `assert`s during synthesis can save debugging time later on.

DrPeterVanNostrand (dismissed):
Thanks, I've wanted this change for a **long** time. Just one comment, but otherwise looks good to me.

storage-proofs-porep/src/stacked/circuit/params.rs
vmx: I've removed it as `layers` and `parent_col.len()` is always true for the first item, as `layers` is `drg_parents_proofs[0].len()`. But I taken into account, that of course the other proofs could potentially be wrong. So I'm adding it back.

DrPeterVanNostrand: Cool, thanks!

DrPeterVanNostrand (approved):

PR #1742, opened 2023-12-18
refactor: use new_groth16_with_feature

vmx:
All users were using a hard-coded PoRep ID for `create_replica()` already, hence we can remove that parameter and use `get_porep_config()` directly.

### DIFF
--- a/fil-proofs-tooling/src/bin/benchy/porep.rs
+++ b/fil-proofs-tooling/src/bin/benchy/porep.rs
@@ -6,10 +6,10 @@ use std::time::{SystemTime, UNIX_EPOCH};
 use anyhow::{ensure, Context};
 use bincode::{deserialize, serialize};
 use fil_proofs_tooling::measure::FuncMeasurement;
-use fil_proofs_tooling::shared::{PROVER_ID, TICKET_BYTES};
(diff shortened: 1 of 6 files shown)


fil-proofs-tooling/src/bin/benchy/porep.rs
```diff
@@ -335,6 +323,12 @@ pub fn run_porep_bench<Tree: 'static + MerkleTreeTrait>(
     test_resume: bool,
     use_synthetic: bool,
 ) -> anyhow::Result<()> {
+    let features = if use_synthetic {
```
cryptonemo, line 326: Not a big deal, but eventually it would be good to see us reduce the boolean flags/checks and instead just pass in the enabled features for a pass through here.

vmx: I though about it, but then you'd need to change the whole CLI parsing.

cryptonemo (approved):

fil-proofs-tooling/src/bin/benchy/porep.rs
vmx: I've started working on it, but it's not trivial. Hence merging this one and going from there then.

PR #1744, opened 2023-12-20
refactor: remove unused PoSt code

vmx:
There are several Proof-of-Spacetime which were never be used in production, remove those in order to lower the maintenance burden.


DrPeterVanNostrand (approved):
I'm fine with removing Election and Rational-PoSt, but I think that we should wait until after break to merge.

cryptonemo (approved):

PR #1745, opened 2023-12-21
Various improvements

vmx:
This PR is a mix of smaller changes that I accumulated over the year. This PR should be reviewed on a commit by commit basis. I thought I bundle them in a single PR so that I don't have to rebase all of them all the time.

The review is not urgent, but I hope we can get all of those eventually merged.

### DIFF
--- a/storage-proofs-core/src/api_version.rs
+++ b/storage-proofs-core/src/api_version.rs
@@ -5,6 +5,8 @@ use std::str::FromStr;
 use anyhow::{format_err, Error, Result};
 use semver::Version;
 
+pub use bellperson::groth16::aggregate::AggregateVersion;
+
 /// The ApiVersion enum is used for mandatory changes that the network
 /// must use and recognize.
 ///
--- a/storage-proofs-core/src/merkle/mod.rs
+++ b/storage-proofs-core/src/merkle/mod.rs
@@ -2,9 +2,9 @@
 
 use std::fs::File;
 
+pub use filecoin_hashers::Hasher;
 pub use merkletree::store::{DiskStore, ExternalReader, Store};
 
-use filecoin_hashers::Hasher;
 use generic_array::typenum::{U0, U2};
 use merkletree::store::LevelCacheStore;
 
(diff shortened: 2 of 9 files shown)


storage-proofs-core/src/api_version.rs
```diff
@@ -5,6 +5,8 @@ use std::str::FromStr;
 use anyhow::{format_err, Error, Result};
 use semver::Version;
 
+pub use bellperson::groth16::aggregate::AggregateVersion;
```
DrPeterVanNostrand, line 8: Is this export being used anywhere in the PR?

storage-proofs-core/src/merkle/mod.rs
```diff
@@ -2,9 +2,9 @@
 
 use std::fs::File;
 
+pub use filecoin_hashers::Hasher;
```
DrPeterVanNostrand, line 5: Is this export being used anywhere in this PR? I'm not sure that I see it. Maybe it's being re-exported because it's part of the public Merkle tree API in `storage-proofs-core`?

vmx: It's explained in the commit message https://github.com/filecoin-project/rust-fil-proofs/pull/1745/commits/c92510e2029188f0ec88cc50e96db0b7410ac4b8:

> The `filecoin_hashers::Hasher` trait is needed a lot in downstream dependants when the domain of the Tree hasher needs to be accessed. Hence re-export it, so that those dependents don't need to depend directly on `filecoin_hashers` only due to this single trait.

storage-proofs-core/src/api_version.rs
vmx: It's explained in the commit message https://github.com/filecoin-project/rust-fil-proofs/pull/1745/commits/c92510e2029188f0ec88cc50e96db0b7410ac4b8:

> `bellperson`s `AggregateVersion` is part of the public API due to the `CompoundProof` trait. With re-exporting it, it's possible that downstream dependants like filecoin proofs API don't need to depend on `bellperson` directly, but can use this re-export instead.

PR #1748, opened 2024-04-05
Improve error handling for threaded verification errors

cryptonemo:
For synth porep, we switch to threaded/parallel proof verification as an optimization.  This takes it a step further and now allows errors to be propagated/returned properly.

Note: There is a FIXME in the code that will be removed after it's properly tested via filecoin-ffi testing.

### DIFF
--- a/storage-proofs-porep/src/stacked/vanilla/proof.rs
+++ b/storage-proofs-porep/src/stacked/vanilla/proof.rs
@@ -20,6 +20,8 @@ use merkletree::{
 use rayon::prelude::{
     IndexedParallelIterator, IntoParallelIterator, ParallelIterator, ParallelSliceMut,
 };
+#[cfg(any(feature = "cuda", feature = "multicore-sdr", feature = "opencl"))]
+use storage_proofs_core::settings::SETTINGS;
 use storage_proofs_core::{
     cache_key::CacheKey,
     data::Data,
@@ -451,23 +453,14 @@ impl<'a, Tree: 'static + MerkleTreeTrait, G: 'static + Hasher> StackedDrg<'a, Tr
 
                         // Check if a proof was detected as invalid
                         let invalid_comm_d_coordinate = invalid_comm_d.lock().expect("failed to get lock on invalid_comm_d");
-                        if invalid_comm_d_coordinate.failure_detected {
-                            return Err(anyhow!(
-                                "Invalid comm_d detected at challenge_index {}",
-                                invalid_comm_d_coordinate.challenge_index));
-                        }
+                        ensure!(!invalid_comm_d_coordinate.failure_detected, "Invalid comm_d detected at challenge_index {}",
+                                invalid_comm_d_coordinate.challenge_index);
                         let invalid_comm_r_coordinate = invalid_comm_r.lock().expect("failed to get lock on invalid_comm_r");
-                        if invalid_comm_r_coordinate.failure_detected {
-                            return Err(anyhow!(
-                                "Invalid comm_r detected at challenge_index {}",
-                                invalid_comm_r_coordinate.challenge_index));
-                        }
+                        ensure!(!invalid_comm_r_coordinate.failure_detected, "Invalid comm_r detected at challenge_index {}",
+                                invalid_comm_r_coordinate.challenge_index);
                         let invalid_encoding_proof_coordinate = invalid_encoding_proof.lock().expect("failed to get lock on invalid_encoding_proof");
-                        if invalid_encoding_proof_coordinate.failure_detected {
-                            return Err(anyhow!(
-                                "Invalid encoding proof generated at layer {}, challenge_index {}",
-                                invalid_encoding_proof_coordinate.layer, invalid_encoding_proof_coordinate.challenge_index));
-                        }
+                        ensure!(!invalid_encoding_proof_coordinate.failure_detected, "Invalid encoding proof generated at layer {}, challenge_index {}",
+                                invalid_encoding_proof_coordinate.layer, invalid_encoding_proof_coordinate.challenge_index);
                     }
 
                     Ok(Proof {
@@ -688,7 +681,6 @@ impl<'a, Tree: 'static + MerkleTreeTrait, G: 'static + Hasher> StackedDrg<'a, Tr
 
         #[cfg(feature = "multicore-sdr")]
         {
-            use storage_proofs_core::settings::SETTINGS;
             if SETTINGS.use_multicore_sdr {
                 info!("multi core replication");
                 create_label::multi::create_labels_for_encoding(
@@ -734,7 +726,6 @@ impl<'a, Tree: 'static + MerkleTreeTrait, G: 'static + Hasher> StackedDrg<'a, Tr
 
         #[cfg(feature = "multicore-sdr")]
         {
-            use storage_proofs_core::settings::SETTINGS;
             if SETTINGS.use_multicore_sdr {
                 info!("multi core replication");
                 create_label::multi::create_labels_for_decoding(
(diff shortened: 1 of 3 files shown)


cryptonemo:
Looks like all the CI failures are 'Infrastructure failures' and I noticed I can no longer restart failed jobs.  Everything passes locally.

magik6k (commented):
Can't `comm_d_proof_inner.validate` and the comm_r last asserts above fail in a similar way if the trees are corrupted?

cryptonemo:
> Can't `comm_d_proof_inner.validate` and the comm_r last asserts above fail in a similar way if the trees are corrupted?

Yes, there are other cases.  One thing (for testing) at a time.  Once wired through FFI and I'm sure it's propagated properly, adding in the others are straight-forward.

cryptonemo:
Good news is that everything worked wired through locally.  I'll look into the other cases.

storage-proofs-porep/src/stacked/vanilla/proof.rs
```diff
@@ -380,16 +393,31 @@ impl<'a, Tree: 'static + MerkleTreeTrait, G: 'static + Hasher> StackedDrg<'a, Tr
...
+                            let invalid_encoding_proof_inner = Arc::clone(&invalid_encoding_proof);
                             scope.execute(move || {
-                                assert!(
-                                    proof_inner.verify(replica_id, &labeled_node),
-                                    "Invalid encoding proof generated at layer {}",
-                                    layer,
-                                );
-                                trace!("Valid encoding proof generated at layer {}", layer);
+                                if !proof_inner.verify(replica_id, &labeled_node) {
+                                    let mut invalid = invalid_encoding_proof_inner.lock().expect("foo");
```
vmx, line 411: The "foo" should be something like "cannot acquire lock" or "lock to be acquired". (The same for other "foo" below).

vmx (commented):
The current code reports the most recent error only. A possible change would be to have a vector of errors and pushing them into a vector. Once done print the contents of the whole vector if it's non-empty.

vmx:
> > The current code reports the most recent error only. A possible change would be to have a vector of errors and pushing them into a vector. Once done print the contents of the whole vector if it's non-empty.
> 
> That's incorrect. Look closer.

Oh right. It's on the first error it encounters. That's probably good enough.

storage-proofs-porep/src/stacked/vanilla/proof.rs
```diff
@@ -402,6 +430,14 @@ impl<'a, Tree: 'static + MerkleTreeTrait, G: 'static + Hasher> StackedDrg<'a, Tr
                                 parents_data_full,
                             ));
                         }
+
+                        // Check if a proof was detected as invalid
+                        let invalid_encoding_proof_coordinate = invalid_encoding_proof.lock().expect("foo");
+                        if invalid_encoding_proof_coordinate.failure_detected {
+                            return Err(anyhow!(format!(
```
vmx, line 437: The `format!()` is not needed, `anyhow!()` supports a formatting string.

vmx:
> > > The current code reports the most recent error only. A possible change would be to have a vector of errors and pushing them into a vector. Once done print the contents of the whole vector if it's non-empty.
> > 
> > 
> > That's incorrect. Look closer.
> 
> Oh right. It's on the first error it encounters. That's probably good enough.

I re-read the code again. Now I think it collects all errors. But wow, there must be a way to make all this easier to follow.

cryptonemo:
Updated the error handling on all 4 of the threaded verifiers added for synth porep proofs.  Wasted some cycles looking into it and of the 4 verifiers that were made threaded for consistency, only 1 (or 2) had the most perf impact.  I recall this now, but again, the parallelism was added across the board for consistency, and now we consistently propagate errors across them all.  There's also still a FIXME in place (on purpose) and some additional testing is needed (via my local `api` and `ffi` branches, soon to be PRs).

storage-proofs-porep/src/stacked/vanilla/proof.rs
```diff
@@ -592,6 +688,7 @@ impl<'a, Tree: 'static + MerkleTreeTrait, G: 'static + Hasher> StackedDrg<'a, Tr
 
         #[cfg(feature = "multicore-sdr")]
         {
+            use storage_proofs_core::settings::SETTINGS;
```
vmx, line 691: I guess those are added due to Clippy warnings. What about importing it at the top again as:

```rust
#[cfg(any(feature = "cuda", feature = "multicore-sdr", feature = "opencl"))]
use storage_proofs_core::settings::SETTINGS;
```

storage-proofs-porep/src/stacked/vanilla/proof.rs
```diff
@@ -402,6 +448,26 @@ impl<'a, Tree: 'static + MerkleTreeTrait, G: 'static + Hasher> StackedDrg<'a, Tr
...
                             ));
                         }
+
+                        // Check if a proof was detected as invalid
+                        let invalid_comm_d_coordinate = invalid_comm_d.lock().expect("failed to get lock on invalid_comm_d");
+                        if invalid_comm_d_coordinate.failure_detected {
+                            return Err(anyhow!(
+                                "Invalid comm_d detected at challenge_index {}",
+                                invalid_comm_d_coordinate.challenge_index));
+                        }
```
vmx, line 458: It can be rewritten with the [`ensure!()` macro](https://docs.rs/anyhow/1.0.82/anyhow/macro.ensure.html):
```rust
ensure!(
    !invalid_comm_d_coordinate.failure_detected,
    "Invalid comm_d detected at challenge_index {}",
    invalid_comm_d_coordinate.challenge_index
);
```
The same is true for the ones below.

storage-proofs-porep/src/stacked/vanilla/proof.rs
```diff
@@ -431,40 +490,70 @@ impl<'a, Tree: 'static + MerkleTreeTrait, G: 'static + Hasher> StackedDrg<'a, Tr
...
+                        );
+                    }
                 });
             }
         });
 
+        let invalid_synth_porep_proof_coordinate = invalid_synth_porep_proof
+            .lock()
+            .expect("failed to get lock on invalid_synth_porep_proof");
+        if invalid_synth_porep_proof_coordinate.failure_detected {
```
vmx, line 551: One more case for `ensure!()`.

storage-proofs-porep/src/stacked/vanilla/proof.rs
```diff
@@ -140,8 +152,7 @@ impl<'a, Tree: 'static + MerkleTreeTrait, G: 'static + Hasher> StackedDrg<'a, Tr
             Challenges::Synth(synth_challenges) => {
                 // If there are no synthetic vanilla proofs stored on disk yet, generate them.
                 if pub_inputs.seed.is_none() {
-                    info!("generating synthetic vanilla proofs in a single partition");
-                    assert_eq!(partition_count, 1);
```
DrPeterVanNostrand, line 144: Looks good to me! I'm not seeing why this one `assert` was removed, but I won't let it delay approval.

DrPeterVanNostrand (approved):

storage-proofs-porep/src/stacked/vanilla/proof.rs
DrPeterVanNostrand: Nvm, I'm looking at #1750 now.

cryptonemo: > Looks good to me! I'm not seeing why this one `assert` was removed, but I won't let it delay approval.

Adding it was an error.  The partition count is only 1 for test sectors in porep, and the logging comment implied the incorrect thing (i.e that the single pass of synth porep proof generation implies that the partition count must be 1, which is wrong; it just meant we were not generating synth porep proofs with regard to challenges being partition specific like in other places).

cryptonemo: ^Fortunately, it was never released with that assert.  It came up quickly in overall testing while looking into this issue though.

PR #1754, opened 2024-05-08
fix: use a fixed seed for Non-interactive PoRep

vmx:
The Interactive PoRep requires a seed for the proof. The Non-interactive PoRep does not. In order to keep the public APIs simple, just make sure that the passed in seed in case of a Non-interactive PoRep is a fixed one. It was choosen to be all zeros as such a byte array is easy to generate.

That fixed seed is used for the proving (and the internal aggregation) as well as the verification.

This commit also removes the restriction that a seed cannot be all zeros, which it also totally could be by coincidence (it's randomly generated in case of the Interactive PoRep).

### DIFF
--- a/filecoin-proofs/src/api/seal.rs
+++ b/filecoin-proofs/src/api/seal.rs
@@ -41,7 +41,8 @@ use crate::{
         get_stacked_verifying_key,
     },
     constants::{
-        DefaultBinaryTree, DefaultPieceDomain, DefaultPieceHasher, SINGLE_PARTITION_PROOF_LEN,
+        DefaultBinaryTree, DefaultPieceDomain, DefaultPieceHasher,
+        POREP_NON_INTERACTIVE_FIXED_SEED, SINGLE_PARTITION_PROOF_LEN,
     },
     parameters::setup_params,
     pieces::{self, verify_pieces},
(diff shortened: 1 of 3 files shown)


cryptonemo:
> This commit also removes the restriction that a seed cannot be all zeros, which it also totally could be by coincidence (it's randomly generated in case of the Interactive PoRep).

No, you can't remove this.  It changes past chain behaviour and could cause an unnecessary fork.  There's also no reason to do this.

Instead of requiring NI-PoRep get passed an arbitrary fixed seed, I'd rather see it overwritten/generated internally and wired through as a fixed value.  So user passed in value is ignored and we pass in our own internal one.  It should also be listed in the spec (the specific value).

DrPeterVanNostrand:
My idea for the fix may be even simpler/dumber than the two above (either requiring a constant be passed in for the `seed` or overwriting the passed in `seed` with a constant); just don't hash the `seed` that was passed in by the user into the SnarkPack transcript (because the public inputs for NI-PoRep don't depend on the seed, and afaik, that means that it **shouldn't** be hashed into the transcript).

So changing [this code block](https://github.com/filecoin-project/rust-fil-proofs/blob/0c920258740a79ebc881f142161375a52f988910/filecoin-proofs/src/api/seal.rs#L744) and [this code block](https://github.com/filecoin-project/rust-fil-proofs/blob/0c920258740a79ebc881f142161375a52f988910/filecoin-proofs/src/api/seal.rs#L838) in PoRep aggregation and verification to something like...

```rust
let hashed_seeds_and_comm_rs: [u8; 32] = {
    let mut hasher = Sha256::new();
    if porep_config.feature_enabled(ApiFeature::NonInteractivePoRep) {
        for comm_r in comm_rs.iter() {
            hasher.update(seed);
        }
    } else {
        for (seed, comm_r) in seeds.iter().zip(comm_rs.iter()) {
            hasher.update(seed);
            hasher.update(comm_r);
    }
    hasher.finalize().into()
};
```

vmx:
I'm closing this one in favour of #1755.

PR #1755, opened 2024-05-08
fix: remove challenge seed from NI-PoRep SnarkPack transcript

DrPeterVanNostrand:
Alternative to #1754. This PR proposes that any PoRep challenge seed(s) (provided as API arguments) not be included in NI-PoRep's SnarkPack transcript, as opposed to using a constant value for NI-PoRep's challenge seed.

Imho, merging of either PR should be held off until a decision is made during the NI-PoRep audit.

### DIFF
--- a/filecoin-proofs/src/api/seal.rs
+++ b/filecoin-proofs/src/api/seal.rs
@@ -352,6 +352,8 @@ pub fn seal_commit_phase1<T: AsRef<Path>, Tree: 'static + MerkleTreeTrait>(
     prover_id: ProverId,
     sector_id: SectorId,
     ticket: Ticket,
+    // Note: when using NI-PoRep the PoRep challenge generation seed is ignored, thus any value can
+    // be passed in here for `seed`.
     seed: Ticket,
     pre_commit: SealPreCommitOutput,
     piece_infos: &[PieceInfo],
--- a/filecoin-proofs/src/api/seal.rs
+++ b/filecoin-proofs/src/api/seal.rs
@@ -740,13 +740,22 @@ pub fn aggregate_seal_commit_proofs<Tree: 'static + MerkleTreeTrait>(
     // If we're not at the pow2 target, duplicate the last proof until we are.
     pad_proofs_to_target(&mut proofs, target_proofs_len)?;
 
-    // Hash all of the seeds and comm_r's pair-wise into a digest for the aggregate proof method.
+    // For standard PoRep, the SnarkPack transcript should include a hash of each aggregated PoRep's
+    // challenge seed and comm_r (pair-wise); however since NI-PoRep does not use a seed to generate
+    // it's challenges, any challenge seeds provided as arguments to this function should be ignored
+    // (and thus not be included in an NI-PoRep's SnarkPack transcript).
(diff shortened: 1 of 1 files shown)


filecoin-proofs/src/api/seal.rs
```diff
@@ -740,13 +740,22 @@ pub fn aggregate_seal_commit_proofs<Tree: 'static + MerkleTreeTrait>(
...
+    // it's challenges, any challenge seeds provided as arguments to this function should be ignored
+    // (and thus not be included in an NI-PoRep's SnarkPack transcript).
     let hashed_seeds_and_comm_rs: [u8; 32] = {
         let mut hasher = Sha256::new();
-        for cur in seeds.iter().zip(comm_rs.iter()) {
-            let (seed, comm_r) = cur;
-            hasher.update(seed);
-            hasher.update(comm_r);
+        if porep_config.feature_enabled(ApiFeature::NonInteractivePoRep) {
+            for comm_r in comm_rs.iter() {
```
vmx, line 750: I think for Non-interactive PoRep, it's always only a single `comm_r` only. Should we even be stricter here and just hash that value, instead of looping through all (I know it's the same outcome, but it might be clearer to the reader).

vmx (commented):
Not hashing the seed at all makes sense (instead of hashing some fixed value.

I think it would make sense to add a comment to the entry point of the sealing at https://github.com/filecoin-project/rust-fil-proofs/pull/1755/files#diff-7360cc04999e8c2b4559e9d4b8bf83afaeaeb2be3e07314287b00fcc6b81d380L348 to note that for the Non-Interactive PoRep the seed is not used and that it can be any arbitrary value.

Would it make sense to refactor the hash calculation into its own function, so that it's again clearer that it does exactly the same thing for proving as well as for verification?

filecoin-proofs/src/api/seal.rs
DrPeterVanNostrand: Yeah, I wasn't totally sure what to do here. The way that it's currently written leaves open the possibility of aggregating multiple NI-PoReps into a single SnarkPack proof, however I don't know if that behavior should be allowed.

DrPeterVanNostrand:
> I think it would make sense to add a comment to the entry point of the sealing

Yup, sounds like a good idea. I'll add it.

filecoin-proofs/src/api/seal.rs
```diff
@@ -352,6 +352,8 @@ pub fn seal_commit_phase1<T: AsRef<Path>, Tree: 'static + MerkleTreeTrait>(
     prover_id: ProverId,
     sector_id: SectorId,
     ticket: Ticket,
+    // Note: when using NI-PoRep the PoRep challenge generation seed is ignored, thus any value can
+    // be passed in here for `seed`.
```
vmx, line 356: Just a note (feel free to ignore). I know we don't really use API docs here (their building is even broken). Though it could make sense to put it into the top of the function (or another note there) as a doc comment, in case we ever properly generate those.

cryptonemo (approved):

vmx (approved):

PR #1756, opened 2024-05-15
fix: increase ni-porep min challenges for test sectors

DrPeterVanNostrand:
Changes the minimum number NI-PoRep challenges for test sectors from 4 to `ceil(12.8 * min_porep_challenges)` (i.e. 26 challenges) and the NI-PoRep partition count for test sectors from 2 to 13; bumping the partition count allows NI-PoRep test sectors to use their current Groth16 keys (i.e. 2 PoRep challenges per circuit/partition).

### DIFF
--- a/filecoin-proofs/src/constants.rs
+++ b/filecoin-proofs/src/constants.rs
@@ -125,11 +125,11 @@ pub(crate) const fn get_porep_interactive_minimum_challenges(sector_size: u64) -
 }
 
 /// Returns the minimum number of challenges used for the non-interactive PoRep fo a certain sector
-/// size.
+/// size, i.e. `ceil(12.8 * interactive_porep_min_challenges)`.
 pub(crate) const fn get_porep_non_interactive_minimum_challenges(sector_size: u64) -> usize {
     match sector_size {
         SECTOR_SIZE_2_KIB | SECTOR_SIZE_4_KIB | SECTOR_SIZE_16_KIB | SECTOR_SIZE_32_KIB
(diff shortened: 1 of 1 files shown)


vmx (commented):
I think that change makes sense. Though we should also think about how much it increases the testing time. The original "4" was chosen to have a similar run time to the Interactive PoRep times. The test sizes are really for developing and there not to long testing times matter.

cryptonemo (approved):
Nice, thanks!

vmx:
It looks like the only longer running test where it makes a difference is `test_ignored_release_filecoin_proofs` where the test time goes from 30 to 40mins. But if it turns out that the total run time is a problem, I guess we could always split tests into separate jobs.

cryptonemo:
> It looks like the only longer running test where it makes a difference is `test_ignored_release_filecoin_proofs` where the test time goes from 30 to 40mins. But if it turns out that the total run time is a problem, I guess we could always split tests into separate jobs.

Since it's actually testing the 'non-interactivity' aspect now where it wasn't previously, I think it's important to keep it and not sweat the automated testing time.

DrPeterVanNostrand:
> The original "4" was chosen to have a similar run time to the Interactive PoRep times

Oh ok, that was a good idea. I didn't realize that was how the number of challenges was chosen so I wasn't even thinking about test times.

DrPeterVanNostrand:
> not sweat the automated testing time

Ok, I'm going to merge.

PR #1757, opened 2024-05-16
feat: expose an API suitable for NI-PoRep proof aggregation

cryptonemo:
feat: add tests to ensure NI-PoRep is tested w and w/o aggregation

### DIFF
--- a/filecoin-proofs/tests/api.rs
+++ b/filecoin-proofs/tests/api.rs
@@ -624,7 +624,7 @@ fn test_seal_proof_aggregation_2kib() -> Result<()> {
         (1, 5, ApiVersion::V1_1_0, vec![]),
         (5, 5, ApiVersion::V1_2_0, vec![ApiFeature::SyntheticPoRep]),
         (
-            257,
+            65,
             5,
             ApiVersion::V1_2_0,
             vec![ApiFeature::NonInteractivePoRep],
@@ -653,7 +653,7 @@ fn test_seal_proof_aggregation_4kib() -> Result<()> {
         (7, 5, ApiVersion::V1_1_0, vec![]),
         (24, 5, ApiVersion::V1_2_0, vec![ApiFeature::SyntheticPoRep]),
         (
-            123,
+            17,
             5,
             ApiVersion::V1_2_0,
             vec![ApiFeature::NonInteractivePoRep],
@@ -682,7 +682,7 @@ fn test_seal_proof_aggregation_32kib() -> Result<()> {
         (220, 5, ApiVersion::V1_1_0, vec![]),
         (500, 5, ApiVersion::V1_2_0, vec![ApiFeature::SyntheticPoRep]),
         (
-            818,
+            5,
             5,
             ApiVersion::V1_2_0,
             vec![ApiFeature::NonInteractivePoRep],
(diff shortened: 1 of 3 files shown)


filecoin-proofs/tests/api.rs
```diff
@@ -614,75 +619,89 @@ fn seal_lifecycle_upgrade<Tree: 'static + MerkleTreeTrait<Hasher = TreeRHasher>>
...
-#[test]
-#[ignore]
-fn test_seal_proof_aggregation_3_2kib_porep_id_v1_1_base_8() -> Result<()> {
-    let proofs_to_aggregate = 3; // Requires auto-padding
+fn test_seal_proof_aggregation_2kib() -> Result<()> {
+    let test_inputs = vec![
+        (1, 5, ApiVersion::V1_1_0, vec![]),
+        (5, 5, ApiVersion::V1_2_0, vec![ApiFeature::SyntheticPoRep]),
+        (
+            257,
```
DrPeterVanNostrand, line 627: Just out of curiosity, was aggregation of 257 test-sector PoRep Groth16 proofs chosen for a specific reason?

cryptonemo: Any of the non-pow2 sizes are testing padding to pow2.  We had an existing test that used that value, so it was replaced with an equivalent.

filecoin-proofs/tests/api.rs
```diff
@@ -1923,7 +1946,9 @@ fn generate_proof<Tree: 'static + MerkleTreeTrait>(
     seed: [u8; 32],
     pre_commit_output: &SealPreCommitOutput,
     piece_infos: &[PieceInfo],
+    aggregation_enabled: bool,
```
DrPeterVanNostrand, line 1949: This is optional, but I had some difficulty following where exactly aggregation was happening in the tests, maybe this parameter could be named something like `return_groth16_proofs` or `return_ni_porep_groth16_proofs`?

`return_groth16_proofs` would always be `true` when running interactive PoRep (because `seal_commit_phase2` always returns Groth16 proofs for interactive PoRep) and when `true` for NI-PoRep `generate_proof` returns Groth16 proofs (conversely when `false`, `generate_proof` returns the aggregate proof generated by `seal_commit_phase2`).

DrPeterVanNostrand (approved):
Looks good to me, I just have a couple optional questions/comments.

filecoin-proofs/tests/api.rs
DrPeterVanNostrand: Ok, sounds good. I was curious because (I believe that) this is saying to aggregate 257 NI-PoRep proofs, so 3341 Groth16 proofs (each NI-PoRep proof containing 13 Groth16/partition proofs), and if 257 was chosen arbitrarily when maybe a number like 160 would result in the same SnarkPack padding (i.e. `ceil(log2(257 * 13)) == ceil(log2(160 * 13)) == 12`).

cryptonemo: Good call on the count -- the FIP actually specifies bounds that I'll add

filecoin-proofs/tests/api.rs
cryptonemo: I'll leave this as-is, since it doesn't affect the exposed API

DrPeterVanNostrand: Sounds good to me.

vmx (commented):
Just making sure I understand correctly what this is about. In the last step of the Ni-PoRep we aggregate the proofs. We cannot aggregate aggregated proofs. Hence if we want to aggregate several Ni-PoReps, we need the unaggregated result and then aggregate all of them. Is that correct?

cryptonemo:
> Just making sure I understand correctly what this is about. In the last step of the Ni-PoRep we aggregate the proofs. We cannot aggregate aggregated proofs. Hence if we want to aggregate several Ni-PoReps, we need the unaggregated result and then aggregate all of them. Is that correct?

Correct, it's specified in the FIP, but wasn't in the implementation.  Fortunately it was caught before shipping.

DrPeterVanNostrand (approved):

PR #1760, opened 2024-07-23
feat: use workspace inheritance for dependencies

### LINKED ISSUE #1645: Use workspace inheritance for dependencies
### Description

[Cargo 1.64.0 has a new feature called workspace inheritance](https://blog.rust-lang.org/2022/09/22/Rust-1.64.0.html#cargo-improvements-workspace-inheritance-and-multi-target-builds). You can define dependencies in your workspace and then crates within your workspace can refer to them. This way you define the version only once. This makes version upgrades way easier, while making sure all crates are on the same version.

For more information see https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#inheriting-a-dependency-from-a-workspace

### Acceptance criteria

All shared dependencies are defined in the workspace and the individual crates refer to them.

### Risks + pitfalls

I'm not sure how tricky it is to get everything right, given we use features and optional dependencies quite a bit. But as soon as CI is green, I'm confident that things are correct.

### Where to begin

Increase the `rust-toolchain` version to 1.64.0. Then look at the dependencies of the crates and determine shared dependencies. Put those into the workspace and inherit their values from the crates themselves.

kamuik16:
Closes #1645 

cc @vmx

### DIFF
--- a/storage-proofs-update/Cargo.toml
+++ b/storage-proofs-update/Cargo.toml
@@ -9,14 +9,14 @@ repository = "https://github.com/filecoin-project/rust-fil-proofs"
 readme = "README.md"
 
 [dependencies]
-filecoin-hashers = { workspace = true, features = ["poseidon", "sha256"]}
+filecoin-hashers = { workspace = true, features = ["poseidon", "sha256"] }
 fr32.workspace = true
 storage-proofs-core.workspace = true
 storage-proofs-porep.workspace = true
 # Sorted alphabetically
 merkletree.workspace = true
 rayon.workspace = true
-serde = { version = "1.0", features = ["derive"]}
+serde = { workspace = true, features = ["derive"] }
 ff.workspace = true
 bellperson.workspace = true
 blstrs.workspace = true
--- a/storage-proofs-update/Cargo.toml
+++ b/storage-proofs-update/Cargo.toml
@@ -9,27 +9,29 @@ repository = "https://github.com/filecoin-project/rust-fil-proofs"
 readme = "README.md"
 
 [dependencies]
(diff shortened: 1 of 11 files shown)


storage-proofs-update/Cargo.toml
```diff
@@ -9,27 +9,29 @@ repository = "https://github.com/filecoin-project/rust-fil-proofs"
...
-merkletree = "0.23.0"
-rayon = "1.0.0"
+filecoin-hashers = { workspace = true, features = ["poseidon", "sha256"]}
+fr32.workspace = true
+storage-proofs-core.workspace = true
+storage-proofs-porep.workspace = true
+# Sorted alphabetically
+merkletree.workspace = true
+rayon.workspace = true
 serde = { version = "1.0", features = ["derive"]}
```
vmx, line 19: `serde` can be used from the workspace.

vmx (changes requested):
Thanks! It generally looks good, I only found two small things.

I also compared the generated `Cargo.lock` file to the one the master branch creates. It's identical which indicates that things are correct.

kamuik16:
> Thanks! It generally looks good, I only found two small things.
> 
> I also compared the generated `Cargo.lock` file to the one the master branch creates. It's identical which indicates that things are correct.

Thanks for the review. Fixed suggestions in https://github.com/filecoin-project/rust-fil-proofs/pull/1760/commits/8774e06bbbe5c189c430412e901fbe972184f9d0.

vmx (approved):
Thanks a lot, this looks good. Before merging I'd like to get an approval from another team member.

cryptonemo (approved):
Looks good, thanks

PR #1761, opened 2024-08-29
feat: properly remove temporary files leftover after running tests

cryptonemo:
It took a while to track down which tests were leaving around `.tmp*` directories and files in the `TEMPDIR` after the test suite was run, but it appears that this resolves them all across normal and 'ignored' tests.


cryptonemo:
Needs more testing, closing for now

PR #1763, opened 2024-09-03
chore: move CI to GitHub Actions

vmx:
Move the CI from CircleCI to GitHub actions.

As GitHub Actions work a bit differently, things needed to change a bit and I took the chance to clean things up a bit. Nonetheless the coverage should be the same.

If you compare it to the CircleCI configuration and wonder how different things are, this list of changes might help:

 - Seperate multicore-sdr run was removed, it's enabled on the normal tests
 - Building benchmarks was removed as `--all-targets` already implies building the benchmarks
 - All the `test` jobs are removed, they are already covered by the release mode tests that run for the whole workspace
 - Some of the `ignored` tests were run separately. Now simply the whole workspace is run in a single job
 - The `no_gpu` test was renamed to `no-default-features`
 - `test_darwin` was renamed to `test_macos`
 - The `test_arm_no_gpu` wasn't ported as the MacOS runner is an M1, hence there is already basic ARM sanity checking


.github/README.md
```diff
@@ -0,0 +1,12 @@
+GitHub Actions setup
+====================
+
+The main CI is happening in the [`ci.yml`], this is where the test are run.
+
+There is a separate workflow called [`proof-params-generate.yml`], that pre-generates the Groth16 parameter files needed for testing. Those parameters are stored as GitHub Artifacts, which are then downloaded by any workflow that needs them. Those artifacts are available for 90 days. Hence this job runs every to months to regenerate them, but also to have enough time in case something breaks.
```
cryptonemo, line 6: typo: every to months -> every two months (?)

.github/README.md
```diff
@@ -0,0 +1,12 @@
+GitHub Actions setup
+====================
+
+The main CI is happening in the [`ci.yml`], this is where the test are run.
```
cryptonemo, line 4: typo: test are run -> tests are run

cryptonemo:
I'm still working on the self hosted runner setup (w/gpu), but if we use that, I think we should be able to skip a lot of the param generation/usage and dep installs, etc.  But that's for a later PR I guess

cryptonemo (approved):

cryptonemo (approved):

PR #1765, opened 2024-10-04
Add seal regression testing support

cryptonemo:
feat: add seal regression testing support
feat: add tests from v16 and v18*

*v17 missed result collection for various reasons, but could always be added later if needed


filecoin-proofs/src/types/mod.rs
```diff
@@ -108,6 +108,18 @@ pub struct SectorUpdateProofInputs {
...
+    pub porep_config: PoRepConfig,
+    pub comm_r: Commitment,
+    pub comm_d: Commitment,
+    pub prover_id: ProverId,
+    pub sector_id: SectorId,
+    pub ticket: [u8; 32],
+    pub seed: [u8; 32],
+    pub proof: Vec<u8>,
+}
+
```
vmx, line 122: If this type is put into the `filecoin-proofs/tests/regression.rs` file it could be private and hence wouldn't be part of the public interface.

filecoin-proofs/Cargo.toml
```diff
@@ -37,6 +37,7 @@ serde = { workspace = true, features = ["rc", "derive"] }
 serde_json.workspace = true
 sha2.workspace = true
 typenum.workspace = true
+file-lock="2.1.10"
```
DrPeterVanNostrand, line 40: Can this be an optional dependency?
```toml
file-lock= { version = "2.1.10", optional = true }

[dev-dependencies]
persist-regression-proofs = ["dep:file-lock"]
```

filecoin-proofs/tests/regression.rs
```diff
@@ -0,0 +1,265 @@
...
+            proof: commit_output.proof.clone(),
+        };
+
+        let json = serde_json::to_string(&record)?;
+
+        // When appending to an existing file, the file lock is required
+        // to avoid corruption when multiple tests running in parallel are
+        // persisting generated proofs
+        let options = FileOptions::new().write(true).create(true).append(true);
+        let mut filelock = FileLock::lock(output_path, true /* block */, options)?;
```
DrPeterVanNostrand, line 100: Just flagging this line in case the "block" comment was meant to be removed before merging.

filecoin-proofs/tests/regression.rs
```diff
@@ -0,0 +1,265 @@
...
+use filecoin_proofs::{Commitment, PoRepConfig, ProverId};
+#[cfg(feature = "persist-regression-proofs")]
+use filecoin_proofs::{
+    MerkleTreeTrait, SealCommitOutput, SealPreCommitOutput, PUBLISHED_SECTOR_SIZES,
+};
+use storage_proofs_core::sector::SectorId;
+
+const V16_SEAL_REGRESSION_RECORDS: &str = "seal_regression_records-v16.json";
+const V18_SEAL_REGRESSION_RECORDS: &str = "seal_regression_records-v18.json";
+
```
DrPeterVanNostrand, line 25: Just an optional thought, but it may be a good idea to use a constant here
```rust
const SEAL_REGRESSION_RECORDS: [(&str, &str); 2] = [
    (V16_SEAL_REGRESSION_RECORDS, "v16"),
    (V18_SEAL_REGRESSION_RECORDS, "v18"),
];

fn run_seal_regression_tests() {
    ...
    for (path, version) in SEAL_REGRESSION_RECORDS {
        ...
    }
    ...
}
```
so that when new regression test files are added, only this files's `const`s need to be modified (rather than requiring both the `const`s and `fn run_seal_regression_tests` be modified).

filecoin-proofs/tests/regression.rs
cryptonemo: That was intentional to give context to what the hard coded `true` value is specifying

DrPeterVanNostrand (approved):

PR #1766, opened 2024-11-14
fix: upgrade hwloc

jmg-duarte:
When running `rust-fil-proofs` under Rust 1.81 the following warning is issued:

```
warning: the following packages contain code that will be rejected by a future version of Rust: bitflags v0.7.0
note: to see what the problems were, use the option `--future-incompat-report`, or run `cargo report future-incompatibilities --id 1`
```

The package depending on `bitflags` 0.7.0 is `hwloc`, which has been replaced by `hwloc2`.
Understandably, `hwloc` (the original library) has suffered  changes as well, which are reflected in `hwloc2`.

The main one being the removal of `allowed_cpuset`, according to the `hwloc` upgrading guide (https://www.open-mpi.org/projects/hwloc/doc/v2.8.0/a00374.php#upgrade_to_api_2x_allowed) these `cpuset` should be enough given that the [`INCLUDE_DISALLOWED`](https://docs.rs/hwloc2/latest/hwloc2/enum.TopologyFlag.html#variant.IncludeDisallowed) flag hasn't been passed (and as far as I know, it isn't a default).

This PR attempts to address these issues.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,7 +1,7 @@
 name: CI
 
-#on: [pull_request, push]
-on: [push]
+on: [pull_request, push]
+#on: [push]
 
 # Cancel a job if there's a new on on the same branch started.
 # Based on https://stackoverflow.com/questions/58895283/stop-already-running-workflow-job-in-github-actions/67223051#67223051
--- a/storage-proofs-porep/Cargo.toml
+++ b/storage-proofs-porep/Cargo.toml
@@ -9,7 +9,7 @@ repository = "https://github.com/filecoin-project/rust-fil-proofs"
 readme = "README.md"
 
 [dependencies]
-filecoin-hashers = { workspace = true, features = ["poseidon", "sha256"]}
+filecoin-hashers = { workspace = true, features = ["poseidon", "sha256"] }
 fr32.workspace = true
 sha2raw.workspace = true
 storage-proofs-core.workspace = true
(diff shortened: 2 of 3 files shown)


cryptonemo:
Hello @jmg-duarte -- thanks for submitting this, we'll take a look!

jmg-duarte:
@cryptonemo I just noticed tests are failing

I can't quite understand why since the docs I pointed too seem to indicate it should work?
Either way, I'm keen on getting this merged, I may need some pointers where to look for references though.

Thanks!

cryptonemo:
> @cryptonemo I just noticed the `stacked::vanilla::cores::tests::test_cores` is failing
> 
> I can't quite understand why since the docs I pointed too seem to indicate it should work? Either way, I'm keen on getting this merged, I may need some pointers where to look for references though.
> 
> Thanks!

That particular test can fail if other tests are running at the same time.  CI should isolate it properly.

jmg-duarte:
If `hwloc2` does not cut it, the alternative I found is [`hwlocality`](https://github.com/HadrienG2/hwlocality), it requires more changes but at least it is maintained 😓

cryptonemo:
> If `hwloc2` does not cut it, the alternative I found is [`hwlocality`](https://github.com/HadrienG2/hwlocality), it requires more changes but at least it is maintained 😓

CI aside, what we need for this to 'cut it' is to determine if there's any noticeable (negative) performance impact on multicore-sdr sealing (in particular for production sector sizes of 32/64GiB).  Gathering that data will take some time.

cryptonemo:
My testing of production sectors checks out(!) @DrPeterVanNostrand @vmx Any inputs, or good to go?

vmx (approved):
@cryptonemo Thanks for testing it out. I think then it's good to go.

PR #1768, opened 2025-01-03
chore: fix building master branch

vmx:
The current build fails due to:

error: package `home v0.5.11` cannot be built because it requires rustc 1.81 or newer, while the currently active rustc version is 1.70.0

This commit increases the Rust version from 1.70.0 to 1.81.0.

Also fix most of the Clippy errors. One about an `Arc` is still there as a fix is non-trivial, needs proper testing and hence deserves a separate PR.


vmx:
The CI runners seem to have an issue. It looks like Rust wasn't installed by rustup, else I think the new version would be checked out automatically.

cryptonemo:
Closing for now in favor of https://github.com/filecoin-project/rust-fil-proofs/pull/1770

PR #1767, opened 2024-12-02
chore(deps): upgrade a few dependencies

hanabi1224:
This PR upgrades a few dependencies in the workspace.

PR #1769, opened 2025-01-03
fix: Arc can be an Rc

vmx:
In this case the value isn't shared across threads, hence it can be an `Rc` instead of an `Arc`.

This was found via a newly introduced Clippy warning.


cryptonemo:
Thanks!  This fix in particular has been in testing for some time along with some other updates, but isn't ready quite yet.  I'll get back to it this week though: https://github.com/filecoin-project/rust-fil-proofs/pull/1767

cryptonemo:
Closing for now in favor of https://github.com/filecoin-project/rust-fil-proofs/pull/1770

PR #1773, opened 2025-02-18
feat: make 1GiB sector harder

th7nder:
Adjusts 1GiB sector's security params to bigger ones.
It requires around 32GiB + 32GiB of swap for generating the params.
Instead of taking 8 seconds for proving, it takes 220 seconds.


th7nder:
Sorry, wrong branched. It was supposed to be used for our internal fork.

cryptonemo:
> Sorry, wrong branched. It was supposed to be used for our internal fork.

No problem!

PR #1778, opened 2025-05-23
tests: put large regression tests behind feature flag

vmx:
Put the regression tests for sector sizes > 32KiB behind the `big-tests` feature flag. This way a test run without the flag doesn't need the parameter files for all sector sizes.

For running with the `big-tests` flag, the parameter files for all sector sizes are expected.

### DIFF
--- a/filecoin-proofs/tests/regression.rs
+++ b/filecoin-proofs/tests/regression.rs
@@ -4,12 +4,16 @@ use std::path::Path;
 
 use anyhow::Result;
 use filecoin_proofs::{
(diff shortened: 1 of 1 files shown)


vmx:
Usually I wait for the CI to be green before merging. But the MacOS tests take another hour to run. The code that was modified isn't run on MacOS anyway, hence I dare to merge it now.

cryptonemo:
These tests only required verifying keys btw, but either way :man_shrugging:

vmx:
@cryptonemo oh, that's good to know. I hope you're still OK with this PR. I think I like the consistency :)

PR #1785, opened 2025-06-12
ci: test gpu on self-hosted runners

galargh:
This introduces the following changes:
- changes the CI workflow to be triggered only on pushes to `master` and all pull requests
- migrates the test release job back to a self-hosted runner (no GPU needed)
- migrates the test no default features job back to a self-hosted runner (no GPU needed)
- reintroduces the test gpu jobs and puts them on self-hosted runners

The test gpu jobs are currently put on `g6e.xlarge` machines because we only have limited number of resources (vCPUs) available. When we are granted the quota increase, I'll migrate the jobs to more powerful runners.

If approved, similar PRs will be created for other repositories we identified as needing self-hosted runners.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,7 +1,10 @@
 name: CI
 
-on: [pull_request, push]
-#on: [push]
+on:
+  pull_request:
+  push:
+    branches:
+      - master
 
 # Cancel a job if there's a new on on the same branch started.
 # Based on https://stackoverflow.com/questions/58895283/stop-already-running-workflow-job-in-github-actions/67223051#67223051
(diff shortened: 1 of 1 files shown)


BigLep (dismissed):
Sweet.  I'm not sure how you figured out the nvdia drier install stuff, but I'm glad you got it working.  Thakns.

.github/workflows/ci.yml
```diff
@@ -40,42 +45,47 @@ jobs:
...
 
       - name: Download the proof params
         uses: ./.github/actions/proof-params-download
         with:
           github-token: ${{ secrets.GITHUB_TOKEN }}
 
+      - uses: dtolnay/rust-toolchain@21dc36fb71dd22e3317045c0c31a3f4249868b17
+        with:
+          toolchain: 1.83
+
```
vmx, line 72: I recommend installing Rust on the self hosted runner via [rustup](https://rustup.rs). This way you wouldn't need this. The benefit is that then invocations of `cargo` would automatically install the Rust version specified in the `rust-toolchain` file that is in the root directory of this repo.

This would then also mean that you wouldn't need the CI file, whenever you update the version in `rust-toolchain`.

vmx (commented):
See inline comment, the rest looks good.

.github/workflows/ci.yml
galargh: That's a good point. I'll add a TODO for this in our backlog. To be addressed when we rebuild the AMI for the machines.

I also want to install the GPU drivers in the AMI as well so that we wouldn't have to do that every time a job runs.

